PR #3, opened 2025-05-22
feat: basic command implementation

wan9chi:


packages/cli/src/index.ts
```diff
@@ -1,30 +1,67 @@
...
+  );
+  exec(program, args);
+}
+
+function exec(program: string, args: string[]) {
+  const { status, error } = spawnSync(program, args, { stdio: "inherit" });
+  if (error !== undefined) {
+    throw error;
+  }
+  process.exit(status ?? 1);
```
sxzz: Use 255 for critical error
```suggestion
  process.exit(status ?? 255);
```

packages/cli/src/index.ts
```diff
@@ -1,30 +1,67 @@
...
+  cli.command(viteCommand, "", () => {
+    execPackageBin("vite", args);
+  });
+}
+cli.command("lib", "", () => {
+  execPackageBin("tsdown", commandArgs);
+});
+
+cli.command("run", "", () => {
+  exec("node", [
```
sxzz:
```suggestion
  exec(process.execPath, [
```

packages/cli/src/index.ts
```diff
@@ -1,30 +1,67 @@
...
-cli.help().parse()
+import yargs from "yargs/yargs";
+import { hideBin } from "yargs/helpers";
+import { spawnSync } from "node:child_process";
+import path from "node:path";
+
+function execPackageBin(binName: string, args: string[]) {
+  // TODO: exec from bundled packages
+  const program = path.join(
+    import.meta.dirname,
```
sxzz, line 9: `import.meta.dirname` is added in Node v21.2.0 and v20.11.0. We should add `engines` to `package.json`

PR #4, opened 2025-06-06
Add global binary `vp` and delegate to `vite-plus`

webpro:
- Add `global` package with `vp` binary (just a temporary name)
- Please see README.md for more details

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "type": "module",
   "bin": {
-    "vite-plus": "./bin/vite-plus"
+    "vite-plus": "./bin/vite-plus",
+    "vpl": "./bin/vpl"
   },
   "exports": {
     ".": {
--- a/packages/cli/bin/vpl
+++ b/packages/cli/bin/vpl
@@ -1,0 +1,2 @@
+#!/usr/bin/env node
+import "../src/cli.ts";
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1,2 +1,2 @@
-import main from "./index.js";
+import main from "./index.ts";
 main();
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "allowImportingTsExtensions": true,
+    "noEmit": true,
     "target": "esnext",
     "lib": ["es2022"],
     "moduleDetection": "force",
--- a/packages/global/bin/vpg
+++ b/packages/global/bin/vpg
@@ -1,0 +1,2 @@
+#!/usr/bin/env node
+import "../src/index.ts";
--- a/packages/global/package.json
+++ b/packages/global/package.json
@@ -4,7 +4,8 @@
   "version": "0.0.0",
   "type": "module",
   "bin": {
-    "vp": "./bin/vp"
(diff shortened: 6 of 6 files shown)


packages/cli/src/index.ts
```diff
@@ -1,67 +1,52 @@
-import yargs from "yargs/yargs";
-import { hideBin } from "yargs/helpers";
 import { spawnSync } from "node:child_process";
 import path from "node:path";
+import { parseArgs } from "node:util";
```
webpro, line 3: Using built-in arg parser is "good enough" imho, and would also work with e.g. Bun.

packages/global/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import { execFileSync } from "node:child_process";
+import { parseArgs } from "node:util";
```
webpro, line 2: Using the built-in arg parser is nice for zero-dependencies in global package.

packages/cli/src/index.ts
```diff
@@ -1,67 +1,52 @@
...
-    execPackageBin("vite", args);
-  });
+export default function main(): void {
+  const { positionals } = parseArgs({ allowPositionals: true });
+
+  const [command, ...rest] = positionals;
+
+  const tasks = rest.filter((task) => task.includes("#"));
+
+  const [subcommand, packageDir] = command === "task" && tasks.length > 0 ? tasks[0].split("#") : [command, "."];
```
webpro, line 25: This is just dummy "task orchestrator", can only handle one task from CLI e.g. `vite-plus task dev#packages/app`

packages/cli/src/index.ts
```diff
@@ -1,67 +1,52 @@
...
-function execPackageBin(binName: string, args: string[]) {
-  // TODO: exec from bundled packages
-  const program = path.join(
-    import.meta.dirname,
-    "../node_modules/.bin",
-    binName,
-  );
-  exec(program, args);
+function execPackageBin(binName: string, args: string[], cwd: string, dir?: string) {
+  const program = path.join(import.meta.dirname, "../node_modules/.bin", binName);
```
sxzz, line 6: I'm not sure the compatibility of pnpm

packages/cli/package.json
```diff
@@ -3,23 +3,31 @@
   "version": "0.0.0",
   "private": true,
   "type": "module",
-  "scripts": {
-    "start": "node --import tsx/esm ./src/index.ts"
```
wan9chi, line 7: What is our dev experience for vite+? Should we start it directly from the source with `tsx`, or do we always build it with `tsdown`? Currently `execPackageBin` in `cli/src/index.ts` assumes that it runs from the source where `../node_modules/.bin` exists.

packages/cli/src/index.ts
wan9chi: Let's revisit this when we want to proxy help messages from the tools.

packages/cli/package.json
webpro: Great point, using verdaccio I was only doing the "real" thing.

How about having `vpg` and `vpl` for development? See d104c6607309518c155559256ce42964aeb0b192. This is running the actual code without anything in between, but does require Node.js v24. From `packages/cli` and `packages/global` we can do e.g. `npm link` to make them globally available.

webpro: - `vpg`: vite-plus global
- `vpl`: vite-plus local

PR #9, opened 2025-06-14
feat(tui): initialize vite tui

Boshen:
Project modified from running `cargo generate ratatui/templates` and then select `component`.

Features:

* Press `esc` or `q` to quit
* Print log to a file via https://crates.io/crates/directories
* spawn tasks
* press up / down to select tasks
* relay task output to tasks pane

image

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ pedantic = { level = "warn", priority = -1 }
 nursery = { level = "warn", priority = -1 }
 cargo = { level = "warn", priority = -1 }
 cargo_common_metadata = "allow"
-multiple_crate_versions = "allow"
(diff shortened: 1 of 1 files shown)


crates/vite_tui/README.md
```diff
@@ -0,0 +1,13 @@
...
+## Development
+
+```bash
+RUST_LOG=DEBUG watchexec 'cargo run'
+```
+
+In another terminal:
+
+```bash
+tail -f ~/Library/Application\ Support/dev.voidzero.vite_tui/vite_tui.log
```
Boshen, line 12: FYI: current mechanism dumps the log without ever deleting it.

PR #16, opened 2025-06-27
Use vite 

webpro:
* Rename vp → vite (and vpg → vite-dev)
* Restore direct execution of vite [toolchain] commands
* Move task into single file
* Remove extra scripts from templates
* Add --help text

(branched from previous PR to avoid git conflicts)

## Example monorepo

Screenshot 2025-06-27 at 3 11 37 PM

## Example lib package

Screenshot 2025-06-27 at 3 14 07 PM

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1,2 +1,2 @@
-import { runTasks } from "./runTasks.ts";
-runTasks();
+import { run } from "./run.ts";
+run();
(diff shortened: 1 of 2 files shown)


packages/cli/src/runTasks.ts
wan9chi, line 1: `runTasks` now becomes the entry to run any subcommand, not just `vite task`. It should be renamed to reflect this.

webpro: Fair enough. They're still tasks to run, but I've renamed to a more generic "run" fn name.

PR #19, opened 2025-07-03
feat: persist cache using SQLite & support passing additional args

wan9chi:
- persist cache using `bincode` format and sqlite
- allow show cache in json with `vite_task --debug`
- support passing additional args: `vite_task some_task -- arg1 arg2`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,11 @@ serde_yml = "0.0.12"
 twox-hash = "2.1.1"
 vite_workspace = { path = "crates/vite_workspace" }
 wax = "0.6.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -48,6 +48,11 @@ serde_yml = "0.0.12"
 twox-hash = "2.1.1"
 vite_workspace = { path = "crates/vite_workspace" }
 wax = "0.6.0"
+bstr = "1.12.0"
+clap = "4.5.40"
```
Boshen, line 52: `clap` is too heavy, we use https://docs.rs/bpaf/latest/bpaf for all our projects.

PR #27, opened 2025-07-21
feat(cli): bootstrap rust addon

Brooooooklyn:


crates/vite_task/src/execute.rs
```diff
@@ -89,7 +89,7 @@ pub struct TaskEnvs {
 }
 
 impl TaskEnvs {
-    pub fn resolve(base_dir: &Path, task: &TaskConfig) -> anyhow::Result<Self> {
+    pub fn resolve(base_dir: &Path, task: &TaskConfig, config_dir: &Str) -> anyhow::Result<Self> {
```
wan9chi, line 92: Can be simplified to
```suggestion
    pub fn resolve(base_dir: &Path, task: &ResolvedTaskConfig) -> anyhow::Result<Self> {
```

There are so many `Config` types right now in vite task. I should add some comments...

packages/cli/index.js
```diff
@@ -0,0 +1,397 @@
...
+      } catch (e) {
+        loadErrors.push(e);
+      }
+    } else {
+      loadErrors.push(new Error(`Unsupported architecture on macOS: ${process.arch}`));
+    }
+  } else if (process.platform === 'freebsd') {
+    if (process.arch === 'x64') {
+      try {
+        return require('./vite-plus.freebsd-x64.node');
```
wan9chi, line 176: I see only macOS/Windows/Linux targets are listed in `napi.targets` in `package.json`.

Is it expected to see `freebsd` and other unlisted OSes/arches here?

Brooooooklyn: The binding template contains all NAPI-RS supported platform/cpu combination

PR #28, opened 2025-07-23
feat: unify the error

Brooooooklyn:
Make the pnpm -r build pass

Migrate `std::sync::Mutex` to the `tokio::sync::Mutex`, Using `std::sync::Mutex` in async code can cause deadlocks and is difficult to debug.


crates/vite_task/src/schedule.rs
```diff
@@ -24,7 +24,7 @@ impl ExecutionPlan {
         // TODO: parallel
         let node_indices = match toposort(&task_graph, None) {
             Ok(ok) => ok,
-            Err(err) => anyhow::bail!("Circular depedency found in the task graph: {:?}", err),
+            Err(err) => anyhow::bail!("Circular dependency found in the task graph: {:?}", err),
```
wan9chi, line 27: Let's set up `typos` 😅

packages/cli/package.json
```diff
@@ -15,15 +15,18 @@
     "node": ">=20.11.0"
   },
   "scripts": {
-    "build": "napi build --platform --release --esm"
+    "build": "pnpm run build:binding && pnpm run build:ts",
+    "build:binding": "napi build --package-json-path ../package.json --cwd binding --platform --release --esm",
+    "build:ts": "tsdown --outDir dist src/index.ts"
```
wan9chi, line 20: If/when vite task can correctly cache the two phases of the `build` script, we should inline `build:binding` and `build:ts` into `build`, as there'd be no need to run them individually. Plus there is extra overhead of catching fs accesses of `pnpm run` itself.

PR #32, opened 2025-07-30
feat(vite-plus): support vite-plus run -r

Brooooooklyn:
For the `vite-plus run build --recursive --topological` command, in the current implementation, the call path is:

crates/vite_task/src/lib.rs: fn main

1. Workspace::load(cwd)? Build monorepo workspace package graph
2. Workspace::resolve_tasks(&tasks, task_args.clone(), recursive_run, topological_run)? Build task graph depends on the tasks to run
3. ExecutionPlan::plan(task_graph, parallel_run, topological_run)? Build execution plan
4. ExecutionPlan::execute(&mut workspace).await? Execute the execution plan

But it's not the best way to do it.

I want to build the task graph in the `Workspace::load(cwd)?`, and then only need to resolve the subtasks graph for the tasks need to run in the Workspace::resolve_tasks(&tasks, task_args.clone(), recursive_run, topological_run)?.

In this implementation, the same script in the package dependencies will be treated as task dependencies too.

Beside the refactor, there is one more problem in current implementation, in the if there are packages, A -> B -> C, A depends on B and B depends on C. If package A has a script `build` and package B doesn't have a script `build`, and package C has a script `build`, the current implementation will not resolve the right tasks graph. The current implementation will resolve the task graph as A#build, but the right task graph should be A#build -> C#build.
The problem is that the current implementation in crates/vite_task/src/config.rs `fn build_starting_with` is resolving the task graph from `A#build` and find the `B#build` in the edges, but it doesn't exist so it stopped.
I have test fixtures for this case in crates/vite_task/fixtures/transitive-dependency-workspace.


crates/vite_task/src/config.rs
```diff
@@ -298,20 +438,76 @@ impl Workspace {
...
+    /// @test/utils#build:
+    ///   [0] ──► [1] ──► [None]
+    ///   (subtasks depend on each other within package)
+    /// ```
+    ///
+    /// #### With --recursive --topological:
+    /// ```no_compile
+    /// @test/core#build ─────────┐
+    ///                           ▼
+    /// @test/utils#build: [0] ──► [1] ──► [None]
```
wan9chi: `@test/core#build` should point to `[0]`

crates/vite_task/src/schedule.rs
```diff
@@ -21,29 +21,69 @@ pub struct ExecutionPlan {
...
+    /// - Group 2: [@test/utils#build\[0\]] (depends on Group 1)
+    /// - Group 3: [@test/utils#build\[1\], @test/other#build] (can run in parallel)
+    #[tracing::instrument(skip(task_graph))]
+    pub fn plan(
+        mut task_graph: StableDiGraph<ResolvedTask, ()>,
+        parallel_run: bool,
+        topological_run: bool,
+    ) -> Result<Self, Error> {
+        // TODO: implement parallel execution grouping
+        let node_indices = if topological_run {
```
wan9chi: `toposort` should always be done here. Other than `-r`, task dependencies may come from `&&`, or explicit configuration. 

`topological_run` only affects whether we add edges between `pkgA#build` and `pkgB#build` in `-r` mode.

crates/vite_task/src/config.rs
```diff
@@ -203,33 +209,127 @@ impl TaskGraphBuilder {
...
+
+                    if let Some(current_node) = current_package_node {
+                        // Only add cross-package dependencies for the FIRST subtask
+                        // (subcommand_index == Some(0) or None for non-compound commands)
+                        let is_first_subtask =
+                            task_id.subcommand_index.map_or(true, |idx| idx == 0);
+
+                        if is_first_subtask {
+                            // Get all dependencies of the current package
+                            let dependencies: Vec<_> =
```
wan9chi: What if a task of the same name only exists in an indirect dependency?

- `pkgA` (has a `build` script)
  - `pkgB` (no `build` script)
    - `pkgC`  (has a `build` script)

I think there should be an edge between `pkgC#build` and `pkgA#build`.

crates/vite_task/src/lib.rs
```diff
@@ -12,36 +12,119 @@ mod str;
...
+            if name.is_empty() {
+                return Err(Error::EmptyPackageName(workspace.dir));
+            }
+            (
+                &vec![if task.contains('#') { task } else { format!("{name}#{}", task).into() }],
+                Arc::<[Str]>::from(args.task_args),
+            )
+        }
+    };
+
```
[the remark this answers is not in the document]
Brooooooklyn: This is a false positive. The temporary vector created with `vec![...]` in the None branch lives until the end of the match statement (line 124). Since tasks is used immediately after on line 126 within the same scope, there's no dangling reference. Rust's borrow checker correctly allows this pattern. You are likely misunderstood Rust's temporary lifetime extension rules.

crates/vite_task/src/schedule.rs
```diff
@@ -21,29 +21,63 @@ pub struct ExecutionPlan {
...
+        parallel_run: bool,
+        topological_run: bool,
+    ) -> Result<Self, Error> {
+        // TODO: implement parallel execution grouping
         let node_indices = match toposort(&task_graph, None) {
             Ok(ok) => ok,
             Err(err) => return Err(Error::CycleDependenciesError(err)),
         };
+
+        // Extract tasks from the graph in the determined order
```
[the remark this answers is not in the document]
Brooooooklyn: The `ExecutionPlan::plan` should always run in topological order; otherwise, the package relation is meaningless.

crates/vite_task/src/config/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+
+    fn with_unique_cache_path<F, R>(test_name: &str, f: F) -> R
+    where
+        F: FnOnce(&std::path::Path) -> R,
+    {
+        let test_id = TEST_COUNTER.fetch_add(1, Ordering::SeqCst);
+        let thread_id = std::thread::current().id();
+        let cache_path = std::env::temp_dir()
+            .join(format!("vite-test-{}-{}-{:?}.db", test_name, test_id, thread_id));
+
```
wan9chi: `std::thread::current().id()` is only locally unique within a single process. This may cause conflicts if multiple processes of this test binary are running concurrently (no that it would happen, but just in case). Wouldn't it be easier to just create tmp folder using `tempfile::tempdir()`?

crates/vite_task/src/config/task_graph_builder.rs
```diff
@@ -0,0 +1,162 @@
...
+use crate::{Error, collections::HashMap};
+
+use petgraph::{Graph, graph::NodeIndex, stable_graph::StableDiGraph};
+use vite_package_manager::{DependencyType, PackageInfo};
+
+use super::{ResolvedTask, TaskId};
+
+#[derive(Default, Debug, Clone)]
+pub struct TaskGraphBuilder {
+    pub(crate) resolved_tasks_and_dep_ids_by_id: HashMap<TaskId, (ResolvedTask, Vec<TaskId>)>,
```
wan9chi, line 9:
```suggestion
    resolved_tasks_and_dep_ids_by_id: HashMap<TaskId, (ResolvedTask, Vec<TaskId>)>,
```

crates/vite_task/src/config/workspace.rs
```diff
@@ -0,0 +1,482 @@
...
+            topological_run,
+        })
+    }
+
+    pub const fn cache(&self) -> &TaskCache {
+        &self.task_cache
+    }
+
+    /// Set the topological_run flag and rebuild the task graph if necessary
+    pub fn set_topological(&mut self, topological_run: bool) -> Result<(), Error> {
```
wan9chi, line 125: This method seems to be only used in tests, and it adds the complexity of maintaining the `topological` state and sync with `task_graph` with duplicated code from `load_with_cache_path`. If we are not going to use it outside of tests, I'd suggest to remove it.

crates/vite_task/src/config/task_graph_builder.rs
```diff
@@ -0,0 +1,162 @@
...
+                }
+            }
+        }
+
+        // Add topological dependencies based on package dependencies when `topological_run` is true
+        // When package A depends on package B, and both have the same task name,
+        // A#task depends on B#task
+        if topological_run {
+            for (task_id, _) in &self.resolved_tasks_and_dep_ids_by_id {
+                if let Some((package_name, task_name)) = task_id.name.split_once('#') {
```
wan9chi: This may get the wrong result if the package name contains `#`.

`task_id.name` was designed for these purposes:
 - Part of the cache key that uniquely identifies a task
 - For displaying in the UI
 - Matching against user-supplied globs (like `pkg#test:*`)

In summary, it is an opaque string that's not supposed to be split back to `package_name` and `task_name`. If we need these names, we should move the code to an eariler stage where `package_name` and `task_name` are not yet concatenated into `task_id.name`.

In this case, I'd suggest to move the `if topological_run { ... }` block to `Workspace::load_with_cache_path`, right after `// Load tasks from vite-task.json` and `// Load tasks from package.json scripts`. There we have all task edge discovery in the same place, and `TaskGraphBuilder` will stay unaware of the topological setting and the package graph.

crates/vite_task/src/config/workspace.rs
Brooooooklyn: `set_topological` is intended for the future TUI/Devtool, allowing developers to dynamically change the `topological` using a long-running Vite-task daemon.

Issue #34, opened 2025-08-06
Refactor how topological option works


wan9chi:
https://github.com/voidzero-dev/vite-plus/pull/32#discussion_r2253512114

1. build workspace with full task graph. Graph node: `ResolvedTask`, Graph edge: `{ introduced_by_topological: bool }`.
2. Render workspace to UI.
3. Whenever user select tasks or toggle topological options,  build a sub task graph out of the full task graph by filtering names and edges.

PR #36, opened 2025-08-06
refactor(task): redesign cache key and TaskId for cache sharing

### LINKED ISSUE #35: Redesign cache key and `TaskId`
...so that `echo a` in three tasks below can share the same cache.
```json
{
  "scripts": {
     "a": "echo a",
     "b": "echo a && echo b",
     "c": "vite run a && echo c"
}
```

Also ensures names in `TaskId` are opaque:
https://github.com/voidzero-dev/vite-plus/pull/32#discussion_r2253564323
https://github.com/voidzero-dev/vite-plus/blob/0e6a883784edbb5d9f102956c5e7c2eab47e4db1/crates/vite_task/src/config/workspace.rs#L436

Brooooooklyn:
- Redesigned cache key to use package name + command fingerprint instead of TaskId
- Added package_name and task_name fields to TaskId struct for better encapsulation
- Renamed TaskId::new to TaskId::parse for user input, added new TaskId::new for known values
- Fixed handling of package/task names containing '#' characters
- Added test for cache sharing between subtasks with identical commands

This enables cache sharing between tasks that run the same command (e.g., "echo a" in different compound commands), improving build performance in monorepos.

Closes #35

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/vite_task/src/config/workspace.rs
+++ b/crates/vite_task/src/config/workspace.rs
@@ -326,7 +326,6 @@ impl Workspace {
                         .cloned()
                         .map(|task_request| {
                             let sharp_pos = task_request.find('#');
-                            // contains multiple '#'
                             if sharp_pos == task_request.rfind('#') {
                                 let (dep_package_node_index, dep_task_name): (NodeIndex, Str) =
                                     if let Some(sharp_pos) = sharp_pos {
@@ -374,6 +373,7 @@ impl Workspace {
                                     subcommand_index: None, // Always points to the main task
                                 })
                             } else {
+                                // contains multiple '#'
                                 Err(Error::AmbiguousTaskRequest {
                                     task_request: task_request.to_string(),
                                 })


crates/vite_task/src/config/mod.rs
```diff
@@ -89,8 +90,49 @@ pub struct CommandFingerprint {
...
+    name: Str,
+    package_name: Str,
+    task_name: Str,
+    subcommand_index: Option<usize>,
+}
+
+impl TaskId {
+    pub(crate) fn new(package_name: Str, task_name: Str, subcommand_index: Option<usize>) -> Self {
+        Self {
+            name: format!("{}#{}", package_name, task_name).into(),
```
wan9chi: `package_name` could be empty. Example: https://github.com/vuejs/core/blob/c875019d49b4c36a88d929ccadc31ad414747c7b/package.json

The original logic is, for tasks in the root package, the full name should just be the task name without `#`:
https://github.com/voidzero-dev/vite-plus/blob/1aef66cbdf53b6fbe86fb27fe6f4766ad90fcbc9/crates/vite_task/src/config.rs#L309-L314

But it was deleted in https://github.com/voidzero-dev/vite-plus/pull/32/

crates/vite_task/src/config/mod.rs
```diff
@@ -89,8 +90,49 @@ pub struct CommandFingerprint {
...
+            package_name,
+            task_name,
+            subcommand_index,
+        }
+    }
+
+    pub(crate) fn parse(name: Str, subcommand_index: Option<usize>) -> Result<Self, Error> {
```
wan9chi: This is the same logic as https://github.com/voidzero-dev/vite-plus/pull/32#discussion_r2253564323. In that PR it was later changed to find the last `#` (`rfind`). Now it's changed back.

Neither is ideal. We should not split the full name. Use `package_name` and `task_name` instead.

crates/vite_task/src/cache.rs
```diff
@@ -37,9 +37,11 @@ pub struct TaskCache {
     conn: Mutex<Connection>,
 }
 
-#[derive(Debug, Hash, Encode, Decode, Serialize)]
+#[derive(Debug, Encode, Decode, Serialize)]
 pub struct TaskCacheKey {
-    pub task_id: TaskId,
+    // Package and command fingerprint for cache sharing between subtasks
+    pub package_name: Str,
```
wan9chi: `package_name` shouldn't be part of the cache key. If a package  name is changed and everything else is untouched. The cache should still be valid.

crates/vite_task/src/config/mod.rs
Brooooooklyn: The `fn parse` is used in only one place: https://github.com/voidzero-dev/vite-plus/pull/36/files#diff-b868338100854124c0da940976be2440a8f29876477e2d6dc1d0e5c0e1b4a4faR414. I can remove the `fn parse` function and inline its logic here since it shouldn't be a generic method on the `TaskId`.

crates/vite_task/src/cache.rs
Brooooooklyn: We are unsure if the `package_name` is used in the command. Since changing `package_name` is a rare occurrence in a monorepo, I believe it is safer to invalidate the cache.

wan9chi: The command fingerprint should be able to invalidate the cache if `package_name` is used in the command.

crates/vite_package_manager/src/lib.rs
```diff
@@ -159,23 +159,42 @@ impl PackageGraphBuilder {
...
+                    name: existing_package_info.package_json.name.to_string(),
+                    path1: existing_package_info.path.clone(),
+                    path2: self.graph[id].path.clone(),
+                });
+            }
+        } else {
+            // For nameless packages, use path as a unique identifier
+            // Store with a synthetic key that won't collide with real package names
+            let synthetic_key = format!("__nameless__{}", package_path).into();
+            self.id_and_deps_by_name.insert(synthetic_key, (id, deps));
```
wan9chi: I suggest we always use `path` to identify packages in `vite_package_manager`. This crate don't need to special-handle the nameless packages, or report duplicate package name errors.

crates/vite_task/docs/task-cache.md
```diff
@@ -0,0 +1,709 @@
...
+```rust
+pub struct StdOutput {
+    pub kind: OutputKind,        // StdOut or StdErr
+    pub content: MaybeString,    // Binary-safe content
+}
+
+pub enum MaybeString {
+    String(String),              // UTF-8 text
+    Binary(Vec<u8>),            // Non-UTF-8 binary data
+}
```
wan9chi: We don't do string manipulations, so no need to distinguish UTF-8 text from binary data.

crates/vite_error/src/lib.rs
```diff
@@ -61,8 +61,20 @@ pub enum Error {
...
+    PackageNotFound(String),
+
+    #[error("Task not found in workspace: {0}")]
+    TaskNotFound(String),
+
+    #[error("Dependency Task '{name}' not found in package located at {package_dir}")]
+    TaskDependencyNotFound { name: String, package_dir: String },
+
+    #[error("{task_request} should not contain multiple '#'")]
+    AmbiguousTaskRequest { task_request: String },
```
wan9chi, line 74: Multiple `'#'` combining with package name before `'#'` not being unique makes the implementation very complex. For now let's just disallow multiple `#` in task requests(args of `vite run` and `dependsOn`).

`'#'` is still allowed in package names. See my changes to `vite-run.md` for details.

crates/vite_task/src/config/workspace.rs
```diff
@@ -121,49 +121,6 @@ impl Workspace {
         &self.task_cache
     }
```
wan9chi, line 123: [#34](https://github.com/voidzero-dev/vite-plus/issues/34) to be done

crates/vite_task/src/config/mod.rs
```diff
@@ -53,12 +56,42 @@ pub struct ViteTaskJson {
...
+        }
+    }
+
+    pub fn matches(&self, task_request: &str) -> bool {
+        if self.name.subcommand_index.is_some() {
+            // never match non-last subcommand
+            return false;
+        }
+        let package_name = self.name.package_name.as_str();
+        // TODO: match tasks in current package if the task_request doesn't contain '#'
```
wan9chi, line 82: [#41](https://github.com/voidzero-dev/vite-plus/issues/41)

crates/vite_task/src/config/mod.rs
```diff
@@ -53,12 +56,42 @@ pub struct ViteTaskJson {
...
+        task_request.get(..package_name.len()) == Some(package_name)
+            && task_request.get(package_name.len()..=package_name.len()) == Some("#")
```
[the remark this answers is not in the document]
wan9chi: There's one more check at line 85. If package_name is 'app' and task_request is 'application#build', `task_request.get(package_name.len() + 1..)` would be 'ication#build', it won't match `Some(&self.name.task_group_name)`.

crates/vite_task/src/config/workspace.rs
```diff
@@ -237,182 +204,222 @@ impl Workspace {
...
+                    let deps: HashSet<TaskId> = task_config_json
                         .depends_on
                         .iter()
                         .cloned()
-                        .map(|name| TaskId { name, subcommand_index: None })
-                        .collect();
+                        .map(|task_request| {
+                            let sharp_pos = task_request.find('#');
+                            // contains multiple '#'
+                            if sharp_pos == task_request.rfind('#') {
```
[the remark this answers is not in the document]
wan9chi: Comment corrected in 14db236

crates/vite_task/src/config/workspace.rs
```diff
@@ -524,49 +533,48 @@ impl Workspace {
...
-        let package = &package_graph[node_idx];
-
-        // Check all dependencies from package.json
-        for dep_name in package.package_json.dependencies.keys() {
-            result.push(dep_name.to_string());
+    if let Some(&node_idx) = package_name_to_node.get(package_path) {
+        // Check all dependencies from the package from
+        for dep_index in package_graph.neighbors(node_idx) {
+            let dep_path = package_graph[dep_index].path.as_str();
+            result.push(dep_path.into());
```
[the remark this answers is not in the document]
wan9chi: Issue logged at #42. Will fix it in a separate PR.

PR #38, opened 2025-08-13
feat: support vite lint/build/test

Brooooooklyn:


packages/cli/src/vite.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Resolves the Vite binary path and environment variables.
+ *
+ * @returns Promise containing:
+ *   - binPath: Absolute path to the Vite CLI entry point (vite.js)
+ *   - envs: Environment variables to set when executing Vite
+ *
+ * The function uses require.resolve to find the vite package installation,
+ * then constructs the path to the CLI binary within the package.
+ */
+export async function vite(): Promise<{
```
fengmk2, line 27: Reuse the 'JsCommandResolvedResult' type definition in binding? This eliminates the need to repeat it several times.

```ts
import type { JsCommandResolvedResult } from '../binding/index.js';
```

fengmk2: No, there will be a circular citation problem.

PR #43, opened 2025-08-21
feat(pm): support vite install

fengmk2:
close ENG-1180

🤖 Code with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/vite_package_manager/src/package_manager.rs
+++ b/crates/vite_package_manager/src/package_manager.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::fs::{self, File};
-use std::io::BufReader;
+use std::io::{BufReader, Seek, SeekFrom};
 use std::path::{Path, PathBuf};
 use std::{env, fmt};
 
@@ -169,14 +169,16 @@ pub fn find_package_root<'a>(original_cwd: &'a Path) -> Result<PackageRoot<'a>, 
 /// The workspace file.
 ///
 /// - `PnpmWorkspaceYaml` is the pnpm workspace file.
+/// - `NpmWorkspaceJson` is the package.json file of a yarn/npm workspace.
 /// - `NonWorkspacePackage` is the package.json file of a non-workspace package.
 #[derive(Debug)]
 pub enum WorkspaceFile {
     /// The pnpm-workspace.yaml file of a pnpm workspace.
     PnpmWorkspaceYaml(File),
+    /// The package.json file of a yarn/npm workspace.
+    NpmWorkspaceJson(File),
     /// The package.json file of a non-workspace package.
     NonWorkspacePackage(File),
-    // TODO(@fengmk2): other workspace file support, like yarn, npm, etc.
 }
 
 /// The workspace root directory and its workspace file.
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -33,9 +33,11 @@ cargo_common_metadata = "allow"
 
 [workspace.dependencies]
 anyhow = "1.0.98"
+atty = "0.2.14"
```
Brooooooklyn: We can use this instead https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html

fengmk2: Using AI to write code will indeed have this problem, and if you are not familiar with it, it will be very difficult to review. still need to accumulate enough experience.

crates/vite_task/src/config/mod.rs
```diff
@@ -135,6 +141,51 @@ impl ResolvedTask {
             args: args.map(|arg| arg.as_ref().into()).collect(),
             resolved_config: resolved_task_config,
             resolved_command,
+            force_refresh_cached: None,
+            replay_cached_outputs: None,
+        })
+    }
+
+    pub(crate) fn resolve_from_built_in_with_comment_result(
```
fengmk2: dry with resolve_from_built_in

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,1151 @@
...
+        target_dir.with_file_name(format!("tmp_{}_{}_{}", timestamp, rid, version,));
+
+    remove_dir_all(&target_dir_tmp).await?;
+    download_and_extract_tgz(&tgz_url, &target_dir_tmp).await?;
+
+    // rename $target_dir_tmp/package to $target_dir_tmp/{bin_name}
+    tracing::debug!("Rename package dir to {}", bin_name);
+    tokio::fs::rename(&target_dir_tmp.join("package"), &target_dir_tmp.join(&bin_name)).await?;
+
+    let file_lock = Arc::new(Mutex::new(()));
```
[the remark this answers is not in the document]
fengmk2: At present, consider multi-threaded locks, multi-process lock problems, etc., and then look at user feedback. Normally we think there will only be one VP CLI executed.

crates/vite_task/src/config/mod.rs
```diff
@@ -121,12 +153,12 @@ impl ResolvedTask {
         let resolved_command = ResolvedTaskCommand {
             fingerprint: CommandFingerprint {
                 cwd: workspace.dir.as_path().to_string_lossy().as_ref().into(),
-                command: link_task,
+                command: link_task.clone().into(),
```
fengmk2: no need

crates/vite_task/src/install.rs
```diff
@@ -0,0 +1,516 @@
...
+        } else {
+            println!("  [{}] {}", i + 1, name);
+        }
+    }
+
+    print!("\nEnter your choice (1-{}) [default: 1]: ", managers.len());
+    io::stdout().flush().unwrap();
+
+    let mut input = String::new();
+    io::stdin().read_line(&mut input).map_err(|e| Error::Io(e))?;
```
fengmk2: map_err no need

crates/vite_package_manager/src/request.rs
```diff
@@ -0,0 +1,547 @@
...
+    ) -> Result<(), Error> {
+        let target_path = target_path.as_ref();
+        tracing::debug!("Downloading {} to {:?}", url, target_path);
+
+        let response = self
+            .client
+            .get(url)
+            .send()
+            .await
+            .map_err(|e| Error::AnyhowError(anyhow::Error::msg(format!("Request failed: {}", e))))?
```
fengmk2: no need

crates/vite_package_manager/src/request.rs
```diff
@@ -0,0 +1,512 @@
+use std::path::Path;
+use std::time::Duration;
+
+use flate2::read::GzDecoder;
+use futures_util::stream::StreamExt;
+use reqwest::Response;
+use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
+use reqwest_retry::{RetryTransientMiddleware, policies::ExponentialBackoff};
```
Brooooooklyn: https://crates.io/crates/backon this crate can also retry the `reqwest`, and it can also retry other operations. If we need to support retry in other places in the future, it might be better to use this crate.

fengmk2: fixed

PR #49, opened 2025-08-25
feat(pm): add find_workspace_root and find_package_root utilities

fengmk2:
Extract workspace and package root detection logic as standalone utilities. These functions can detect:
- pnpm workspaces via pnpm-workspace.yaml
- npm/yarn workspaces via package.json workspaces field
- Regular packages via package.json

Includes comprehensive tests for all detection scenarios.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/vite_package_manager/src/package_manager.rs
+++ b/crates/vite_package_manager/src/package_manager.rs
@@ -5,22 +5,22 @@ use std::path::Path;
 use vite_error::Error;
 
 /// The package root directory and its package.json file.
-///
-/// If the package.json file is not found, the package_json field will be None.
 #[derive(Debug)]
 pub struct PackageRoot<'a> {
     pub path: &'a Path,
-    pub package_json: Option<File>,
+    pub package_json: File,
 }
 
 /// Find the package root directory from the current working directory. `original_cwd` must be absolute.
+///
+/// If the package.json file is not found, will return PackageJsonNotFound error.
 pub fn find_package_root<'a>(original_cwd: &'a Path) -> Result<PackageRoot<'a>, Error> {
     let mut cwd = original_cwd;
     loop {
         // Check for package.json
         match File::open(cwd.join("package.json")) {
             Ok(file) => {
-                return Ok(PackageRoot { path: cwd, package_json: Some(file) });
+                return Ok(PackageRoot { path: cwd, package_json: file });
             }
             Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
                 // File doesn't exist, continue searching
@@ -35,8 +35,8 @@ pub fn find_package_root<'a>(original_cwd: &'a Path) -> Result<PackageRoot<'a>, 
             // Move up one directory
             cwd = parent;
         } else {
-            // We've reached the root, return the original directory
-            return Ok(PackageRoot { path: original_cwd, package_json: None });
+            // We've reached the root, return PackageJsonNotFound error.
+            return Err(Error::PackageJsonNotFound(original_cwd.to_path_buf()));
         }
     }
 }
@@ -57,15 +57,17 @@ pub enum WorkspaceFile {
 /// The workspace root directory and its workspace file.
 ///
 /// If the workspace file is not found, but a package is found, `workspace_file` will be `NonWorkspacePackage` with the `package.json` File.
-///
-/// If neither workspace nor package is found, `workspace_file` be None, and `path` will be `original_cwd`.
 #[derive(Debug)]
 pub struct WorkspaceRoot<'a> {
     pub path: &'a Path,
-    pub workspace_file: Option<WorkspaceFile>,
+    pub workspace_file: WorkspaceFile,
 }
 
 /// Find the workspace root directory from the current working directory. `original_cwd` must be absolute.
+///
+/// If the workspace file is not found, but a package is found, `workspace_file` will be `NonWorkspacePackage` with the `package.json` File.
+///
+/// If neither workspace nor package is found, will return PackageJsonNotFound error.
 pub fn find_workspace_root<'a>(original_cwd: &'a Path) -> Result<WorkspaceRoot<'a>, Error> {
     let mut cwd = original_cwd;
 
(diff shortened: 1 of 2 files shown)


crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        // Check for package.json with workspaces field
+        let package_json_path = cwd.join("package.json");
+        if package_json_path.exists() {
+            let package_json: serde_json::Value =
+                serde_json::from_slice(&fs::read(&package_json_path)?)?;
+
+            if package_json.get("workspaces").is_some() {
+                return Ok(cwd.into());
```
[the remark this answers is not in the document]
fengmk2: I create a track issue https://github.com/voidzero-dev/vite-plus/issues/50, will handle it later.

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
```
wan9chi: `workspaces` in `package.json` is for npm/yarn and pnpm doesn't support it, right? We haven't yet implemented yarn/npm workspace. Better throw an error here, otherwise there'll be confursing error later at the package discovery stage.

crates/vite_package_manager/src/package_manager.rs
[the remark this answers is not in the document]
wan9chi: Is the suggested check necessary? I don't think we should ignore the malformed package.json and continue going up.

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,129 @@
...
+            cwd = parent;
+        } else {
+            // We've reached the root, return the original directory
+            return original_cwd.as_ref().to_path_buf();
+        }
+    }
+}
+
+/// Find the workspace root directory from the current working directory.
+pub fn find_workspace_root(original_cwd: impl AsRef<Path>) -> Result<PathBuf, Error> {
```
wan9chi: Let's also return the content of the workspace file instead of just the path, something like:
```rust
pub struct WorkspaceRoot {
   pub path: PathBuf,
   pub workspace_file: WorkspaceFile,
}
pub enum WorkspaceFile {
  PnpmWorkspaceYaml(File),
  NonWorkspacePackage(File),
}
```

We can add more variants as the we support npm/yarn.

This will allow the caller to further parse the workspace file without trying each possible path again.

wan9chi: The same goes to `find_package_root`, let's return an opened package.json `File` there:
```rust
pub struct PackageRoot {
   pub path: PathBuf,
   pub package_json: File,
}
```

We should generally avoid `Path::exists` (see the warning of [its docs](https://doc.rust-lang.org/1.89.0/std/path/struct.Path.html#method.exists)). Just open the file and check the `NotFound` error.

crates/vite_package_manager/src/package_manager.rs
fengmk2: yes, pnpm only support pnpm-workspace.yaml.
I agree, will chagne to throw error here.

crates/vite_package_manager/src/package_manager.rs
fengmk2: Great suggestion, I will improve

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// The pnpm-workspace.yaml file of a pnpm workspace.
+    PnpmWorkspaceYaml(File),
+    /// The package.json file of a non-workspace package.
+    NonWorkspacePackage(File),
+    // TODO(@fengmk2): other workspace file support, like yarn, npm, etc.
+}
+
+/// The workspace root directory and its workspace file.
+///
+/// If the workspace file is not found, the workspace_file field will be `NonWorkspacePackage`.
```
wan9chi:
```suggestion
/// If the workspace file is not found, but a package is found, `workspace_file` will be `NonWorkspacePackage` with the `package.json` File.
/// 
/// If neither workspace nor package is found, `workspace_file` be None, and `path` will be `original_cwd`.
```

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub path: PathBuf,
+    pub package_json: Option<File>,
+}
+
+/// Find the package root directory from the current working directory.
+pub fn find_package_root(original_cwd: impl AsRef<Path>) -> Result<PackageRoot, Error> {
+    let mut cwd = original_cwd.as_ref();
+    loop {
+        // Check for package.json
+        if let Ok(file) = File::open(cwd.join("package.json")) {
```
wan9chi: Better check whether the error is `NotFound` if `File::open` returns one. If `File::open` returns another kind of error such a permission error, we want to return the error instead of silently ignoring it and continuing findup.

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,179 @@
...
+            return Ok(PackageRoot { path: cwd.into(), package_json: Some(file) });
+        }
+        if let Some(parent) = cwd.parent() {
+            // Move up one directory
+            cwd = parent;
+        } else {
+            // We've reached the root, return the original directory
+            return Ok(PackageRoot { path: original_cwd.as_ref().into(), package_json: None });
```
wan9chi: Maybe we should just return `Ok(None)` (same goes for `find_workspace_root`) here and let the caller decide what to do. The caller may want to fallback to `original_cwd`, or raise an error. For the latter case, the `Path` to `PathBuf` allocation would be unnecessary.

crates/vite_package_manager/src/package_manager.rs
```diff
@@ -0,0 +1,179 @@
...
+/// The package root directory and its package.json file.
+///
+/// If the package.json file is not found, the package_json field will be None.
+#[derive(Debug)]
+pub struct PackageRoot {
+    pub path: PathBuf,
+    pub package_json: Option<File>,
+}
+
+/// Find the package root directory from the current working directory.
```
wan9chi:
```suggestion
/// Find the package root directory from the current working. `original_cwd` must be absolute.
```

If `original_cwd` is relative we won't be able to go up to the root. 

We are dealing with all kinds of absolute/relative paths in vite-plus. We should make them safe at type-level. Issue logged: https://github.com/voidzero-dev/vite-plus/issues/55. For now let's just document it (and `find_workspace_root`).

crates/vite_package_manager/src/package_manager.rs
wan9chi: Looks like the `PathBuf` allocation is unnecessary in any cases. You can try returning the borrowed path:

```
#[derive(Debug)]
pub struct PackageRoot<'a> {
    pub path: &'a Path,
    pub package_json: File,
}

pub fn find_package_root(original_cwd: &Path) -> Result<Option<PackageRoot<'_>>, Error>
```

PR #52, opened 2025-08-25
feat(task): allow to run local package tasks with vite run

Brooooooklyn:
This PR allows `vite run` to locate task names without `#` in the nearest package. 
Currently, only one task can be run if the task name does not contain `#`; providing multiple tasks in this case will result in a `OnlyOneTaskRequest` error.

### DIFF
--- a/crates/vite_task/src/config/workspace.rs
+++ b/crates/vite_task/src/config/workspace.rs
@@ -330,70 +330,16 @@ impl Workspace {
                 return Err(Error::OnlyOneTaskRequest(task_requests.join(" ")));
             }
             // For non-recursive mode, find the task in the full task graph
-            // If task doesn't contain '#', use the current package determined at load time
             for task_request in task_requests {
-                if !task_request.contains('#') {
-                    // Task without '#' - look for it in the current package
-                    let mut found_in_current = false;
-
-                    for package in self.package_graph.node_weights() {
-                        if Some(package.path.as_str()) == self.current_package_path.as_deref() {
-                            // Check if this package has the requested task
-                            let task_id_to_check = TaskId {
-                                task_group_id: TaskGroupId {
-                                    task_group_name: task_request.clone(),
-                                    package_path: package.path.clone().into(),
-                                },
-                                subcommand_index: None,
-                            };
-
-                            for (task_node_index, task) in self.task_graph.node_references() {
-                                if task.id() == task_id_to_check {
-                                    found_in_current = true;
-                                    remaining_task_node_indexes.insert(task_node_index);
-                                    break;
-                                }
-                            }
-                            break;
-                        }
+                let mut has_matched_task = false;
+                for (task_node_index, task) in self.task_graph.node_references() {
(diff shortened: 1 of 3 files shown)


crates/vite_task/src/config/workspace.rs
```diff
@@ -26,33 +28,61 @@ use super::{
...
+        } else {
+            // Get relative path from workspace root to package root
+            let path = current_package_root.strip_prefix(&workspace_root).map_err(|err| {
+                Error::PathPrefixError {
+                    err,
+                    message: "package root is not a subpath of workspace root".to_string(),
+                    path: current_package_root.to_path_buf(),
+                }
+            })?;
+            Ok((workspace_root, path.to_string_lossy().to_string()))
```
wan9chi: In the current codebase, all relative paths are `Str` which is always valid utf8. For non-utf8 relative paths we throw an error at https://github.com/voidzero-dev/vite-plus/blob/b02419e6bc7c72cdaf2292f0eb226952c4eeb6c5/crates/vite_package_manager/src/lib.rs#L273

We may discuss and change this requirement later. For now, I think `current_package_root` should also be strictly valid utf-8. For non-utf8 package path, better throw an error, or set to `None`. Lossy converstion doesn't seem right.

crates/vite_task/src/config/workspace.rs
```diff
@@ -26,33 +28,61 @@ use super::{
...
 impl Workspace {
+    /// Determines the current package path relative to the workspace root.
+    /// Returns an empty string if the current directory is the workspace root itself.
+    /// Returns the workspace root and the current package path.
+    fn determine_current_package_path(original_cwd: &Path) -> Result<(&Path, String), Error> {
+        let workspace_root = find_workspace_root(original_cwd)?.path;
+        let current_package_root = find_package_root(original_cwd)?.path;
+        if current_package_root == workspace_root {
+            // We're at workspace root with a package.json
+            Ok((workspace_root, String::new()))
```
wan9chi: Is this check necessary? I think `strip_prefix` in the `else` branch can handle two same paths correctly.

crates/vite_task/src/config/workspace.rs
```diff
@@ -26,33 +28,61 @@ use super::{
...
-    pub(crate) topological_run: bool,
 }
 
 impl Workspace {
+    /// Determines the current package path relative to the workspace root.
+    /// Returns an empty string if the current directory is the workspace root itself.
+    /// Returns the workspace root and the current package path.
+    fn determine_current_package_path(original_cwd: &Path) -> Result<(&Path, String), Error> {
+        let workspace_root = find_workspace_root(original_cwd)?.path;
+        let current_package_root = find_package_root(original_cwd)?.path;
```
wan9chi: If the current package doesn't exist, the error shouldn't be bubbled up to the user. Let's handle it as specified in https://github.com/voidzero-dev/vite-plus/blob/5339f5d9bf735f934fe534d86d50e09fff21a05f/crates/vite_task/docs/vite-run.md#default-behavior

PR #58, opened 2025-08-26
feat: add crate `vite_path`

wan9chi:
- move `Str` to a new crate `vite_str` so it can be shared for all other crates.
- add crate `vite_path`


crates/vite_path/src/absolute.rs
```diff
@@ -0,0 +1,218 @@
...
+    fn non_absolute() {
+        assert!(AbsolutePath::new(Path::new("foo/bar")).is_none())
+    }
+
+    #[test]
+    fn strip_prefix() {
+        let abs_path = AbsolutePath::new(Path::new(if cfg!(windows) {
+            "C:\\Users\\foo\\bar"
+        } else {
+            "/home/foo/bar"
```
fengmk2, line 165: if strip_prefix `"/home/foo/"` (end with `/`) can we get the same result `"foo/bar"` ?

wan9chi: Yes. test added in 1a2cd456b95d9f3cf02406f8c002fde6ee7dbb6a

crates/vite_path/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+bincode = { workspace = true }
+ref-cast = { workspace = true }
+vite_str = { workspace = true }
+thiserror = { workspace = true }
+
+[lints]
+workspace = true
+
+[dev-dependencies]
+assert2 = "0.3.16"
```
fengmk2, line 19: Should dev-dependencies also be added to `workspace.dependencies` for unified version management? I will also use `assert2` in vite install, so we can unify the version.

crates/vite_path/src/relative.rs
```diff
@@ -0,0 +1,265 @@
...
+use bincode::{Decode, Encode, de::Decoder, error::DecodeError};
+
+use ref_cast::{RefCastCustom, ref_cast_custom};
+
+/// A relative path with additional guarantees to make it portable:
+///
+/// - It is valid utf-8
+/// - It uses slashes `/` as separators, not backslashes `\`
+/// - There's no backslash `\` in components (this is valid in unix systems but not portable to Windows)
+#[derive(RefCastCustom, PartialEq, Eq)]
```
fengmk2, line 16: Can `RefCast` be used directly? This way, we wouldn't need to write unsafe code manually

crates/vite_path/src/relative.rs
```diff
@@ -0,0 +1,265 @@
...
+        self.as_relative_path()
+    }
+}
+
+/// Error when converting a path containing invalid data to `RelativePathbuf`
+#[derive(thiserror::Error, Debug)]
+pub enum InvalidPathDataError {
+    /// One of the components contains non-utf8 data.
+    #[error("path is not portable because contains non-utf8 data")]
+    NonUtf8,
```
Brooooooklyn, line 156: Does this mean that Vite Plus cannot operate in any directory containing non-UTF8 characters, even if the user is not care about portable cache?

crates/vite_path/src/relative.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// One of the components contains backslashes `\`.
+    ///
+    /// This is valid in unix systems but not portable to Windows
+    #[error("path is not portable because it contains backslash ('\\') in its components")]
+    BackslashInComponent,
```
Brooooooklyn, line 161: The same problem as NonUtf8: if I have a file like `a\\b.txt`, and I’m not using Windows or portable cache at all, vite plus still cannot run properly

crates/vite_path/Cargo.toml
Brooooooklyn: Let me create a PR to move all dependencies back into the workspace after the Vite installation is complete.

PR #69, opened 2025-08-29
fix(task): cwd of builtin commands

Brooooooklyn:


crates/vite_task/src/execute.rs
```diff
@@ -237,10 +236,12 @@ impl TaskEnvs {
         let env_path =
             all_envs.entry("PATH".into()).or_insert_with(|| Arc::<OsStr>::from(OsStr::new("")));
         let paths = split_paths(env_path);
-        let node_modules_bin = base_dir.join(&task.config.cwd).join("node_modules/.bin");
+        let node_modules_bin = base_dir.join(task.config.cwd.as_path()).join("node_modules/.bin");
         *env_path = join_paths(
             iter::once(node_modules_bin)
-                .chain(iter::once(base_dir.join(&task.config_dir).join("node_modules/.bin")))
+                .chain(iter::once(
+                    base_dir.join(task.config_dir.as_path()).join("node_modules/.bin"),
```
fengmk2: Is `base_dir.join(task.config.cwd.as_path())` the same as `base_dir.join(task.config_dir.as_path())`? Should we check if they are the same to avoid adding them repeatedly?

crates/vite_task/src/execute.rs
```diff
@@ -237,10 +236,12 @@ impl TaskEnvs {
         let env_path =
             all_envs.entry("PATH".into()).or_insert_with(|| Arc::<OsStr>::from(OsStr::new("")));
         let paths = split_paths(env_path);
-        let node_modules_bin = base_dir.join(&task.config.cwd).join("node_modules/.bin");
+        let node_modules_bin = base_dir.join(task.config.cwd.as_path()).join("node_modules/.bin");
```
fengmk2: the base_dir is the workspace_root dir? Should we rename it?

crates/vite_path/src/relative.rs
```diff
@@ -101,6 +132,14 @@ impl<'a, Context> Decode<Context> for RelativePathBuf {
     }
 }
 
+impl<'de, Context> BorrowDecode<'de, Context> for RelativePathBuf {
+    fn borrow_decode<D: BorrowDecoder<'de, Context = Context>>(
+        decoder: &mut D,
+    ) -> Result<Self, DecodeError> {
+        Ok(Self(Str::borrow_decode(decoder)?))
+    }
+}
```
wan9chi: This doesn't do the same checks as the `Decode` impl. 


Instead of manual `BorrowDecoder` impl ,`bincode::impl_borrow_decode!(RelativePathBuf)` should be enough.

crates/vite_task/docs/task-cache.md
```diff
@@ -99,7 +99,7 @@ The command fingerprint captures the complete execution context:
 
 ```rust
 pub struct CommandFingerprint {
-    pub cwd: Str,                                      // Working directory
+    pub cwd: Str,                                      // Working directory, relative to workspace root
```
wan9chi: There are so many places where `Str` can be replaced with `RelativePathBuf`, like this one. Feel free to do it in  this PR or I'll do it in a separate one.

In principle, `Str` is always `RelativePathBuf` and `PathBuf` is most likely `AbsolutePathBuf`.

Ideally we convert `std::path` types to `vite_path` types right after we get `Path`/`PathBuf` from std, and convert back right before we pass a path to std.

crates/vite_task/src/config/workspace.rs
```diff
@@ -109,8 +110,15 @@ impl Workspace {
 
         Ok(Self {
             package_graph: Graph::new(),
-            workspace_dir: workspace_root.to_path_buf(),
-            cwd,
+            workspace_dir: AbsolutePathBuf::new(workspace_root.to_path_buf())
+                .ok_or_else(|| Error::PathIsNotAbsolute(workspace_root.to_path_buf()))?,
+            cwd: RelativePathBuf::try_from(cwd.strip_prefix(workspace_root).map_err(|_| {
```
wan9chi: `cwd` and `workspace_root` should be `AbsolutePathBuf` in the first place, and then we can use `AbsolutePath::strip_prefix` to get a `RelativePath`

crates/vite_error/src/lib.rs
```diff
@@ -107,6 +107,15 @@ pub enum Error {
...
 
+    #[error(transparent)]
+    FromPathError(#[from] vite_path::relative::FromPathError),
+
+    #[error("Path is not valid utf8: {0:?}")]
+    PathNotValidUtf8(AbsolutePathBuf),
+
+    #[error("No package.json found at {0:?}")]
+    NoPackageJsonFound(AbsolutePathBuf),
+
```
wan9chi: They're not used anywhere
```suggestion

```

crates/vite_path/src/absolute.rs
```diff
@@ -44,6 +44,11 @@ impl AbsolutePath {
         &self.0
     }
 
+    /// Gets the underlying [`Path`] as a [`PathBuf`]
+    pub fn to_path_buf(&self) -> PathBuf {
+        self.0.to_path_buf()
+    }
```
wan9chi: It's not used. I suggest to remove it to discourage AI from falling back to `PathBuf`. `as_path` and `AsRef<Path>` should be enough to interact with std/tokio fs functions.
```suggestion

```

crates/vite_path/src/relative.rs
```diff
@@ -46,6 +45,9 @@ impl RelativePath {
     pub fn as_path(&self) -> &Path {
         Path::new(self.as_str())
     }
+    pub fn to_path_buf(&self) -> PathBuf {
+        PathBuf::from(self.as_str())
+    }
```
wan9chi:
```suggestion
```

crates/vite_task/src/config/mod.rs
```diff
@@ -120,15 +120,13 @@ impl ResolvedTask {
             program: bin_path.into(),
         });
         let task_config: TaskConfig = builtin_task.clone().into();
-        let resolved_task_config = ResolvedTaskConfig {
-            // TODO: config_dir should be cwd
-            config_dir: RelativePathBuf::default(),
-            config: task_config,
-        };
+        let cwd = workspace.cwd.strip_prefix(&workspace.workspace_dir)?.unwrap_or_default();
```
wan9chi: Better to raise a "cwd outside the workspace root" error than sliently force the cwd to be workspace root

wan9chi: Or, if possible, just let `workspace.cwd` be a `RelativePathBuf` relative to `workspace.workspace_dir`.

PR #71, opened 2025-08-29
feat(cli): release local cli

fengmk2:


.github/workflows/release.yml
```diff
@@ -130,14 +130,21 @@ jobs:
           pattern: bindings-*
           merge-multiple: true
 
+      - name: Copy *.node to cli/binding
```
Brooooooklyn: https://github.com/voidzero-dev/vite-plus/blob/main/.github/workflows/release.yml#L127-L131 We are now copying all platform .node files to the dist directory through the download-artifacts action, so the binding directory should not be published together and can be bundled in.

PR #99, opened 2025-09-11
ci: always install zig in release workflow

Brooooooklyn:


packages/global/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@voidzero-dev/global",
-  "version": "0.0.0",
+  "version": "0.0.0-7dcab9b3dd6074ba920256fb67632f2009285760",
```
fengmk2: no need

PR #100, opened 2025-09-12
fix(task): propagate task exit codes to CLI process

fengmk2:
### TL;DR

Propagate task execution exit codes to the CLI, ensuring failed tasks return appropriate error codes.

### What changed?

- Modified the execution flow to return exit status from tasks instead of just success/failure
- Updated all command handlers (lint, fmt, test, etc.) to propagate exit status
- Changed the main CLI entry point to exit with the appropriate code when tasks fail
- Added exit status to debug logs for better troubleshooting
- Created a new test case to verify exit code propagation behavior

### How to test?

1. Run a task that succeeds: `vite run script1` - should exit with code 0
2. Run a task that fails: `vite run script2` - should exit with the same code as the failed task
3. Try the new test case: `cd packages/cli/tests/cases/exit-code && vite-plus run script2` - should exit with code 1

### Why make this change?

Previously, when a task failed, the CLI would still exit with code 0, which is misleading for CI/CD pipelines and scripts that rely on exit codes to determine success or failure. This change ensures that the CLI properly propagates the exit code of failed tasks, making it more reliable for automation and consistent with standard CLI behavior.

### DIFF
--- a/crates/vite_task/src/schedule.rs
+++ b/crates/vite_task/src/schedule.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{process::ExitStatus, sync::Arc};
 
 use futures_core::future::BoxFuture;
 use futures_util::future::FutureExt as _;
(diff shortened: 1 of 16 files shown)


crates/vite_task/src/schedule.rs
```diff
@@ -67,17 +67,27 @@ impl ExecutionPlan {
     /// 2. If cache hit: replay the cached output
     /// 3. If cache miss: execute the task and cache the result
     #[tracing::instrument(skip(self, workspace))]
-    pub async fn execute(self, workspace: &mut Workspace) -> anyhow::Result<()> {
+    pub async fn execute(self, workspace: &mut Workspace) -> Result<(), Error> {
```
wan9chi: I don't think non-zero exit code can be considered an error in the task runner. Better return it in `Ok`.

packages/cli/tests/cases/exit-code/snap.txt
```diff
@@ -0,0 +1,23 @@
+> vite-plus run script1 # script1 run, create the cache and should be success
+Cache not found
+success
+
+> exit code: 0
```
wan9chi: This is a little bit confusing as it looks the same as a new command. I suggest putting it in front of the command line, like `[1]>vite-plus run`. Also, consider hiding it for success commands (which are most common) to make the snap file clean.

crates/vite_task/src/schedule.rs
fengmk2: If not treated as an error, then every place that calls execute needs to return exit_status to let its caller recursively handle a lot of repetitive code.

fengmk2: changed

packages/cli/tests/main.ts
```diff
@@ -46,9 +46,14 @@ function runTestCase(name: string) {
   const newSnap: string[] = [];
 
   for (const command of steps.commands) {
-    newSnap.push(`> ${command}`);
-    const output = cp.execSync(command, { env, cwd: caseTmpDir, encoding: 'utf-8' });
-    newSnap.push(output);
+    try {
+      const output = cp.execSync(command, { env, cwd: caseTmpDir, encoding: 'utf-8' });
+      newSnap.push(`[0]> ${command}`);
```
wan9chi: Consider hiding `[0]` for success commands (which are most common) to make the snap file clean.

packages/cli/tests/main.ts
```diff
@@ -46,9 +46,14 @@ function runTestCase(name: string) {
...
+      newSnap.push(output);
+    } catch (error) {
+      newSnap.push(`[${error.status}]> ${command}`);
+      newSnap.push(error.output.filter((line: string) => line).join('\n'));
```
wan9chi: Why filter empty lines in error cases?

fengmk2: output will contains null value, need to ignore it

wan9chi: Empty strings are also falsy but shouldn't be filtered out. Maybe we need to explicitly check for `null`?

PR #106, opened 2025-09-13
feat(install): add install shortcut command

fengmk2:
### TL;DR

Added an alias `i` for the `install` command and improved argument handling for hyphenated values.

### What changed?

- Enhanced the `install` command in `vite_task` crate:
  - Added documentation comments
  - Added `disable_help_flag = true` and an alias `i` for the command
  - Replaced `last = true` with `allow_hyphen_values = true` and `trailing_var_arg = true` to properly handle hyphenated arguments
- Added test cases for the `install` command:
  - Created a test for the `i` shortcut with pnpm
  - Added test for yarn package installation
  - Added snapshots to validate expected output

### How to test?

1. Run `vite-plus i` to verify the shortcut works
2. Run `vite-plus install -D tslib@2.8.1` to verify installing a package with flags works correctly
3. Test with different package managers (pnpm, yarn) to ensure compatibility

### Why make this change?

The previous implementation didn't properly handle hyphenated arguments for the install command, which prevented users from using common package manager flags like `-D` (for dev dependencies). This change ensures that all arguments are correctly passed to the underlying package manager and adds a convenient shortcut alias for the install command.

### DIFF
--- a/packages/cli/snap-tests/npm-install-with-options/snap.txt
+++ b/packages/cli/snap-tests/npm-install-with-options/snap.txt
@@ -1,0 +1,44 @@
+> vite-plus install --help # print help message
+Cache not found
+Install a package
+
+Usage:
+npm install [<package-spec> ...]
+
+Options:
+[-S|--save|--no-save|--save-prod|--save-dev|--save-optional|--save-peer|--save-bundle]
+[-E|--save-exact] [-g|--global]
+[--install-strategy <hoisted|nested|shallow|linked>] [--legacy-bundling]
+[--global-style] [--omit <dev|optional|peer> [--omit <dev|optional|peer> ...]]
+[--include <prod|dev|optional|peer> [--include <prod|dev|optional|peer> ...]]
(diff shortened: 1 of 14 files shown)


packages/cli/snap-tests/command-install/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "command-install-test",
```
fengmk2: need to add yarn and npm tests

packages/cli/snap-tests/npm-install-with-options/snap.txt
```diff
@@ -0,0 +1,44 @@
...
+Cache not found
+
+added 1 package, and audited 2 packages in <variable>ms
+
+found 0 vulnerabilities
+
+npm warn config production Use `--omit=dev` instead.
+
+> ls node_modules
+@oxlint
```
fengmk2, line 34: It's amazing that npm has such a basic bug, creating a scoped directory for dependencies without installing any packages

PR #116, opened 2025-09-15
test: move snap test to tools and init snap test for global cli

wan9chi:
# Migrate CLI tests to snap-test tool

This PR moves the CLI snapshot tests from a custom implementation to a dedicated `snap-test` tool in the tools package:

- Move test cases from `packages/cli/tests/cases` to `packages/cli/snap-tests`
- Create a new `snap-test` binary in the tools package
- Update the CLI's test script to use the new `snap-test` tool
- Add snapshot tests for the global package
- Update CI workflow to use the bootstrap-cli script and run tests for both CLI and global packages

The new approach provides a more consistent testing experience across packages and simplifies the test runner implementation.

### DIFF
--- a/packages/tools/src/snap-test.ts
+++ b/packages/tools/src/snap-test.ts
@@ -45,7 +45,7 @@ function runTestCase(name: string) {
   env['PATH'] = [
     ...env['PATH']!.split(path.delimiter),
     // Extend PATH to include the package's bin directory
-    path.join(path.dirname(import.meta.dirname), 'bin')
+    path.resolve('bin')
   ].join(path.delimiter);
 
   const newSnap: string[] = [];


packages/tools/src/snap-test.ts
```diff
@@ -42,10 +44,8 @@ function runTestCase(name: string) {
   };
   env['PATH'] = [
     ...env['PATH']!.split(path.delimiter),
-    // Extend PATH to include the CLI's bin directory
-    path.join(path.dirname(import.meta.dirname), 'bin'),
-    // Also include node_modules/.bin for local dev dependencies
-    path.join(path.dirname(import.meta.dirname), 'node_modules/.bin'),
+    // Extend PATH to include the package's bin directory
+    path.resolve('bin')
```
fengmk2, line 48: It should still be necessary to add `path.resolve('node_modules/.bin')`, otherwise using `node /path/to/snap-test.ts` directly will fail to find commands like `json-edit`.

packages/global/package.json
```diff
@@ -27,7 +28,8 @@
   },
   "devDependencies": {
     "tsdown": "catalog:",
-    "@voidzero-dev/vite-plus": "workspace:*"
+    "@voidzero-dev/vite-plus": "workspace:*",
+    "@voidzero-dev/vite-plus-tools": "workspace:"
```
fengmk2, line 32: should add @voidzero-dev/vite-plus-tools to local cli devDependencies too?

packages/tools/src/snap-test.ts
fengmk2: oh json-edit should work

PR #128, opened 2025-09-16
feat(task): add NO_COLOR environment variable support

fengmk2:
# Add support for NO_COLOR environment variable

This PR adds support for the [NO_COLOR](https://no-color.org/) environment variable, which is a standard for disabling colored output in command-line tools.

- Added `NO_COLOR` to the list of passthrough environment variables
- Modified the color detection logic to respect `NO_COLOR` when present
- Added tests to verify that `NO_COLOR` is properly passed through and respected
- Created a new snap test to verify `NO_COLOR` behavior in the CLI
- Set `NO_COLOR=true` in snap tests to ensure consistent output

When `NO_COLOR` is set, the code now avoids setting `FORCE_COLOR`, ensuring that color output is properly disabled in subprocesses.

### DIFF
--- a/crates/vite_task/src/execute.rs
+++ b/crates/vite_task/src/execute.rs
@@ -539,7 +539,7 @@ mod tests {
         assert!(result3.all_envs.contains_key("NO_COLOR"));
         let no_color_value = result3.all_envs.get("NO_COLOR").unwrap();
         assert_eq!(no_color_value.to_str().unwrap(), "1");
-        // FORCE_COLOR should not automatically added since NO_COLOR is set
+        // FORCE_COLOR should not be automatically added since NO_COLOR is set
         assert!(!result3.all_envs.contains_key("FORCE_COLOR"));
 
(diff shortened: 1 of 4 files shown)


packages/tools/src/snap-test.ts
```diff
@@ -41,11 +41,14 @@ function runTestCase(name: string) {
     ...steps.env,
     // Indicate CLI is running in test mode
     VITE_PLUS_CLI_TEST: '1',
+    NO_COLOR: 'true',
+    // set CI=true make sure snap-tests are stable on GitHub Actions
+    CI: 'true',
```
wan9chi, line 53: Are these for the new case or all the cases. If it's just for the new case they should go to `steps.json`.

fengmk2: These are set by default for all cases to align with the default CI environment configuration. I added a case to ensure these environment variables persist.

PR #129, opened 2025-09-16
feat(tools): ignore pnpm install unstable output

fengmk2:
This PR adds Vitest for unit testing alongside the existing snapshot tests. Key changes:

1. Added a new `vitest.config.ts` file to configure test discovery
2. Created a utility function `replaceUnstableOutput` to normalize test output by replacing unstable values like:
    - Semver versions
    - Timestamps
    - File paths
    - PNPM output patterns
    - Current working directory
3. Renamed the existing snapshot test commands from `test` to `snap-test` in package.json files
4. Updated the CI workflow to use the new `snap-test` command
5. Modified the root `test` command to run both Vitest tests and snapshot tests

The PR includes a test case for the new utility function with appropriate snapshots.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "install-global-cli": "npm install -g ./packages/global",
     "typecheck": "tsc -b tsconfig.json",
     "lint": "vite lint",
-    "test": "pnpm -r test",
+    "test": "vite test && pnpm -r snap-test",
     "prepare": "husky"
   },
   "devDependencies": {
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "install-global-cli": "npm install -g ./packages/global",
     "typecheck": "tsc -b tsconfig.json",
     "lint": "vite lint",
-    "test": "pnpm -r test",
+    "test": "vitest run && pnpm -r snap-test",
```
fengmk2: should use `vite test`

PR #131, opened 2025-09-16
feat(binding): read vite config from workspace root

Brooooooklyn:
This PR adds basic support for reading `vite.config.*` files from the workspace root and passes the `lint` and `fmt` configurations to the Rust side. 

This is a temporary solution to enable support for `lint` and `fmt` settings in `vite.config.*`. We need to pass the JavaScript configuration object to `oxlint` and `oxfmt` due to the use of JS plugins.

### DIFF
--- a/packages/cli/snap-tests/workspace-root-vite-config/snap.txt
+++ b/packages/cli/snap-tests/workspace-root-vite-config/snap.txt
@@ -46,4 +46,4 @@ Checking formatting...
 packages/app-b/index.js (<variable>ms)
 
 Format issues found in above 1 files. Run without `--check` to fix.
-Finished in 3ms on 1 files using 16 threads.
+Finished in <variable>ms on 1 files using <variable> threads.
--- a/packages/tools/src/snap-test.ts
+++ b/packages/tools/src/snap-test.ts
@@ -88,6 +88,7 @@ async function runTestCase(name: string) {
 }
 
 function replaceUnstableOutput(output: string) {
-  return output.replace(/\d+(?:\.\d+)?s|\d+ms/, '<variable>ms')
-    .replace(/with \d+ rules using \d+ threads/, 'with <variable> rules using <variable> threads');
+  return output.replaceAll(/\d+(?:\.\d+)?s|\d+ms/g, '<variable>ms')
+    .replaceAll(/with \d+ rules/g, 'with <variable> rules')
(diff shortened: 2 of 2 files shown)


packages/cli/build.ts
```diff
@@ -0,0 +1,28 @@
...
+await build({
+  input: ['./src/bin.ts', './src/index.ts'],
+  external: [/^node:/, 'rolldown-vite'],
+  output: {
+    format: 'esm',
+  },
+});
+
+if (output) {
+  await copyFile(output.path, `./dist/${parse(output.path).base}`);
```
fengmk2: Great, no more manual copying is needed [nicerocket](https://app.graphite.dev/settings/meme-library?org=voidzero-dev)

packages/cli/package.json
```diff
@@ -19,9 +19,7 @@
     "node": ">=20.11.0"
   },
   "scripts": {
-    "build": "pnpm run build:binding && pnpm run build:ts",
-    "build:binding": "napi build --package-json-path ../package.json --cwd binding --platform --release --esm",
-    "build:ts": "tsdown",
+    "build": "oxnode ./build.ts",
```
fengmk2: Is it because Node v20 doesn't support running ts directly?

packages/cli/src/index.ts
```diff
@@ -8,20 +10,77 @@ import {
...
+type UniversalResolvedConfig = ResolvedConfig & {
+  lint?: LintConfig;
+  fmt?: FmtConfig;
+};
+
 function defineConfig(
-  config: UserConfig & { extends?: string; test?: ViteUserConfig['test'] },
+  config: UserConfig & {
+    extends?: string;
+    test?: ViteUserConfig['test'];
```
fengmk2: after merge this task should be close too https://linear.app/voidzero/issue/ENG-1458/support-vitestnpmvoidzero-devvite-plus-alias

packages/cli/package.json
Brooooooklyn: because we have oxnode, and I want to dogfood it!

packages/cli/src/index.ts
Brooooooklyn: we need to re-export more stuffs from `rolldown-vite` `vitest` `oxlint` and `oxfmt` for this task

PR #142, opened 2025-09-17
feat(cli): export `vite/client.d.ts`

fengmk2:
### TL;DR

Added client module with TypeScript type definitions to support `import 'vite/client'` pattern.

### What changed?

- Created a new `client.ts` file that imports from 'rolldown-vite/client'
- Updated the build script to include client.ts in the TypeScript compilation
- Added client module to package.json exports to expose the type definitions
- Fixed CI workflow by adding explicit `pnpm install` step before bootstrap-cli
- Enhanced lint command to include type-aware linting and typecheck

### How to test?

1. Build the CLI package
2. Verify that `client.d.ts` is correctly generated in the dist folder
3. Import from `vite/client` in a TypeScript project and confirm type definitions are properly loaded

### Why make this change?

This change enables TypeScript users to get proper type definitions when importing from the client module, matching the existing pattern in Vite. By exposing the client types through the package exports, TypeScript projects will have better developer experience with proper type checking and autocompletion when using the `import 'vite/client'` pattern.

### DIFF
--- a/packages/cli/build.ts
+++ b/packages/cli/build.ts
@@ -59,8 +59,12 @@ const { options } = parseJsonSourceFileConfigFileContent(tsconfig, sys, projectD
 
 const host = createCompilerHost(options);
 
+const srcDir = join(projectDir, 'src');
 const program = createProgram({
-  rootNames: [join(projectDir, 'src', 'index.ts')],
+  rootNames: [
+    join(srcDir, 'index.ts'),
+    join(srcDir, 'client.ts'),
+  ],
   options: {
     ...options,
     emitDeclarationOnly: true,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,6 +146,7 @@ jobs:
         run: |
           pnpm test
           git diff --exit-code
+
   install-e2e-test:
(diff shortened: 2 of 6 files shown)


packages/cli/build.ts
```diff
@@ -40,4 +40,5 @@ await build({
 
 if (output) {
   await copyFile(output.path, `./dist/${parse(output.path).base}`);
+  await copyFile('./src/client.d.ts', `./dist/client.d.ts`);
```
fengmk2: use better way https://app.graphite.dev/github/pr/voidzero-dev/vite-plus/144/ci-reuse-the-cli-dist-in-release-pipeline?utm_source=chrome-extension#thread-PRRT_kwDOOELNJM5a0_Dc

packages/global/package.json
```diff
@@ -19,7 +19,8 @@
   ],
   "scripts": {
     "build": "rolldown -c rolldown.config.ts",
-    "snap-test": "snap-test"
+    "snap-test": "snap-test",
+    "typecheck": "tsc --noEmit"
```
fengmk2: no need

PR #170, opened 2025-09-21
fix: report cache status from nested runners

wan9chi:
# Improve nested task runner display

This PR enhances how nested task runners are displayed in the terminal:

1. Inner runners (like `vite lint`, `vite run`, etc.) now handle displaying their own command and cache status
2. Added environment variable `VITE_OUTER_COMMAND` to track and display the original command
3. Added helper method `has_inner_runner()` to detect commands that start inner runners

These changes provide a cleaner, more intuitive output when running nested tasks, showing proper cache status information without redundant command displays.

### DIFF
--- a/crates/vite_task/src/config/task_command.rs
+++ b/crates/vite_task/src/config/task_command.rs
@@ -56,7 +56,7 @@ impl TaskCommand {
         let Some(subcommand) = parsed_command.args.first() else {
             return false;
         };
-        matches!(subcommand.as_str(), "run" | "lint" | "fmt" | "build" | "test")
+        matches!(subcommand.as_str(), "run" | "lint" | "fmt" | "build" | "test" | "lib")
     }
 }
 


crates/vite_task/src/config/task_command.rs
```diff
@@ -44,6 +44,20 @@ impl TaskCommand {
...
+        let Self::Parsed(parsed_command) = self else {
+            return false;
+        };
+        if parsed_command.program != "vite" {
+            return false;
+        }
+        let Some(subcommand) = parsed_command.args.first() else {
+            return false;
+        };
+        matches!(subcommand.as_str(), "run" | "lint" | "fmt" | "build" | "test")
```
fengmk2: add `lib` built-in command for tsdown

wan9chi: 477611b0edbb8422131c17b19c931de1efe2fde3

PR #178, opened 2025-09-22
chore(cli): add fmt to ready task

fengmk2:


packages/global/snap-tests/new-monorepo/snap.txt
```diff
@@ -52,6 +52,34 @@ pnpm-workspace.yaml
...
+    "overrides": {
+      "vite": "catalog:",
+      "tsdown": "catalog:"
+    }
+  },
+  "resolutions": {
+    "vite": "catalog:",
+    "tsdown": "catalog:"
+  },
+  "packageManager": "pnpm@10.17.0"
```
fengmk2: should replace version here

PR #202, opened 2025-09-25
feat(task): support fast failure exit

fengmk2:
### TL;DR

Implement fail-fast behavior in the task execution pipeline to skip remaining tasks when a dependency fails.

### What changed?

- Modified `ExecutionPlan::execute` to track task failures and skip subsequent tasks
- Added a new `ExecutionFailure::SkippedDueToFailedDependency` status for skipped tasks
- Created a new snap test in `packages/cli/snap-tests/fail-fast/` to verify the behavior
- The test demonstrates that when a task fails, all dependent tasks are skipped with a clear indication in the execution summary

### How to test?

Run the new snap test in the `fail-fast` directory:

```bash
cd packages/cli/snap-tests/fail-fast
vite run test
vite run ready
```

The output should show that when `script2` fails, `script3` is skipped. Similarly, in the nested case, when `test` fails (due to `script2` failing), `script4` is also skipped.

### Why make this change?

This change improves the developer experience by not wasting time running tasks that are guaranteed to fail due to a dependency failure. It also provides clear feedback about why certain tasks were skipped, making it easier to diagnose and fix issues in the task pipeline.

### DIFF
--- a/crates/vite_task/src/schedule.rs
+++ b/crates/vite_task/src/schedule.rs
@@ -32,6 +32,7 @@ pub struct PreExecutionStatus {
     pub cache_status: CacheStatus,
     pub display_options: DisplayOptions,
 }
+
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub enum CacheStatus {
     /// Cache miss with reason.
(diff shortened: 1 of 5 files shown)


packages/cli/snap-tests/fail-fast/snap.txt
```diff
@@ -0,0 +1,61 @@
...
+Task Details:
+────────────────────────────────────────────────
+  [1] script1: $ echo 'success 1' ✓
+      → Cache miss: no previous cache entry found
+  ·······················································
+  [2] script2: $ node failure.js ✗ (exit code: 1)
+      → Cache miss: no previous cache entry found
+  ·······················································
+  [3] test: $ vite run script3 ⊘ (skipped: dependency failed)
+      → Cache miss: no previous cache entry found
```
wan9chi, line 27: It's weird that skipped tasks still report cache status, since they don't look up cache at all.

crates/vite_task/src/schedule.rs
```diff
@@ -127,8 +128,31 @@ impl ExecutionPlan {
     #[tracing::instrument(skip(self, workspace))]
     pub async fn execute(self, workspace: &mut Workspace) -> Result<ExecutionSummary, Error> {
         let mut execution_statuses = Vec::<ExecutionStatus>::with_capacity(self.steps.len());
+        let mut has_failed = false;
         for step in self.steps {
-            execution_statuses.push(Self::execute_resolved_task(step, workspace).await?);
+            if has_failed {
```
wan9chi, line 133: This incorrectly skips tasks that are not yet executed but do not depend on the failed tasks.

fengmk2: Yes, it looks like it can't be handled simply and needs to be determined based on the task_graph

PR #230, opened 2025-10-16
style: fix clippy::redundant_closure_for_method_calls warnings

Brooooooklyn:
Replace redundant closures with direct method references:
- with_argv.rs: Replace |p| p.is_null() with <*const i8>::is_null
- which.rs: Replace |s| s.to_owned() with BStr::to_owned

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/crates/fspy_shared_unix/src/exec/which.rs
+++ b/crates/fspy_shared_unix/src/exec/which.rs
@@ -86,7 +86,7 @@ mod tests {
 
     #[test]
     fn test_concat() {
-        let s = concat(&["a".into(), "bc".into(), "".into(), "e".into()], |s| s.to_owned());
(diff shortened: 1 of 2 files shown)


crates/fspy_shared_unix/src/exec/which.rs
```diff
@@ -86,7 +86,7 @@ mod tests {
 
     #[test]
     fn test_concat() {
-        let s = concat(&["a".into(), "bc".into(), "".into(), "e".into()], |s| s.to_owned());
+        let s = concat(&["a".into(), "bc".into(), "".into(), "e".into()], BStr::to_owned);
```
fengmk2, line 89: Rust's performance optimization is really mechanical; even such detailed call overheads can be eliminated.

PR #233, opened 2025-10-16
style: fix clippy::redundant_clone warning

Brooooooklyn:
style: fix clippy::redundant_clone warning

Remove unnecessary `to_owned()` call in raw_exec.rs. The `name` variable
is already owned and being moved into the closure, so cloning it is
redundant.

Fixed 1 occurrence in:
- crates/fspy_preload_unix/src/client/raw_exec.rs

All tests pass and clippy is clean with -D clippy::redundant_clone.

Co-Authored-By: Claude <noreply@anthropic.com>

style: fix clippy::pedantic and clippy::nursery warnings

This commit addresses warnings from the pedantic and nursery lint groups:

**Auto-fixed (57 warnings):**
- Converted manual range slicing to `.take()` method calls
- Updated uninlined format arguments
- Simplified redundant field names in struct initialization
- Other automated fixes across 23 files

**Manual fixes:**

1. **build.rs**: Added underscores to long integer literals for readability
   - Fixed 4 occurrences of literals > 100 digits

2. **which.rs**: Replaced unsafe transmute with proper pointer operations
   - Used `std::ptr::copy_nonoverlapping` for safe byte copying
   - Added explicit type annotations for clarity

3. **exec/mod.rs**:
   - Replaced `Default::default()` with explicit `ParseShebangOptions::default()`
   - Converted nested if-let to `Option::map_or_else` for PATH resolution

4. **Documentation**: Added missing `# Errors` and `# Panics` sections
   - payload.rs: encode_payload function
   - spawn/mod.rs: handle_exec function
   - vite_path lib.rs: current_dir function
   - absolute.rs: strip_prefix method
   - relative.rs: strip_prefix method

5. **vite_path/relative.rs**: Added `#[allow(clippy::unsafe_derive_deserialize)]`
   - Safe to derive Deserialize as validation happens at construction time

6. **vite_tui/app.rs**:
   - Changed `App::new()` from `Result<Self>` to `Self` (no errors possible)
   - Replaced wildcard match arm with explicit `Action::Error(_)` variant

All tests pass successfully after these changes.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/crates/fspy_shared_unix/src/exec/shebang.rs
+++ b/crates/fspy_shared_unix/src/exec/shebang.rs
@@ -8,7 +8,7 @@ pub struct Shebang {
     pub arguments: Vec<BString>,
 }
 
-fn is_whitespace(c: u8) -> bool {
+const fn is_whitespace(c: u8) -> bool {
     c == b' ' || c == b'\t'
 }
 
@@ -54,7 +54,7 @@ pub fn parse_shebang(
     let arguments: Vec<BString> = if options.split_arguments {
         arguments_buf
             .split(|ch| is_whitespace(*ch))
-            .flat_map(|arg| {
+            .filter_map(|arg| {
                 let arg = arg.trim_ascii();
                 if arg.is_empty() { None } else { Some(arg.as_bstr().to_owned()) }
             })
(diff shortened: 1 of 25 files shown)


crates/fspy_shared_unix/src/exec/shebang.rs
```diff
@@ -54,7 +54,7 @@ pub fn parse_shebang(
     let arguments: Vec<BString> = if options.split_arguments {
         arguments_buf
             .split(|ch| is_whitespace(*ch))
-            .flat_map(|arg| {
```
fengmk2, line 57: This is not equivalent here

Brooooooklyn: It's equivalent here, so that the clippy warning is triggered here. Use `filter_map` for better semantics.

PR #238, opened 2025-10-21
feat(pm): add update command for package dependencies

fengmk2:
### TL;DR

Add `vite update` command to update packages across different package managers.

### What changed?

- Added a new `update.rs` module to handle package updates across pnpm, yarn, and npm
- Implemented `UpdateCommandOptions` to configure update behavior with various flags
- Added command resolution logic for each package manager with appropriate flags
- Created comprehensive test suite for all package managers and their versions
- Added snap tests for real-world usage scenarios
- Included documentation in RFC format

### How to test?

Test the new update command with different package managers:

```bash
# Basic update
vite update react

# Update to latest version
vite update react --latest

# Update all packages
vite update

# Update in specific workspace
vite update --filter app

# Update recursively in all workspaces
vite update -r

# Update only dev dependencies
vite update -D

# Interactive mode
vite update -i
```

### Why make this change?

This change provides a unified interface for updating packages across different package managers. It simplifies workflows by automatically detecting and using the correct package manager syntax, providing consistent commands regardless of the underlying tool. The implementation handles the differences between pnpm, yarn (v1 and v2+), and npm, making it easier for developers to maintain dependencies without remembering package manager-specific commands.

### DIFF
--- a/crates/vite_package_manager/src/lib.rs
+++ b/crates/vite_package_manager/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod add;
+pub mod commands;
 mod config;
 mod install;
 pub mod package;
--- a/crates/vite_package_manager/src/commands/mod.rs
+++ b/crates/vite_package_manager/src/commands/mod.rs
@@ -1,0 +1,1 @@
+pub mod update;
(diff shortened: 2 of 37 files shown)


crates/vite_package_manager/src/lib.rs
```diff
@@ -20,6 +21,7 @@ use wax::Glob;
 
 pub use crate::{
     add::{AddCommandOptions, SaveDependencyType},
+    commands::update::UpdateCommandOptions,
```
fengmk2: no need

PR #243, opened 2025-10-22
refactor(pm): debug

fengmk2:


crates/vite_task/src/lib.rs
```diff
@@ -409,6 +409,20 @@ pub async fn main<
         >,
     >,
 ) -> Result<std::process::ExitStatus, Error> {
+    let result = tokio::process::Command::new("pnpm")
```
fengmk2, line 412: Found a strange issue: running `./target/debug/vt add foo -w` directly succeeds, but running it through napi causes problems. It will freeze here.

```bash
VITE_LOG=trace ./packages/cli/bin/vite add foo -w                                        node(v22.21.0) 
2025-10-22T05:47:18.191108Z TRACE mio::poll: registering event source with poller: token=Token(1), interests=READABLE
```

@Brooooooklyn @branchseer

PR #244, opened 2025-10-22
build: fix reqwest build on Linux

Brooooooklyn:


Cargo.toml
```diff
@@ -98,7 +98,7 @@ rand = "0.9.1"
 ratatui = "0.29.0"
 rayon = "1.10.0"
 ref-cast = "1.0.24"
-reqwest = { version = "0.12", features = ["stream", "native-tls", "json"], default-features = false }
+reqwest = { version = "0.12", features = ["stream", "rustls-tls", "json"], default-features = false }
```
wan9chi: `rustls-tls` introduces `ring` which is hard to cross-compile for aarch64 Windows on macOS. Consider using `native-tls-vendored` which statically links to OpenSSL on Linux and dynamically links to the system's TLS library on Windows/macOS. It makes cross-compiling possible for both Windows and Linux on macOS.

Brooooooklyn: separate the features on different platform is not work: https://github.com/voidzero-dev/vite-plus/actions/runs/18710582811/job/53357856177

Brooooooklyn: @branchseer you can use `cargo-xwin clippy --target x86_64-pc-windows-msvc --all-targets --all-features` to run clippy on macOS for Windows codes

fengmk2: > @branchseer you can use `cargo-xwin clippy --target x86_64-pc-windows-msvc --all-targets --all-features` to run clippy on macOS for Windows codes

Should this experience be added to the contributor documentation?

PR #251, opened 2025-10-24
chore(cli): restore to run snap-test parallel

fengmk2:
I re-ran it 10 times locally and it hasn't failed again

### DIFF
--- a/packages/tools/src/snap-test.ts
+++ b/packages/tools/src/snap-test.ts
@@ -36,26 +36,19 @@ process.on('exit', () => fs.rmSync(tempTmpDir, { recursive: true, force: true })
 const casesDir = path.resolve('snap-tests');
 
 const filter = process.argv[2] ?? ''; // Optional filter to run specific test cases
-const isCI = process.env.CI === 'true';
 
 const tasks: Promise<void>[] = [];
 for (const caseName of fs.readdirSync(casesDir)) {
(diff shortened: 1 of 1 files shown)


packages/tools/src/snap-test.ts
```diff
@@ -37,17 +37,15 @@ const casesDir = path.resolve('snap-tests');
 
 const filter = process.argv[2] ?? ''; // Optional filter to run specific test cases
 
-// const tasks: Promise<void>[] = [];
+const tasks: Promise<void>[] = [];
 for (const caseName of fs.readdirSync(casesDir)) {
   if (caseName.startsWith('.')) continue; // Skip hidden files like .DS_Store
   if (caseName.includes(filter)) {
-    // FIXME: parallel run will cause [Error: Broken pipe (os error 32)] { code: 'GenericFailure' }
```
fengmk2: still panicked at CI env

Added via Giphy

fengmk2: @branchseer When you have time, can you take a look at this? The error is different from before https://github.com/voidzero-dev/vite-plus/actions/runs/18766740642/job/53543124754

fengmk2: But this stack was thrown from the napi side. How can I print more debugging information to investigate? @Brooooooklyn

PR #259, opened 2025-10-27
chore(LICENSE): BUSL-1.1

Boshen:
Reference: https://spdx.org/licenses/BUSL-1.1.html


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "vite-plus-monorepo",
+  "license": "BUSL-1.1",
```
fengmk2, line 3: The `license` field needs to be added to all packages that will be published to the npm registry.

PR #265, opened 2025-10-28
chore: support Windows and MacOS in binding version update

fengmk2:
Add `replace-file-content` utility to fix cross-platform compatibility in release workflow

## What changed?

- Created a new cross-platform Node.js utility script `replace-file-content` that replaces text in files
- Updated the release workflow to use this utility instead of `sed` for modifying the binding version
- Added tests for the new utility to ensure it works correctly
- Added the tools package as a workspace dependency in the root package.json

## Why?

The previous approach using `sed` in the release workflow is not cross-platform compatible, which can cause issues when running on different operating systems. The new utility provides a consistent way to replace file content across all platforms.

## How to test?

The PR includes snapshot tests for the new utility that verify it correctly replaces content in files. You can also manually test by running:

```
pnpm exec replace-file-content <filename> <search-text> <replacement-text>
```

For example:
```
pnpm exec replace-file-content example.toml 'version = "0.0.0"' 'version = "1.0.0"'
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,8 +48,7 @@ jobs:
       - uses: oxc-project/setup-node@fdbf0dfd334c4e6d56ceeb77d91c76339c2a0885 # v1.0.4
 
       - name: Set binding version
-        run: |
-          sed -i 's/version = "0.0.0"/version = "0.0.0-${{ github.sha }}"/' packages/cli/binding/Cargo.toml
+        run: pnpm exec replace-file-content packages/cli/binding/Cargo.toml 'version = "0.0.0"' 'version = "0.0.0-${{ github.sha }}"'
 
       - name: Build
         if: ${{ matrix.settings.target == 'x86_64-unknown-linux-gnu' }}
--- a/packages/tools/src/replace-file-content.ts
+++ b/packages/tools/src/replace-file-content.ts
@@ -1,0 +1,19 @@
+#!/usr/bin/env node
+
+import { readFileSync, writeFileSync } from 'node:fs';
+import path from 'node:path';
+
+const filename = process.argv[2];
+const searchValue = process.argv[3];
+const newValue = process.argv[4];
+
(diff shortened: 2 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -49,7 +49,12 @@ jobs:
 
       - name: Set binding version
         run: |
-          sed -i 's/version = "0.0.0"/version = "0.0.0-${{ github.sha }}"/' packages/cli/binding/Cargo.toml
+          if [ "$RUNNER_OS" = "Windows" ]; then
+            pwsh -Command "(Get-Content packages/cli/binding/Cargo.toml) -replace 'version = \"0.0.0\"', 'version = \"0.0.0-${{ github.sha }}\"' | Set-Content packages/cli/binding/Cargo.toml"
+          else
+            sed -i.bak 's/version = "0.0.0"/version = "0.0.0-${{ github.sha }}"/' packages/cli/binding/Cargo.toml && rm packages/cli/binding/Cargo.toml.bak
+          fi
+        shell: bash
```
wan9chi: Consider using a Node.js script to do this? It's cross-platform and easier to test locally.

packages/tools/src/replace-file-content.ts
```diff
@@ -0,0 +1,19 @@
...
+
+if (!filename || !searchValue || !newValue) {
+  console.error('Usage: replace-file-content <filename> <searchValue> <newValue>');
+  console.error('Example: replace-file-content example.toml \'version = "0.0.0"\' \'version = "0.0.1"\'');
+  process.exit(1);
+}
+
+const filepath = path.resolve(filename);
+const content = readFileSync(filepath, 'utf-8');
+const newContent = content.replace(searchValue, newValue);
```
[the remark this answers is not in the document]
fengmk2: Currently only the first one needs to be replaced

.github/workflows/release.yml
fengmk2: done!

PR #271, opened 2025-10-28
feat: add rolldown and rolldown-vite as sub packages

Brooooooklyn:


packages/cli/build.ts
```diff
@@ -58,29 +67,196 @@ if (output) {
...
+        transform(_, id, meta) {
+          if (id.endsWith('vite/src/node/constants.ts')) {
+            const { magicString } = meta;
+            if (magicString) {
+              magicString.replace(
+                `export const VITE_PACKAGE_DIR: string = resolve(
+  fileURLToPath(import.meta.url),
+  '../../..',
+)`,
+                `export const VITE_PACKAGE_DIR: string = path.join(fileURLToPath(/** #__KEEP__ */import.meta.url), '..', '..', '..', '..', '..')`,
```
fengmk2: A comment needs to be added to explain this, otherwise it will be difficult to remember the path `'..', '..', '..', '..', '..'`.

packages/cli/src/resolve-vite.ts
```diff
@@ -27,17 +29,8 @@ export async function vite(): Promise<{
...
-    // First try to resolve vite package.json
-    pkgJsonPath = resolve('vite/package.json');
-  } catch {
-    // Fallback to rolldown-vite package.json (for direct rolldown-vite installations)
-    pkgJsonPath = resolve('rolldown-vite/package.json');
-  }
-
   // Vite's CLI binary is located at bin/vite.js relative to the package root
-  const binPath = join(dirname(pkgJsonPath), 'bin', 'vite.js');
+  const binPath: string = join(dirname(fileURLToPath(import.meta.url)), 'vite', 'node', 'vite');
```
fengmk2: https://github.com/voidzero-dev/vite-task/blob/main/crates/vite_task/src/config/task_command.rs#L45

The whitelist check for `vite dev` in vite-task also needs to be adjusted accordingly.

.github/workflows/release.yml
```diff
@@ -45,38 +47,45 @@ jobs:
         if: ${{ matrix.settings.target == 'aarch64-unknown-linux-gnu' }}
         run: rustup target add aarch64-unknown-linux-musl
 
+      - name: Add rolldown host target
+        if: ${{ matrix.settings.target == 'aarch64-unknown-linux-gnu' }}
+        working-directory: rolldown
+        run: rustup target add x86_64-unknown-linux-gnu
```
fengmk2, line 53: why add x86_64-unknown-linux-gnu?

package.json
```diff
@@ -13,8 +13,8 @@
     "copy-bindings": "cp -r ./packages/cli/binding/*.node ./packages/cli/dist && cp -r ./packages/cli/binding/*.node ./packages/global/dist",
     "install-global-cli": "npm install -g ./packages/global",
     "typecheck": "tsc -b tsconfig.json",
-    "lint": "vite lint && vite run typecheck",
```
fengmk2, line 16: no more typecheck?

Brooooooklyn: The `typecheck` will check both `rolldown` and `rolldown-vite`, which can lead to numerous errors.
I'll add it back in e2e tests

packages/cli/package.json
```diff
@@ -14,42 +14,369 @@
...
+      "import": {
+        "types": "./dist/vitest/dist/index.d.ts",
+        "default": "./dist/vitest/dist/index.js"
+      },
+      "require": {
+        "types": "./dist/vitest/index.d.cts",
+        "default": "./dist/vitest/index.cjs"
+      }
+    },
+    "./browser": {
```
fengmk2, line 54: Should it be prefixed with `vitest` like `tsdown` and `vitepress`?
Otherwise, there is a high probability that it will conflict with vite's exports later on.

```json
"./vitest/browser": {
```

.github/workflows/release.yml
Brooooooklyn: The Vite+ CLI requires the Rolldown binary, and `x86_64-unknown-linux-gnu` is used for building Rolldown on the host system.

packages/cli/package.json
```diff
@@ -14,42 +14,369 @@
...
+    },
+    "unrun": {
+      "optional": true
+    },
+    "yaml": {
+      "optional": true
+    }
   },
   "devDependencies": {
-    "@napi-rs/cli": "catalog:",
```
fengmk2, line 46: `catalog:` Is this also not working anymore?

.gitignore
```diff
@@ -4,3 +4,7 @@ dist
 .claude/settings.local.json
 *.tsbuildinfo
 .DS_Store
+rolldown/pnpm-lock.yaml
+rolldown/pnpm-workspace.yaml
```
wan9chi: These are unnecessary?  `rolldown` is already listed.

.github/actions/clone/action.yml
wan9chi, line 1: Is there an easy way to checkout repos for local build?

packages/cli/package.json
Brooooooklyn: if we add prefix in the vitest exports, the vitest ecosystem packages would fail to resolve correct import path

.github/actions/clone/action.yml
Brooooooklyn: see readme

packages/cli/package.json
Brooooooklyn: The `catalog:` cannot be used when the `vite: file./packages/cli` override exists; I believe this is a limitation of pnpm.

packages/cli/package.json
fengmk2: OK, let's maintain the status quo for now.

.github/actions/clone/action.yml
Brooooooklyn: Added `just build` command and README doc

PR #275, opened 2025-10-29
refactor(tool): unify the cli entrypoint

Brooooooklyn:
I'm preparing to add more tools to `@voidzero-dev/vite-plus-tools`. Before adding more commands and tools, I want to unify the CLI entry point first.


packages/tools/src/bin.js
```diff
@@ -0,0 +1,4 @@
+import '@oxc-node/core/register';
```
fengmk2, line 1: I learned it, is this the best way to use it?

Brooooooklyn: No, it's not the best way, but it avoids calling `oxnode` before executing `cli.ts`.

PR #278, opened 2025-10-30
feat(cli): add sync remote subcommand

Brooooooklyn:


pnpm-workspace.yaml
```diff
@@ -7,10 +7,12 @@ catalog:
   '@napi-rs/cli': ^3.1.5
   '@oxc-node/cli': ^0.0.32
   '@oxc-node/core': ^0.0.32
+  '@std/yaml': npm:@jsr/std__yaml@^1.0.10
```
fengmk2, line 10: Is this JSR dependency support provided by pnpm?

packages/tools/src/index.ts
```diff
@@ -1,5 +1,6 @@
-import { replaceFileContent } from './replace-file-content';
-import { snapTest } from './snap-test';
+import { replaceFileContent } from './replace-file-content.js';
+import { snapTest } from './snap-test.js';
+import { syncRemote } from './sync-remote-deps.js';
```
wan9chi: Consider lazy importing the tools, if `sync-remote-deps` introduces heavy dependencies that are not used by the other tools.

pnpm-workspace.yaml
Brooooooklyn: yes, configured in `.npmrc`

PR #281, opened 2025-10-31
chore: add vite-task as cargo git dependency

wan9chi:


.github/workflows/ci.yml
```diff
@@ -52,6 +52,9 @@ jobs:
       - run: rustup target add x86_64-unknown-linux-musl
         if: ${{ matrix.os == 'ubuntu-latest' }}
 
+      - name: Configure Git for access to vite-task
+        run: git config --global url."https://x-access-token:${{ secrets.VITE_TASK_TOKEN }}@github.com/".insteadOf "https://github.com/"
```
fengmk2, line 56: @branchseer I just remembered, `.github/workflows/release.yml` also needs this logic added

wan9chi: Thanks for the reminder. I'll add it.

PR #283, opened 2025-11-01
chore: deps upgrade and clippy/fmt fix

Brooooooklyn:


crates/vite_task/src/ui.rs
```diff
@@ -38,11 +38,8 @@ pub fn get_display_command(display_options: DisplayOptions, task: &ResolvedTask)
     };
 
     let cwd = task.resolved_command.fingerprint.cwd.as_str();
-    Some(format!(
-        "{}$ {}",
-        if cwd.is_empty() { format_args!("") } else { format_args!("~/{cwd}") },
-        display_command
-    ))
```
wan9chi, line 45: Didn't expect bumping rust version would break this...🤔

PR #292, opened 2025-11-04
test: execute steps using @yarnpkg/shell

wan9chi:
### What changed?

- Removed `2>&1` redirections from test commands as we can capture stdout/stderr stably.
- Update snapshots with stdout/stderr interleaved.
- Refactored the snap test runner to use `@yarnpkg/shell`

### Why make this change?

To get prepared to run on Windows

### DIFF
--- a/packages/tools/src/snap-test.ts
+++ b/packages/tools/src/snap-test.ts
@@ -164,8 +164,10 @@ async function runTestCase(name: string, tempTmpDir: string, casesDir: string) {
       cwd,
       stdin: null,
       // Declared to be `Writable` but `FileHandle` works too.
-      stderr: outputStream as any,
-      stdout: outputStream as any,
+      // @ts-expect-error
+      stderr: outputStream,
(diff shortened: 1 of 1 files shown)


packages/tools/src/snap-test.ts
```diff
@@ -142,26 +149,41 @@ async function runTestCase(name: string, tempTmpDir: string, casesDir: string) {
...
+    // To ensure stable ordering, we redirect outputs to a file instead.
+    const outputStreamPath = path.join(caseTmpDir, 'output.log');
+    const outputStream = await open(outputStreamPath, 'w');
+
+    const exitCode = await execute(stripComments(command), [], {
+      env,
+      cwd,
+      stdin: null,
+      // Declared to be `Writable` but `FileHandle` works too.
+      stderr: outputStream as any,
```
fengmk2: use `@ts-expect-error` instead `any` will be better.

PR #300, opened 2025-11-12
fix(cli): Windows path issues

Brooooooklyn:


packages/cli/binding/src/commands/doc.rs
```diff
@@ -16,8 +16,12 @@ pub async fn doc<
 ) -> Result<ExecutionSummary, Error> {
     let ResolveCommandResult { bin_path, envs } =
         resolve_doc_command().await.map_err(|e| Error::Anyhow(e.into()))?;
-    let wrapped_command =
-        || async { Ok(ResolveCommandResult { bin_path: "node".into(), envs: envs.clone() }) };
+    let wrapped_command = || async {
+        Ok(ResolveCommandResult {
+            bin_path: (if cfg!(windows) { "node.exe" } else { "node" }).into(),
```
fengmk2: No need to change this here? which-rs already handles cross-platform compatibility
https://github.com/harryfei/which-rs

wan9chi: It could be `node.cmd` in `PATH` if the user installs node using volta or mise.

If `node.exe` works but `node` doesn't, we should find the root cause in vite-task.

PR #306, opened 2025-11-13
fix(cli): re-export * from vitest

Brooooooklyn:


package.json
```diff
@@ -12,7 +12,7 @@
     "bootstrap-cli:ci": "pnpm --filter=@voidzero-dev/global build && pnpm install-global-cli",
     "install-global-cli": "npm install -g ./packages/global",
     "typecheck": "tsc -b tsconfig.json",
-    "lint": "vite lint",
+    "lint": "vite lint --type-aware --threads 4",
```
fengmk2, line 15: Why is it necessary to specify the number of threads? Is there a problem that has occurred?

Brooooooklyn: https://github.com/oxc-project/oxc/issues/14375

vitest.config.ts
```diff
@@ -1,4 +1,4 @@
-import { defineConfig } from '@voidzero-dev/vite-plus';
+import { defineConfig } from '@voidzero-dev/vite-plus/config';
```
fengmk2, line 1: Is exporting `defineConfig` from `/config` the final solution?

Brooooooklyn: not sure... let's test in more repos

PR #310, opened 2025-11-18
feat: add package.json script migration with ast-grep

fengmk2:
### TL;DR

Added AST-based migration tool for package.json scripts to help users transition to vite-plus.

### What changed?

- Created a new `vite_migration` crate with functionality to rewrite package.json scripts using AST-grep rules
- Added AST-grep dependencies to support parsing and transforming bash scripts
- Implemented `rewritePackageJsonScripts` function exposed to JavaScript via NAPI
- Added YAML rules file for common script migrations (vite → vite dev, oxlint → vite lint, etc.)
- Updated the migration process to use AST-based transformation instead of simple string replacements
- Fixed template snapshots to reflect the new script transformations

### How to test?

1. Create a project with package.json containing scripts like `"dev": "vite"` or `"lint": "oxlint"`
2. Run the migration tool: `vp gen migration`
3. Verify that scripts are properly transformed (e.g., `"dev": "vite dev"`, `"lint": "vite lint"`)
4. Test with complex scripts containing pipes, conditionals, and multiple commands

### Why make this change?

The previous migration approach used simple string replacements which couldn't handle complex script commands. Using AST-grep provides more robust script transformations by properly parsing the bash syntax and applying targeted replacements. This ensures a smoother transition for users migrating from standalone tools to vite-plus, especially for projects with complex npm scripts.


Cargo.lock
```diff
@@ -128,6 +128,71 @@ dependencies = [
...
+ "tree-sitter-nix",
+ "tree-sitter-php",
+ "tree-sitter-python",
+ "tree-sitter-ruby",
+ "tree-sitter-rust",
+ "tree-sitter-scala",
+ "tree-sitter-solidity",
+ "tree-sitter-swift",
+ "tree-sitter-typescript",
+ "tree-sitter-yaml",
```
wan9chi, line 193: Lots of unnecessary deps introduced. We should enable only needed features: https://docs.rs/crate/ast-grep-language/latest/features

fengmk2: After trying, once the builtin-parser is activated, everything will be turned on, but there is no way to turn off the default builtin-parser.

https://docs.rs/crate/ast-grep-language/latest/source/src/parsers.rs

wan9chi: I changed `ast-grep-language = "0.40.0"`​ to `ast-grep-language = { version = "0.40.0", default-features = false, features = ["tree-sitter-json"] }`​, and saw no compilation errors. Would this work?

fengmk2: The build will not report an error, but it will prompt a missing feature at runtime.

```bash
thread 'package::tests::test_rewrite_script' (27411764) panicked at ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ast-grep-language-0.40.0/src/parsers.rs:47:3:
not implemented: tree-sitter parser is not implemented when feature flag [builtin-parser] is off.
```

wan9chi: That's a weird design. 🤔

PR #322, opened 2025-11-27
chore: update deps

Brooooooklyn:


packages/multiplexer/src/control.ts
```diff
@@ -44,10 +44,7 @@ export class ControlPanel {
 
   getDimensions(): Dimensions {
```
fengmk2: Has packages/multiplexer been deprecated? Can it be deleted?

packages/cli/binding/index.d.ts
```diff
@@ -25,10 +25,7 @@ export interface CliOptions {
   /** Optional working directory override */
   cwd?: string;
   /** Read the vite.config.ts in the Node.js side and return the `lint` and `fmt` config JSON string back to the Rust side */
-  resolveUniversalViteConfig: (
```
fengmk2, line 28: this auto-generated file should be ignore?

Brooooooklyn: Let's format it. I've called `oxfmt` in `build.ts` to ensure it will be formatted after the build.

PR #329, opened 2025-12-03
test: ignore unstable snap-test output

fengmk2:
### TL;DR

Added support for ignoring "npm notice Access token expired or revoked" warnings in test output.

### What changed?

- Added a new regex pattern in `replaceUnstableOutput()` function to filter out "npm notice Access token expired or revoked" warning messages
- Added a corresponding test case to verify the new functionality
- Updated test snapshots to include the new test case

### How to test?

1. Run the test suite to verify that the new test case passes
2. Verify that when running commands that might produce the "npm notice Access token expired or revoked" warning, these messages are properly filtered out from the output

### Why make this change?

These npm access token warnings are non-deterministic and can cause test failures when they appear in output that's being compared against snapshots. By filtering them out, we improve test stability and prevent false negatives in test runs.

### DIFF
--- a/packages/tools/src/__tests__/__snapshots__/utils.spec.ts.snap
+++ b/packages/tools/src/__tests__/__snapshots__/utils.spec.ts.snap
@@ -20,6 +20,12 @@ added 3 packages in <variable>ms
 Done in <variable>ms"
 `;
 
+exports[`replaceUnstableOutput() > replace ignore npm notice access token expired or revoked warning log 1`] = `
+"line 1
+line 2
+line 3"
+`;
+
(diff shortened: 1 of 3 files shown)


packages/tools/src/__tests__/__snapshots__/utils.spec.ts.snap
```diff
@@ -20,6 +20,8 @@ added 3 packages in <variable>ms
 Done in <variable>ms"
 `;
 
+exports[`replaceUnstableOutput() > replace ignore npm notice access token expired or revoked warning log 1`] = `"npm notice Access token expired or revoked. Please try logging in again."`;
```
fengmk2: not working here

PR #341, opened 2025-12-11
refactor(cli): remove package manager commands from local cli

fengmk2:
### TL;DR

Simplified CLI command structure by removing package manager related commands.

### What changed?

- Removed package manager related commands (`add`, `remove`, `update`, `dedupe`, `outdated`, `why`, `link`, `unlink`, and `pm`)
- Added helpful command descriptions to the remaining commands
- Simplified the `is_package_manager_command()` function to only check for the `install` command
- Updated the help text to reflect the new command structure

### How to test?

1. Run `vite -h` to verify the updated command list
2. Verify that only the core commands are available: `run`, `lint`, `fmt`, `build`, `test`, `lib`, `dev`, `doc`, `cache`, and `install`
3. Check that each command has a descriptive help text

### Why make this change?

This change streamlines the CLI interface by focusing on core functionality and removing rarely used package manager commands. This makes the CLI more approachable for new users and reduces maintenance overhead. The simplified command structure aligns better with the primary purpose of the tool while still maintaining the essential `install` command for backward compatibility.


.github/workflows/release.yml
```diff
@@ -2,6 +2,10 @@ name: Release
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      # release test, will be removed after testing
```
fengmk2: https://github.com/voidzero-dev/vite-plus/actions/runs/20161739372

PR #342, opened 2025-12-12
fix: update build-types filter from rolldown-vite to vite

fengmk2:
### TL;DR

Updated build scripts to use the correct package filter for building types and added a postbuild script to verify output files.

### What changed?

- Changed the package filter in GitHub action from `rolldown-vite` to `vite` for the `build-types` command in three different build configurations
- Added a `postbuild` script to the core package that runs after build completion to list files in the `dist/vite/node` directory

### How to test?

1. Run the build process using the GitHub action to verify that types are correctly built using the `vite` filter
2. Check the build logs to confirm that the `postbuild` script executes and displays the contents of the `dist/vite/node` directory

### Why make this change?

The package filter was incorrectly referencing `rolldown-vite` instead of `vite`, which could cause type generation to fail or target the wrong package. The added postbuild script helps with debugging by providing visibility into the build output files, making it easier to verify that the correct files are being generated.

### DIFF
--- a/.github/actions/build-upstream/action.yml
+++ b/.github/actions/build-upstream/action.yml
@@ -20,7 +20,7 @@ runs:
           pnpm --filter rolldown build-binding:release --target ${{ inputs.target }} --use-napi-cross
         fi
         pnpm --filter rolldown build-node
-        pnpm --filter rolldown-vite build-types
+        pnpm --filter vite build-types
         pnpm --filter=@voidzero-dev/vite-plus-core build
         pnpm --filter=@voidzero-dev/vite-plus-test build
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -72,7 +72,8 @@
     "node": "^20.19.0 || >=22.12.0"
   },
   "scripts": {
-    "build": "oxnode -C dev ./build.ts"
+    "build": "oxnode -C dev ./build.ts",
+    "postbuild": "ls -la dist/vite/node"
```
fengmk2: come back

fengmk2: release test https://github.com/voidzero-dev/vite-plus/actions/runs/20160047627/job/57870256424

PR #348, opened 2025-12-16
docs: add vitest deps bundle doc

Brooooooklyn:


packages/test/BUNDLING.md
```diff
@@ -0,0 +1,365 @@
...
+```
+
+### Config File Updates
+
+```typescript
+// Before
+import { playwright } from '@vitest/browser-playwright';
+
+// After
+import { playwright } from 'vitest/browser-playwright';
```
fengmk2: can import from `@voidzero-dev/vite-plus-test`?

PR #355, opened 2025-12-18
docs: deploy on cloudflare pages

fengmk2:


docs/package.json
```diff
@@ -4,13 +4,12 @@
   "private": true,
   "scripts": {
     "dev": "vitepress dev",
-    "build": "vite doc build",
-    "preview": "vite doc preview"
+    "build": "vitepress build",
+    "preview": "vitepress preview"
   },
   "devDependencies": {
-    "@voidzero-dev/vite-plus": "workspace:*",
```
fengmk2, line 11: hard to build vite-plus on cf, fallback to use vitepress on docs only

PR #359, opened 2025-12-22
test(e2e): add vite migration e2e

fengmk2:
> [!NOTE]
> - **E2E/CI:** Adds `rollipop` to the matrix with custom commands and `skip-install`, sets `GITHUB_TOKEN` in job env, and avoids incorrect `.gitignore` usage during migration.
> - **Ecosystem helper:** `patch-project.ts` gains a `vite migrate` path using tgz via `VITE_PLUS_OVERRIDE_PACKAGES` and `VITE_PLUS_VERSION`; `repo.json` and `.gitignore` updated to include `rollipop`.
> - **Migration tooling:** Replaces hardcoded override map with env-driven `VITE_PLUS_OVERRIDE_PACKAGES` and `VITE_PLUS_VERSION`, updating package.json and workspace rewrite logic across pnpm/yarn/npm.
> - **Core package:** Fixes `exports["./lib"].default` to `./dist/tsdown/index.js` and adds `@babel/*` dependencies.
> - **Lockfile:** Updated for new dependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b6debb303daa1a3ebda43c3346e6462205d15dce. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #360, opened 2025-12-23
fix(cli): fix tsgolint path resolution for yarn install

fengmk2:
### What changed?

- Added `existsSync` import from `node:fs` to check if files exist
- Modified the tsgolint binary resolution logic to:
  - First try to resolve from `oxlint-tsgolint/bin/tsgolint` for non-Windows platforms
  - For Windows, attempt to find the binary in the local `node_modules/.bin` directory
  - Add a fallback to the current working directory's `node_modules/.bin` if not found
- Simplified the environment variable assignment by moving the platform-specific logic to the path resolution

```bash
yarn lint

Failed to find tsgolint executable: OXLINT_TSGOLINT_PATH points to 
'~/rollipop/node_modules/@voidzero-dev/vite-plus/node_modules/.bin/tsgolint' 
which does not exist
```

### DIFF
--- a/packages/cli/src/resolve-lint.ts
+++ b/packages/cli/src/resolve-lint.ts
@@ -11,6 +11,7 @@
  * provides ESLint-compatible linting with significantly better performance.
  */
 
+import { existsSync } from 'node:fs';
 import { dirname, join } from 'node:path';
 import { relative } from 'node:path/win32';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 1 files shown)


packages/cli/src/resolve-lint.ts
```diff
@@ -33,22 +34,28 @@ export async function lint(): Promise<{
...
+    oxlintTsgolintPath = join(
+      dirname(fileURLToPath(import.meta.url)),
+      '..',
+      'node_modules',
+      '.bin',
+      `tsgolint.cmd`,
+    );
+    if (!existsSync(oxlintTsgolintPath)) {
+      // Fallback to the cwd node_modules
+      oxlintTsgolintPath = join(process.cwd(), 'node_modules', '.bin', `tsgolint.cmd`);
```
fengmk2: e2e test will add on another pull request https://github.com/voidzero-dev/vite-plus/pull/359

PR #367, opened 2025-12-25
feat(migration): skip rewriting imports for packages in peerDependencies

fengmk2:
When a project has vite in peerDependencies (like a Vite plugin), the
import rewriter now skips rewriting vite imports to avoid accidentally
bundling vite-plus into the library.

- Split AST-grep rules into separate VITE, VITEST, and TSDOWN rule sets
- Added get_skip_packages_from_root() to check root package.json for peerDependencies
- Apply rules conditionally based on which packages are in peerDependencies
- Added comprehensive unit tests for peerDependencies scenarios
- Added snap test: migration-skip-vite-peer-dependency
- Updated CLAUDE.md with build and snap-test instructions

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/crates/vite_migration/src/import_rewriter.rs
+++ b/crates/vite_migration/src/import_rewriter.rs
@@ -1,4 +1,7 @@
-use std::path::{Path, PathBuf};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
 
 use vite_error::Error;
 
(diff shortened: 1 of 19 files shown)


crates/vite_migration/src/import_rewriter.rs
```diff
@@ -265,6 +274,72 @@ transform:
 fix: $NEW_IMPORT
 "#;
 
+/// Packages to skip rewriting based on peerDependencies
```
fengmk2: should includes dependencies too.

PR #431, opened 2026-01-15
feat: Add an `--agent` flag and interactive prompt.

cpojer:
This PR adds an `--agent` flag and an interactive "agent" prompt to the `new` and `migrate` commands. It allows selecting the coding agent that the user users, and will write the `AGENT.md` file from the package into the user's project using the name preferred by the selected agent.

For migrate, it will propose appending to the file if it already exists.

The PR also reorders the `vite new` prompts.
**Before: template, package manager, name, git repo**
**After: template, name, package manager, agent, git repo** 

I will add actual content into the AGENTS.md in a follow-up.

### DIFF
--- a/packages/global/AGENTS.md
+++ b/packages/global/AGENTS.md
@@ -1,0 +1,3 @@
+# Using Vite+, the Unified Toolchain for the Web
+
+*TODO*
--- a/packages/global/src/utils/agent.ts
+++ b/packages/global/src/utils/agent.ts
@@ -1,0 +1,168 @@
+import fs from 'node:fs';
+import fsPromises from 'node:fs/promises';
+import path from 'node:path';
+
+import * as prompts from '@clack/prompts';
+
+import { pkgRoot } from './path.js';
+
+export const AGENTS = [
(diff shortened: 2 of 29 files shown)


packages/global/AGENTS.md
```diff
@@ -0,0 +1,3 @@
+# Using Vite+, the Unified Toolchain for the Web
+
+*TODO*
```
fengmk2, line 3: Will the content here be updated before release?

cpojer: Yes. I will hand-write it in the next PR so we can review the content in detail.

packages/global/src/utils/agent.ts
```diff
@@ -0,0 +1,168 @@
+import fs from 'node:fs';
+import fsPromises from 'node:fs/promises';
+import path from 'node:path';
+
+import * as prompts from '@clack/prompts';
+
+import { pkgRoot } from './path.js';
+
+export const AGENTS = [
+  { id: 'chatgpt-codex', label: 'ChatGPT (Codex)', targetPath: 'AGENTS.md' },
```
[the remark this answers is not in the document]
cpojer: Haha are you bullying codex or what

PR #434, opened 2026-01-15
Add AGENTS.md.

cpojer:
This PR specs out the `AGENTS.md` file that is copied into projects via `vite new` or `vite migrate`. It's an initial version meant to steer LLMs and reduce confusion. It focused on:

- Vite+ is not Vite
- Do not install the individual tools like Oxlint
- Do not try to import from or invoke individual tools like Vitest
- Ask the LLM to always use `vite`, and captures a few gotchas like `vite test` not invoking the package.json "scripts.test".
- Brevity

This version is probably not perfect. It intentionally does not list every single command (agents can run `vite help`). My suggestion is to ship it, use it, and then iterate it when we find issues.

I fixed a few typos in the CLI's help output.

_Handwritten, no AI._

### DIFF
--- a/packages/global/AGENTS.md
+++ b/packages/global/AGENTS.md
@@ -1,3 +1,52 @@
+<!--VITE PLUS START-->
 # Using Vite+, the Unified Toolchain for the Web
 
-*TODO*
+This project is using Vite+, a modern toolchain built on top of Vite, Rolldown, Vitest, tsdown, Oxlint, and Oxfmt. Vite+ wraps these tools and package manager commands in a single, global CLI called `vite`. Vite+ is distinct from Vite, but it invokes Vite through `vite dev` and `vite build`.
+
+## Vite+ Workflow
+
+`vite` is a global binary that handles the full development lifecycle. Run `vite help` to print a list of commands and `vite <command> --help` for information about a specific command.
+
+### Vite+ Commands
+
+- dev - Run the development server
+- build - Build for production
+- lint - Lint code
+- test - Run tests
+- fmt - Format code
+- lib - Build library
+- new - Create a new monorepo package (in-project) or a new project (global)
+- run - Run tasks from `package.json` scripts
+
+These commands map to their corresponding tools. For example, `vite dev --port 3000` runs Vite's dev server and works the same as Vite. `vite test` runs JavaScript tests through the bundled Vitest. The versions of individual tools can be checked using `vite <command> --version`. For example, `vite lint --version` prints the bundled Oxlint version, and `vite test --version` prints the bundled Vitest version. This is useful when researching documentation, features, and bugs.
+
+### Package Manager Commands
+
+Vite+ automatically detects and wraps the underlying package manager such as pnpm, npm, or Yarn through the `packageManager` field in `package.json` or package manager-specific lockfiles.
+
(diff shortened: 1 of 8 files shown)


packages/global/AGENTS.md
```diff
@@ -1,3 +1,47 @@
...
+`vite` is a global binary that handles the full development lifecycle. Run `vite help` to print a list of commands and `vite <command> --help` for information about a specific command.
+
+### Vite+ Commands
+
+- dev - Run the development server
+- build - Build for production
+- lint - Lint code
+- test - Run tests
+- fmt - Format code
+- run - Run tasks from `package.json` scripts
```
yyx990803, line 19: Also `vite new`? Feels important enough to list

packages/global/AGENTS.md
```diff
@@ -1,3 +1,47 @@
...
+
+These commands map to their corresponding tools. For example, `vite dev` runs Vite's dev server. `vite test` runs JavaScript tests through the bundled Vitest. The versions of individual tools can be checked using `vite <command> --version`. For example, `vite lint --version` prints the bundled Oxlint version, and `vite test --version` prints the bundled Vitest version. This is useful when researching documentation, features, and bugs.
+
+### Package Manager Commands
+
+Vite+ forwards common commands to the underlying package manager such as pnpm, npm, or Yarn. Check the `package.json` or lockfiles to know which package manager is used in this project.
```
yyx990803: We should be more explicit that (1) Vite+ automatically determines which package manager to use. and (2) the criteria it uses: specifically the `packageManager` field in `package.json` or the package-manager-specific lockfiles such as `pnpm-lock.yaml`

packages/global/AGENTS.md
yyx990803: Should we add a few more examples here so that the agent understands how these commands map better to respective tools (especially less trained ones like `oxlint` and `oxfmt`)?

e.g.

- `vite dev --port 3000` works exactly the same as normal Vite
- `vite lint --type-aware` works out of the box without having to install `oxlint-tsgolint`

Maybe also link to respective CLI docs of underlying tools for each command?

cpojer: Added `--port 3000` in the example test, and the `--type-aware` in common pitfalls telling the agent they don't need to install `oxlint-tsgolint`.

cpojer: I originally omitted out `new` since this is only in the context of a project, not outside of it, but since we can use it to add new packages in a monorepo I'll add it back.

packages/global/AGENTS.md
```diff
@@ -1,3 +1,51 @@
...
+This project is using Vite+, a modern toolchain built on top of Vite, Rolldown, Vitest, tsdown, Oxlint, and Oxfmt. Vite+ wraps these tools and package manager commands in a single, global CLI called `vite`. Vite+ is distinct from Vite, but it invokes Vite through `vite dev` and `vite build`.
+
+## Vite+ Workflow
+
+`vite` is a global binary that handles the full development lifecycle. Run `vite help` to print a list of commands and `vite <command> --help` for information about a specific command.
+
+### Vite+ Commands
+
+- dev - Run the development server
+- build - Build for production
```
fengmk2, line 13: Add `lib` command introduction?

PR #436, opened 2026-01-15
fix: should not download binding to dist

fengmk2:


packages/global/package.json
```diff
@@ -53,10 +53,8 @@
     "packageName": "@voidzero-dev/vite-plus-cli",
     "targets": [
       "aarch64-apple-darwin",
-      "x86_64-apple-darwin",
       "aarch64-unknown-linux-gnu",
       "x86_64-unknown-linux-gnu",
-      "aarch64-pc-windows-msvc",
```
fengmk2, line 59: Align vite-plus configuration

PR #439, opened 2026-01-16
feat: Add the version of all local tools to `vite --version`.

cpojer:
This PR adds the bundled tooling versions to the `vite --version` command. This will be helpful for debugging, for LLMs to know what they are actually using, and for issue reporting.

The implementation works by using the build step to write the bundled tool versions to `package.json`, which are then printed through `vite --version`.

Let me know if you'd like the design or text to be different.


packages/global/src/version.ts
```diff
@@ -1,39 +1,186 @@
...
+      packageName: 'oxlint-tsgolint',
+    },
+    {
+      command: 'lib',
+      displayName: 'tsdown',
+      packageName: '@voidzero-dev/vite-plus-core',
+      bundledVersionKey: 'tsdown',
+    },
+    {
+      command: 'doc',
```
fengmk2, line 146: Do not display the dependency version of the `doc` for now, as we also did not display the `doc` command in the helper message.

packages/core/package.json
```diff
@@ -194,5 +194,9 @@
   },
   "optionalDependencies": {
     "fsevents": "~2.3.3"
+  },
+  "bundledVersions": {
```
fengmk2, line 198: Is there any way to ensure that we don't forget to update the `bundledVersions` field when we update dependency versions later?  
  
Would it be better to automatically collect the correct dependency information during the `build` stage and generate a bundledVersions.json file? No manual maintenance required.

cpojer: The build commands are writing/updating those for us already!

PR #458, opened 2026-01-19
fix: Fix TypeScript types + run tsgo in CI.

cpojer:
This PR fixes the TypeScript setup:

* Removes `dev` as custom condition which was pulling in the wrong files from Rolldown.
* Fixes a few TypeScript errors I introduced
* Runs tsgo on CI (Ubuntu only, we don't need to run it on every platform)


package.json
```diff
@@ -29,6 +29,7 @@
     "@types/node": "catalog:",
     "@typescript/native-preview": "catalog:",
     "@voidzero-dev/vite-plus-tools": "workspace:*",
+    "bingo": "catalog:",
```
fengmk2, line 32: why add bingo and zod to devDeps?

cpojer: Just dev deps, because they are used in the `templates/generator` folder but aren't installed directly, so TypeScript complains.

packages/cli/tsconfig.json
```diff
@@ -1,26 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "module": "ESNext",
```
fengmk2, line 4: @cpojer should keep `module: ESNext` and `moduleResolution: bundler` configs? After the commit merge, the e2e ci test fails https://github.com/voidzero-dev/vite-plus/actions/runs/21161971415

cpojer: Oh, is this relevant for the build of the CLI itself? If so, I can add it back to the `build.ts` directly (and we can switch the cli build to `vite lib` next).

fengmk2: I was just looking into what change caused this last night and only suspected a correlation here — let me verify whether it’s related to this configuration

PR #479, opened 2026-01-21
fix(cli): restore tsconfig-based TypeScript compilation

fengmk2:
Partially revert de120bee8 to restore the original TypeScript build
approach for packages/cli:
    
- Use tsconfig.json to define compilation files via include/exclude
  instead of globSync in build.ts
- Let tsconfig.json control outDir and noEmit settings
- Simplify build.ts by using fileNames from parsed tsconfig
    
This keeps the tsgo CI integration and other TypeScript fixes from
the original commit while reverting only the CLI build changes.

### DIFF
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -1,12 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "./dist",
-    "noEmit": false,
-    "allowImportingTsExtensions": false,
-    "rootDir": "./src",
     "experimentalDecorators": true
   },
-  "include": ["src"],
-  "exclude": ["node_modules", "dist/**/*"]
+  "files": [],
+  "include": [],
+  "exclude": ["**/*"]
 }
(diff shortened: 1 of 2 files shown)


packages/cli/tsconfig.json
```diff
@@ -1,9 +1,12 @@
...
+    "noEmit": false,
+    "allowImportingTsExtensions": false,
+    "rootDir": "./src",
     "experimentalDecorators": true
   },
-  "files": [],
-  "include": [],
-  "exclude": ["**/*"]
+  "include": ["src"],
+  "exclude": ["node_modules", "dist/**/*"]
```
cpojer: When we do this, it kinda confused TypeScript about how many projects we have. Would it be possible that we keep it as is, and move all these options into `build.ts` directly?

  
That also makes it easier to clean up once we move the `build.ts` code to use tsdown to build the cli. Alternatively, we could also migrate to tsdown now.

fengmk2: I try again

PR #488, opened 2026-01-23
docs: initiate shared theme

fengmk2:
merge from https://github.com/voidzero-dev/vite-plus/pull/464


.ignore
```diff
@@ -1,4 +0,0 @@
-# For watchexec https://github.com/watchexec/watchexec/tree/main/crates/cli#features
-
-target/**
-**/node_modules/**
```
fengmk2, line 4: I found the cause of Vitepress not rendering the new navigation bar was due to this file after removing files one by one.

vite.config.ts
```diff
@@ -21,7 +21,7 @@ export default <UserConfig>{
         },
       },
     ],
-    ignorePatterns: ['**/snap-tests/**', '**/snap-tests-todo/**'],
+    ignorePatterns: ['**/snap-tests/**', '**/snap-tests-todo/**', '**/docs/**'],
```
fengmk2: should fix lint error on docs

docs/.vitepress/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
```
cpojer, line 1: If we are keeping this config separate, it would be best to include it as a reference in the root.

docs/.vitepress/theme/data/testimonials.ts
```diff
@@ -0,0 +1,52 @@
...
+    quote:
+      "I verified Rolldown's bundling behavior in my OSS projects, confirming proper operation and faster bundling than Rollup",
+    logo: '/logos/bolt.svg',
+    logoAlt: 'Bolt',
+    name: 'Kazupon',
+    title: 'Engineer',
+    company: 'Plaid',
+    image: '@local-assets/people/kazupon.jpg',
+  },
+];
```
cpojer: Can you zero out this array for now?

fengmk2: OK

docs/.vitepress/tsconfig.json
fengmk2: 😭 The tsgo error is from the external vitepress-theme package trying to re-export a type from a .vue file, which our generic .vue declaration  doesn't support. This would need to be fixed in the theme package itself.  
  
I will add a reference after vitepress-theme package typeerror fixed in a future pull request.

fengmk2: https://linear.app/voidzero/issue/VP-162/add-docs-tsconfigjson-as-a-reference-from-root-tsconfig

PR #490, opened 2026-01-23
fix(ci): include env vars in NAPI cache key for release builds

fengmk2:
Add RELEASE_BUILD, DEBUG, and VERSION environment variables to the
cache key. Since VERSION contains the commit SHA, this ensures cache
miss on each release build without needing a skip-cache parameter.

### DIFF
--- a/.github/actions/build-upstream/action.yml
+++ b/.github/actions/build-upstream/action.yml
@@ -16,11 +16,12 @@ runs:
 
     # Compute cache key once before any builds modify files
     # (packages/cli/package.json is modified by syncTestPackageExports during build-ts)
+    # Include env vars (RELEASE_BUILD, DEBUG, VERSION) to ensure cache miss on release builds
     - name: Compute NAPI binding cache key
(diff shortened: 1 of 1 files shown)


.github/actions/build-upstream/action.yml
```diff
@@ -25,6 +29,7 @@ runs:
     # Cache only NAPI bindings (the slow part, especially on Windows)
```
fengmk2, line 26: <p class='graphite__hidden'><i>[Re: line +26]</i></p>
should add  RELEASE_BUILD env to cache key
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/490?utm_source=unchanged-line-comment">Graphite</a>.</p>

fengmk2: https://github.com/voidzero-dev/vite-plus/blob/main/packages/cli/build.ts#L79

PR #506, opened 2026-01-26
fix(lib): reload config on watch restart

fengmk2:
Pass the build function as the restart callback to `buildWithConfigs`
instead of an empty function, ensuring config is re-resolved when
watch mode detects config file changes.

Ref: https://github.com/rolldown/tsdown/commit/1756b03

close VP-160


packages/cli/snap-tests-todo/command-lib-watch-restart/run-watch.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# Start vite lib watch in background, redirect output to file, and save PID
+vite lib --watch > .watch-output.log 2>&1 &
```
fengmk2, line 3: rolldown watch mode is not working on Github Action CI env, but I run this snap-test on local it was worked. Skip it for now.

PR #564, opened 2026-02-09
fix(cli): use project Node.js version for pnpm/yarn shim dispatch

fengmk2:
Package managers like pnpm/yarn were using the install-time Node.js
version instead of resolving from the project context (engines.node,
.node-version, etc.), causing "Unsupported engine" warnings when the
project requires a different Node version.


packages/global/snap-tests/shim-pnpm-uses-project-node-version/snap.txt
```diff
@@ -0,0 +1,5 @@
+> node -v # Node version resolved from engines.node
+v22.22.0
+
+> pnpm exec node -v # pnpm should use same project Node version
+v22.22.0
```
fengmk2: There is a problem, incorrect.

PR #587, opened 2026-02-10
chore: Fork `@clack/prompts`.

cpojer:
This PR forks `@clack/prompts` and replaces `cyan` with `blueBright` for interactive commands. It gives Vite+ a bit more identity, and we can customize things as we like.

This is lightweight because it still uses `@clack/core` under the hood. Unfortunately `@clack/prompts` doesn't allow any customization at all.

Fixed lint errors in the prompts folder but left everything else untouched.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "license": "BUSL-1.1",
   "type": "module",
   "scripts": {
-    "build": "pnpm -F @voidzero-dev/* -F vite-plus-prompts build && pnpm -F @voidzero-dev/* -F vite-plus build && pnpm -F vite-plus-cli build",
+    "build": "pnpm -F @voidzero-dev/* -F vite-plus build && pnpm -F vite-plus-cli build",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "license": "BUSL-1.1",
   "type": "module",
   "scripts": {
-    "build": "pnpm -F @voidzero-dev/* -F vite-plus build && pnpm -F vite-plus-cli build",
+    "build": "pnpm -F @voidzero-dev/* -F vite-plus-prompts build && pnpm -F @voidzero-dev/* -F vite-plus build && pnpm -F vite-plus-cli build",
```
fengmk2: ` -F @voidzero-dev/*` should includes "@voidzero-dev/vite-plus-prompts" too?

cpojer: You are right, my mistake. https://github.com/voidzero-dev/vite-plus/pull/587/changes/7ec2533b9230b9346751c7647f8efe28e1fa29d4 should fix the CI issue 🙏

PR #620, opened 2026-02-24
feat: add `vp exec` command for local node_modules/.bin execution

fengmk2:
Add `vp exec` as the equivalent of `pnpm exec` — prepends
`./node_modules/.bin` to PATH and executes a command. Commands resolve
through the modified PATH (local bins first, then system PATH), matching
pnpm exec's real behavior. No remote fallback unlike `vpx`.

Global CLI: delegates to local CLI when vite-plus is a dependency,
otherwise handles exec directly (prepend PATH + spawn).

Local CLI: prepends PM bin dir and node_modules/.bin to PATH, sets
VITE_PLUS_PACKAGE_NAME env var, supports --shell-mode/-c flag.

closes [VP-194](https://linear.app/voidzero/issue/VP-194/equivalent-of-pnpm-filter-pkgname-exec-some-cmd)

### DIFF
--- a/packages/cli/AGENTS.md
+++ b/packages/cli/AGENTS.md
@@ -18,6 +18,7 @@ This project is using Vite+, a modern toolchain built on top of Vite, Rolldown, 
 - lib - Build library
 - migrate - Migrate an existing project to Vite+
 - new - Create a new monorepo package (in-project) or a new project (global)
+- exec - Execute a command in workspace packages (supports `--filter`, `-r`, `--parallel`)
 - run - Run tasks from `package.json` scripts
 
 These commands map to their corresponding tools. For example, `vp dev --port 3000` runs Vite's dev server and works the same as Vite. `vp test` runs JavaScript tests through the bundled Vitest. The version of all tools can be checked using `vp --version`. This is useful when researching documentation, features, and bugs.


packages/cli/binding/src/exec/filter.rs
```diff
@@ -0,0 +1,718 @@
+use std::collections::{BTreeMap, HashSet, VecDeque};
+
+use petgraph::Direction;
+use rustc_hash::{FxHashMap, FxHashSet};
+use vite_path::AbsolutePath;
+
+/// Parsed package selector for `--filter` flag (pnpm-compatible syntax).
```
fengmk2, line 7: @branchseer I implement a Rust version using pnpm filter, and once your implementation is ready, I will switch over. I have added snap tests to ensure that the functionality remains consistent during the subsequent transition.

packages/cli/snap-tests/command-exec/snap.txt
```diff
@@ -0,0 +1,61 @@
...
+  vp exec eslint .
+  vp exec tsc --noEmit
+
+[1]> vp exec nonexistent-cmd-12345 # command not found error
+[1m[31merror:[39m[0m Command 'nonexistent-cmd-12345' not found in node_modules/.bin
+
+Hint: Run 'vp install' to install dependencies, or use 'vpx' for remote fallback.
+
+> vp run foo # vp exec works in package.json scripts
+$ vp exec node -e "console.log(5173)" ⊘ cache disabled
```
fengmk2, line 60: missing 5173 output

PR #621, opened 2026-02-24
feat: add clippy disallowed-macros rules for println!/eprintln!

fengmk2:
Enforce that all user-facing output goes through `vite_shared::output`
functions instead of raw print macros. Also document CLI output
conventions for both Rust and TypeScript in CLAUDE.md.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -80,7 +80,7 @@ vp run dev                         # runs dev script from package.json
 
 ## Clippy Rules
 
-All Rust code must follow the custom clippy rules defined in `.clippy.toml` (disallowed types, macros, and methods).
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -78,6 +78,17 @@ vp run dev                         # runs dev script from package.json
 
 - **Passing to std functions**: `AbsolutePath` implements `AsRef<Path>`, use `.as_path()` when explicit `&Path` is required
 
+## Clippy Rules
+
+All **new** Rust code must follow the custom clippy rules defined in `.clippy.toml` (disallowed types, macros, and methods). Existing code may not fully comply due to historical reasons.
```
fengmk2, line 83: https://github.com/voidzero-dev/vite-plus/issues/622 All clippy issues need to be fixed before open sourcing.

PR #643, opened 2026-02-26
feat(cli): add `vp check` command for format, lint, and type checks

fengmk2:
Add `vp check` as a built-in composite command that runs `vp fmt --check`
then `vp lint --type-aware --type-check` sequentially with fail-fast
semantics. Supports `--no-fmt`, `--no-lint`, `--no-type-aware`, and
`--no-type-check` flags to selectively disable checks.

This is the first composite command in vite-plus — all existing built-in
commands delegate to a single underlying tool.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "zod": "catalog:"
   },
   "lint-staged": {
-    "*.@(js|ts|tsx)": "vp check --fix",
+    "*.@(js|ts|tsx|md|yaml|yml)": "vp check --fix",
     "*.rs": "cargo fmt --"
   },
   "engines": {


rfcs/check-command.md
```diff
@@ -0,0 +1,209 @@
+# RFC: `vp check` Command
```
fengmk2, line 1: @cpojer Could you please check if it matches your expected functionality?

rfcs/check-command.md
```diff
@@ -0,0 +1,209 @@
...
+```
+1. vp fmt --check                          (verify formatting, don't auto-fix)
+2. vp lint --type-aware --type-check       (lint + type checking)
+```
+
+If any step fails, `vp check` exits immediately with a non-zero exit code.
+
+### Future (Parallel)
+
+In a later iteration, steps could run in parallel and aggregate errors.
```
camc314: I suggest we don't parallelize this - it's going to lead to FS contention as well as CPU contention - both oxfmt & oxlint will be using all available threads.

cpojer: Agreed. Format first + lint after.

PR #648, opened 2026-02-27
chore: Switch to MIT License.

cpojer:
This PR switches Vite+ to MIT license.

cc @yyx990803

### DIFF
--- a/packages/core/build.ts
+++ b/packages/core/build.ts
@@ -51,6 +51,7 @@ await buildVite();
 await bundleTsdown();
 await bundleVitepress();
 await mergePackageJson();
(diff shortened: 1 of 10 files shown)


packages/core/build.ts
```diff
@@ -51,6 +51,7 @@ await buildVite();
 await bundleTsdown();
 await bundleVitepress();
 await mergePackageJson();
+await syncLicenseFromRoot();
```
fengmk2, line 54: @cpojer The `test` package also needs to copy the license file.

cpojer: https://github.com/voidzero-dev/vite-plus/pull/651

PR #652, opened 2026-02-27
fix: PowerShell 5.1 compatibility for install.ps1

fengmk2:
- Replace 3-argument Join-Path calls with nested 2-argument calls
(PS5 only supports 2 arguments)
- add UTF-8 BOM to install.ps1 for PowerShell 5.1 compatibility
- Add PowerShell 5.1 CI job using `shell: powershell` to prevent
regressions

closes https://github.com/voidzero-dev/vite-plus-discussions/issues/14

### DIFF
--- a/packages/cli/install.ps1
+++ b/packages/cli/install.ps1
@@ -1,4 +1,4 @@
-# Vite+ CLI Installer for Windows
+﻿# Vite+ CLI Installer for Windows
 # https://viteplus.dev/install.ps1
 #
 # Usage:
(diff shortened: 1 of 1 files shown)


packages/cli/install.ps1
```diff
@@ -382,7 +382,7 @@ exec "`$VITE_PLUS_HOME/current/bin/vp.exe" "`$@"
 
     # Print success message
     Write-Host ""
-    Write-Host "${GREEN}✔${NC} ${BOLD_BRIGHT_BLUE}VITE+${NC} successfully installed!"
+    Write-Host "${GREEN}+${NC} ${BOLD_BRIGHT_BLUE}VITE+${NC} successfully installed!"
```
cpojer: Is this change expected?

fengmk2: yes, PS 5.1 don't support utf-8 https://github.com/voidzero-dev/vite-plus-discussions/issues/14  
I'll verify.

fengmk2: It will indeed report an error. https://github.com/voidzero-dev/vite-plus/actions/runs/22476203552/job/65103581586

packages/cli/install.ps1
```diff
@@ -1,4 +1,4 @@
-# Vite+ CLI Installer for Windows
```
fengmk2, line 1: add UTF-8 BOM to install.ps1 for PowerShell 5.1 compatibility

PR #656, opened 2026-02-27
feat(release): add Windows ARM64 (aarch64-pc-windows-msvc) support

fengmk2:
## Summary

- Add `aarch64-pc-windows-msvc` as a build target in the release pipeline, NAPI config, and publish scripts
- Add install test job for Windows ARM64 (`windows-11-arm` runner)
- Document supported platforms in the getting started guide

https://github.com/voidzero-dev/vite-plus-discussions/issues/13

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,6 +52,8 @@ jobs:
             os: ubuntu-latest
           - target: x86_64-pc-windows-msvc
             os: windows-latest
+          - target: aarch64-pc-windows-msvc
+            os: windows-latest
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -64,14 +66,6 @@ jobs:
       - name: Rustup Adds Target
         run: rustup target add ${{ matrix.settings.target }}
 
-      - name: Add musl target (x86_64)
```
Brooooooklyn, line 67: I think fspy need this

Brooooooklyn: @branchseer Do we still need to have the musl target installed for Linux targets?

wan9chi: Only required for fspy tests. Can be safely removed in vite-plus

PR #668, opened 2026-03-02
chore: CLI improvements.

cpojer:
This PR introduces general helpers to unify the printing of CLI output in Rust and in JS. As you can see from the snapshots, it already aligns a bunch of printing that was previously looking differently. It figures out the widths dynamically, which compresses output quite a bit and saves vertical space.

This was 100% done by Codex, but I think it's mostly good.

Later, I will improve the overall printing style.

### DIFF
--- a/packages/tools/src/utils.ts
+++ b/packages/tools/src/utils.ts
@@ -3,7 +3,16 @@ import path from 'node:path';
 
 import { Minimatch } from 'minimatch';
 
+const ANSI_ESCAPE_REGEX = new RegExp(
+  `${String.fromCharCode(27)}(?:[@-Z\\\\-_]|\\[[0-?]*[ -/]*[@-~])`,
+  'g',
+);
+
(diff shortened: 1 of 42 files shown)


packages/tools/src/utils.ts
```diff
@@ -3,7 +3,16 @@ import path from 'node:path';
...
 
+const ANSI_ESCAPE_REGEX = new RegExp(
+  `${String.fromCharCode(27)}(?:[@-Z\\\\-_]|\\[[0-?]*[ -/]*[@-~])`,
+  'g',
+);
+
 export function replaceUnstableOutput(output: string, cwd?: string) {
+  // Normalize line endings and strip ANSI escapes so snapshots are stable
+  // across CI platforms and terminal capabilities.
+  output = output.replaceAll(ANSI_ESCAPE_REGEX, '').replaceAll(/\r\n/g, '\n').replaceAll(/\r/g, '');
```
fengmk2, line 14: I love this!

PR #680, opened 2026-03-04
chore: Unify the "Vite+" header in Rust and JS.

cpojer:
This PR:

* Moves the code for printing the header and gradient to Rust, and exposes it back to JS.
* Adds the header to most commands such as `vp check`, giving Vite+ more of an identity (and likely makes it easier for LLMs to remember what they are using).
* Changes the header's purple color to "magenta" as defined by the user, and uses that for the gradient. The gradient was inverted, and VITE+ is now magenta. Since the gradient now goes from two user defined colors (foreground and magenta), it should work in all sort of terminal setups.

CleanShot 2026-03-04 at 10 05 50@2x


crates/vite_global_cli/src/cli.rs
```diff
@@ -1478,6 +1477,7 @@ pub async fn run_command(cwd: AbsolutePathBuf, args: Args) -> Result<ExitStatus,
             packages,
             pass_through_args,
         } => {
+            print_runtime_header();
```
fengmk2, line 1480: Do all commands print a header? In some commands that don't produce any output, might this lead to user misunderstanding? For example, `vp fmt`.

cpojer: I added it to as many as I could find. It looks good on commands like `vp check`, but yeah `vp fmt` is maybe not ideal?

CleanShot 2026-03-04 at 10 18 53@2x

CleanShot 2026-03-04 at 10 18 56@2x

What do you think we should do here?

cpojer: (Decided to keep as-is for now, and possibly change it later)

Issue #677, opened 2026-03-03
Init IDE configs


Boshen:
IDE configs such as VSCode require some really complicated json. Vite+ should ship better integration with it.

PR #682, opened 2026-03-04
chore: Add rfcs/init-editor-configs

leaysgur:
Part of #677 

I wrote an RFC following the project's guideline(?).

I would like it reviewed before implementation.


rfcs/init-editor-configs.md
```diff
@@ -0,0 +1,138 @@
...
+Based on [oxc-vscode's own `.vscode/settings.json`](https://github.com/oxc-project/oxc-vscode/blob/main/.vscode/settings.json).
+
+```json
+{
+  "editor.defaultFormatter": "oxc.oxc-vscode",
+  "editor.formatOnSave": true,
+  "editor.formatOnSaveMode": "file",
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": "never"
+  },
```
cpojer, line 48: I would change this to:

```
"editor.codeActionsOnSave": {
  "source.fixAll.oxc": "explicit"
},
```

We should run all fixers AND organize imports when users save imho.

rfcs/init-editor-configs.md
```diff
@@ -0,0 +1,138 @@
...
+## Behavior
+
+### Existing file handling
+
+When a config file already exists:
+
+- **Interactive mode**: Prompt with Merge / Skip options
+  - Merge: Add new keys without overwriting existing user settings. For `extensions.json`, deduplicate recommendations array.
+  - Skip: Leave unchanged
+- **Non-interactive mode**: Skip silently with info log
```
cpojer: Why not keep the behaviors the same in both (merge)?

leaysgur: I intended for it to behave the same as `--agent`, but since your point seems fine, I'll change it.

PR #691, opened 2026-03-04
fix(docs): pin @voidzero-dev/vitepress-theme to exact version

fengmk2:
Pinning the VitePress theme to an exact version provides more predictable builds and prevents potential breaking changes from patch updates. The icon package update likely includes new icons or fixes for existing ones.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -14,7 +14,7 @@
     "vue3-carousel": "^0.16.0"
   },
   "devDependencies": {
-    "@voidzero-dev/vitepress-theme": "^4.3.0",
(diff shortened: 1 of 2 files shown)


docs/package.json
```diff
@@ -14,7 +14,7 @@
     "vue3-carousel": "^0.16.0"
   },
   "devDependencies": {
-    "@voidzero-dev/vitepress-theme": "^4.3.0",
+    "@voidzero-dev/vitepress-theme": "4.3.0",
```
fengmk2, line 17: @mdong1909 the latest theme missing some files

Image

PR #732, opened 2026-03-09
chore: Update website theme.

cpojer:
Updates the theme of the website to the latest version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1394,7 +1394,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


pnpm-workspace.yaml
```diff
@@ -40,6 +41,7 @@ catalog:
   '@vitejs/plugin-react': ^5.1.2
   '@vitest/browser': ^4.0.9
   '@vitest/browser-playwright': ^4.0.9
+  '@voidzero-dev/vitepress-theme': 4.7.1
```
mdong1909: Lets update to 4.8.0. There was a new version that was just released

cpojer: Oops, done!

docs/package.json
```diff
@@ -8,13 +8,14 @@
     "preview": "vitepress preview"
   },
   "dependencies": {
+    "@iconify/vue": "catalog:",
```
fengmk2, line 11: @cpojer Sorry, I forgot again in the previous review that our current deployment on netlify uses `npm install` 😢 , so currently the dependencies under `docs/` need to be written as version and cannot use pnpm's `catalog:` https://github.com/voidzero-dev/vite-plus/blob/main/netlify.toml#L3

cpojer: https://github.com/voidzero-dev/vite-plus/pull/737

PR #733, opened 2026-03-09
fix: sync run-config.ts and add vite-plus-cli to CI tests

wan9chi:
## Summary
- Sync `run-config.ts` with upstream vite-task to add the new `inputs` cache config field
- Add `vite-plus-cli` to `cargo test` in CI so the `run_config_types_in_sync` test is actually run — previously it was silently skipped because CI only tested `crates/*` packages, and `vite-plus-cli` lives at `packages/cli/binding/`

## Test plan
- [x] `cargo test -p vite-plus-cli run_config_types_in_sync` passes
- [ ] CI passes

https://claude.ai/code/session_01RSBPa1P1A94WvPMDa8YaQE

### DIFF
--- a/packages/cli/binding/src/exec/workspace.rs
+++ b/packages/cli/binding/src/exec/workspace.rs
@@ -348,7 +348,7 @@ mod tests {
 
     /// Create a cross-platform absolute path for tests.
     /// On Unix `/workspace/...`, on Windows `C:\workspace\...`.
-    fn test_absolute_path(suffix: &str) -> Arc<AbsolutePathBuf> {
+    fn test_absolute_path(suffix: &str) -> Arc<vite_path::AbsolutePath> {
         #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -92,7 +92,8 @@ jobs:
           RUSTFLAGS: '-D warnings --cfg tokio_unstable' # also update .cargo/config.toml
 
       # Test all crates/* packages. New crates are automatically included.
-      - run: cargo test $(for d in crates/*/; do echo -n "-p $(basename $d) "; done)
+      # Also test vite-plus-cli (lives outside crates/) to catch type sync issues.
+      - run: cargo test $(for d in crates/*/; do echo -n "-p $(basename $d) "; done) -p vite-plus-cli
```
fengmk2, line 96: 💯

PR #736, opened 2026-03-09
chore: Improve `vp create` output.

cpojer:
We are printing way too much stuff that nobody needs to see in `vp create`. This PR moves the pre-commit hook question to before it runs any commands, suppresses all the output, fixes inconsistent newline rendering, and prints a clean summary that matches the output from our website about the `vp create` command.

Example:
CleanShot 2026-03-09 at 15 23 42@2x

I will do a pass on `vp migrate` later as well.

### DIFF
--- a/packages/cli/src/create/templates/remote.ts
+++ b/packages/cli/src/create/templates/remote.ts
@@ -7,6 +7,7 @@ import {
   formatDlxCommand,
   runCommand,
   runCommandAndDetectProjectDir,
+  runCommandSilently,
 } from '../command.js';
 import type { TemplateInfo } from './types.js';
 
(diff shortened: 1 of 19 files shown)


packages/cli/src/create/templates/remote.ts
```diff
@@ -62,19 +70,25 @@ export async function runRemoteTemplateCommand(
...
+    prompts.log.info(`Running: ${gray(`${command} ${args.join(' ')}`)}`);
+  }
   if (detectCreatedProjectDir) {
     return await runCommandAndDetectProjectDir(
       { command, args, cwd, envs },
       templateInfo.parentDir,
     );
   }
+  if (silent) {
+    return await runCommandSilently({ command, args, cwd, envs });
```
fengmk2, line 90: Image

When the template is `vite:library`, it cannot be installed silently here. Currently, create-tsdown only supports interactive mode calls; otherwise, it will hang. I will temporary fix it by adding a default parameter.

PR #742, opened 2026-03-09
feat(migrate): Prettier → Oxfmt migration via vp fmt --migrate=prettier

fengmk2:
Add Prettier → Oxfmt migration support to `vp migrate`, following the
same architecture as the existing ESLint → Oxlint migration: Rust shell
script rewriter → NAPI binding → TypeScript orchestration.

- Rust rewriter rewrites `prettier` → `vp fmt`, strips Prettier-only
  flags (--write, --cache, --single-quote, --config, --plugin, etc.),
  and converts --list-different/-l → --check
- Detects all Prettier config files (.prettierrc*, prettier.config.*,
  package.json#prettier) and .prettierignore
- Removes prettier + prettier-plugin-* dependencies
- Rewrites scripts in package.json and lint-staged configs
- Supports both fresh migration and re-migration (already-vite-plus)
- Includes snap test fixtures for basic, rerun, lint-staged, and
  eslint+prettier combo scenarios

### DIFF
--- a/crates/vite_migration/src/prettier.rs
+++ b/crates/vite_migration/src/prettier.rs
@@ -5,6 +5,7 @@ const PRETTIER_CONFIG: ScriptRewriteConfig = ScriptRewriteConfig {
     target_subcommand: "fmt",
     boolean_flags: &[
         "--write",
+        "-w",
         "--cache",
         "--no-config",
         "--no-editorconfig",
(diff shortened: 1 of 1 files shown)


crates/vite_migration/src/prettier.rs
```diff
@@ -0,0 +1,533 @@
...
+                return false;
+            }
+        }
+    }
+    true
+}
+
+/// Fix pipe spacing in brush-parser Display output.
+/// brush-parser renders pipes as `cmd1 |cmd2` instead of `cmd1 | cmd2`.
+fn normalize_pipe_spacing(s: &str) -> String {
```
fengmk2: There is duplicate code with [eslint.rs](http://eslint.rs)

PR #751, opened 2026-03-10
fix(migrate): use latest @oxlint/migrate instead of pinning to oxlint version

fengmk2:
## Summary

- Use `@oxlint/migrate` (latest) instead of pinning to the installed oxlint version
- `@oxlint/migrate` publishes less frequently than `oxlint`, so when the catalog's `^1.51.0` resolves to `1.52.0` in a fresh install (no lockfile), `@oxlint/migrate@1.52.0` doesn't exist and the migration fails

## Test plan

- [ ] CI e2e tests pass (vue-mini eslint migration no longer fails)

### DIFF
--- a/packages/cli/src/migration/migrator.ts
+++ b/packages/cli/src/migration/migrator.ts
@@ -1,5 +1,4 @@
 import fs from 'node:fs';
-import { createRequire } from 'node:module';
 import path from 'node:path';
 
 import * as prompts from '@voidzero-dev/vite-plus-prompts';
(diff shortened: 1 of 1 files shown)


packages/cli/src/migration/migrator.ts
```diff
@@ -147,7 +147,8 @@ export function detectEslintProject(
 }
 
 function getInstalledOxlintVersion(): string {
-  const oxlintMainPath = resolve('oxlint');
+  const require = createRequire(import.meta.url);
```
fengmk2: e2e ci fails

PR #766, opened 2026-03-10
fix: resolve `vp run` crash when task script contains global CLI subcommands

fengmk2:
## Summary

- `vp run` crashed when a task script contained global-only commands like `vp config` because `CLIArgs::try_parse_from` failed on unrecognized subcommands with `ErrorKind::InvalidSubcommand`
- Now `InvalidSubcommand` errors return `HandledCommand::Synthesized` with `UserCacheConfig::disabled()`, letting the global CLI binary handle them as regular subprocesses without caching
- Also disabled cache for `vp exec` and `vp check` in task scripts, since they run arbitrary/composite commands with side effects

## Test plan

- [x] Added unit test `global_subcommands_produce_invalid_subcommand_error` confirming `config`, `create`, `env`, `migrate` produce `InvalidSubcommand`
- [x] Added snap test `command-run-with-vp-config` verifying `vp run foo` with `"foo": "vp config"` no longer crashes
- [x] All existing snap tests pass

### DIFF
--- a/packages/cli/binding/src/cli.rs
+++ b/packages/cli/binding/src/cli.rs
@@ -676,14 +676,18 @@ impl CommandHandler for VitePlusCommandHandler {
         ) {
             Ok(args) => args,
             Err(err) if err.kind() == ErrorKind::InvalidSubcommand => {
-                return Ok(HandledCommand::Verbatim);
+                return Ok(HandledCommand::Synthesized(
+                    command.to_synthetic_plan_request(UserCacheConfig::disabled()),
+                ));
             }
             Err(err) => return Err(err.into()),
         };
         match cli_args {
             CLIArgs::Synthesizable(SynthesizableSubcommand::Check { .. }) => {
                 // Check is a composite command — run as a subprocess in task scripts
-                Ok(HandledCommand::Verbatim)
+                Ok(HandledCommand::Synthesized(
+                    command.to_synthetic_plan_request(UserCacheConfig::disabled()),
+                ))
             }
             CLIArgs::Synthesizable(subcmd) => {
                 let resolved = self.resolver.resolve(subcmd, &command.envs, &command.cwd).await?;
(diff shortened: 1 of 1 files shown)


packages/cli/binding/src/cli.rs
```diff
@@ -671,8 +671,15 @@ impl CommandHandler for VitePlusCommandHandler {
...
         // Parse "vp <args>" using CLIArgs — always use "vp" as the program name
         // so clap shows "Usage: vp ..." even if the original command was "vite ..."
-        let cli_args =
-            CLIArgs::try_parse_from(iter::once("vp").chain(command.args.iter().map(Str::as_str)))?;
+        let cli_args = match CLIArgs::try_parse_from(
+            iter::once("vp").chain(command.args.iter().map(Str::as_str)),
+        ) {
+            Ok(args) => args,
+            Err(err) if err.kind() == ErrorKind::InvalidSubcommand => {
+                return Ok(HandledCommand::Verbatim);
```
wan9chi: Should we disable cache for unknown subcommands (and maybe also for exec/check)? You can check handler of 'dev' for how to disable cache.

fengmk2: let me see

packages/cli/binding/src/cli.rs
```diff
@@ -671,12 +671,23 @@ impl CommandHandler for VitePlusCommandHandler {
...
         // so clap shows "Usage: vp ..." even if the original command was "vite ..."
-        let cli_args =
-            CLIArgs::try_parse_from(iter::once("vp").chain(command.args.iter().map(Str::as_str)))?;
+        let cli_args = match CLIArgs::try_parse_from(
+            iter::once("vp").chain(command.args.iter().map(Str::as_str)),
+        ) {
+            Ok(args) => args,
+            Err(err) if err.kind() == ErrorKind::InvalidSubcommand => {
+                return Ok(HandledCommand::Synthesized(
+                    command.to_synthetic_plan_request(UserCacheConfig::disabled()),
```
fengmk2, line 680: @branchseer take a look again.

PR #785, opened 2026-03-12
docs: Vite Task docs updates.

cpojer:
Updating docs for pending Vite Task API changes.

### DIFF
--- a/crates/vite_global_cli/src/command_picker.rs
+++ b/crates/vite_global_cli/src/command_picker.rs
@@ -304,10 +304,7 @@ fn render_picker(
     let pad = if vite_shared::header::is_warp_terminal() { " " } else { "" };
     let max_width = usize::from(columns).saturating_sub(4 + pad.len());
     let viewport_end = (viewport_start + viewport_size).min(filtered_indices.len());
-    let instruction = truncate_line(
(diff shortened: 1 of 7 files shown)


docs/guide/run.md
```diff
@@ -87,7 +88,7 @@ export default defineConfig({
 });
 ```
 
-If `command` is omitted, the task uses the matching script from `package.json`. A task name can come from `vite.config.ts` or `package.json`, but not both.
+If you want to run an existing `package.json` script as-is, use `vp run <script>`. If you want task-level caching, dependencies, or environment/input controls, define a task with an explicit `command`. A task name can come from `vite.config.ts` or `package.json`, but not both.
 
 ::: info
 Tasks defined in `vite.config.ts` are cached by default. Plain `package.json` scripts without a matching task entry are not. See [When Is Caching Enabled?](/guide/cache#when-is-caching-enabled) for the full resolution order.
```
wan9chi: "Plain `package.json` scripts without a matching task entry" doesn't make too much sense now.

wan9chi: There is also a similar saying in the "When Is Caching Enabled?" section.

cpojer: Fixed!

PR #789, opened 2026-03-12
fix(cli): gracefully handle missing vp fmt during config initialization

fengmk2:
The `vpFmt` function now gracefully handles cases where `vp fmt` is not available or fails to execute. Instead of throwing an error when formatting fails, it now shows a warning message and continues execution. Additionally, the entire formatting operation is wrapped in a try-catch block to handle cases where `vp fmt` is not yet installed (such as during migration processes), allowing the operation to proceed without formatting rather than failing completely.

### DIFF
--- a/packages/cli/src/init-config.ts
+++ b/packages/cli/src/init-config.ts
@@ -170,7 +170,9 @@ async function vpFmt(cwd: string, filePath: string): Promise<void> {
     },
   });
   if (result.exitCode !== 0) {
-    warnMsg(`Failed to format ${filePath} with vp fmt`);
+    warnMsg(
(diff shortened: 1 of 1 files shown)


packages/cli/src/init-config.ts
```diff
@@ -170,9 +170,7 @@ async function vpFmt(cwd: string, filePath: string): Promise<void> {
     },
   });
   if (result.exitCode !== 0) {
-    throw new Error(
-      `Failed to format ${filePath} with vp fmt:\n${result.stdout.toString()}${result.stderr.toString()}`,
```
fengmk2, line 174: should keep the stdout and stderr output.

PR #795, opened 2026-03-12
feat(deps): upgrade oxlint to 1.54.0 and oxfmt to 0.39.0

fengmk2:
Remove forced `fmt: {}` injection that was required by older oxfmt versions.
The new oxfmt/oxlint releases no longer require `lint` and `fmt` fields in
vite.config.ts, so `injectFmtDefaultsIfMissing()` and `injectFmtDefaults()`
are removed along with empty placeholder fields in snap test fixtures.

### DIFF
--- a/packages/cli/snap-tests/cache-clean/snap.txt
+++ b/packages/cli/snap-tests/cache-clean/snap.txt
@@ -1,12 +1,10 @@
 > vp run hello # create the cache for 'echo hello'
 $ vp fmt
-No config found, using defaults. Please add a config file or try `oxfmt --init` if needed.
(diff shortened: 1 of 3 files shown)


packages/cli/snap-tests/cache-clean/snap.txt
```diff
@@ -1,10 +1,12 @@
 > vp run hello # create the cache for 'echo hello'
 $ vp fmt
+No config found, using defaults. Please add a config file or try `oxfmt --init` if needed.
```
fengmk2: :grinning: don't fix this for now.

PR #814, opened 2026-03-13
fun: Fun mode.

cpojer:
Gotta have a little fun.


packages/cli/assets/error.png
yangshun, line 1: This image is 1.77MB, not exactly small. Might be better if the file size can be reduced

cpojer: That's your concern with this PR?

Issue #826, opened 2026-03-13
Support editor setting for  Zed


xingwangzhe:
### Description

<img width="520" height="511" alt="Image" src="https://github.com/user-attachments/assets/2db42b00-9d76-4333-bea8-4c78d37f4120" />

Zed is probably now the most popular code editor after VSCode。

### Suggested solution

`.zed/settings.json`

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

shinebayar-g:
@alexanderzinolidis `oxc-zed` [v0.4.5](https://github.com/oxc-project/oxc-zed/releases/tag/v0.4.5) was released yesterday which includes https://github.com/oxc-project/oxc-zed/pull/141 which should have fixed this issue. I haven't had a time to check it yet, but you're probably using the new version.

Update: `vite-plus` is now working with the oxc-zed extension.

silvenon:
Indeed, v0.4.6 fixed it! Initially it wasn't working for me so I thought it was still broken, but running `vp lint` revealed that the Vite Plus was failing to resolve `vite.config.ts` itself because for some reason I have to specify extensions for local modules.

I'd prefer Oxc LSP to fail if the specified `configFile` failed to resolve, so that I know when it stopped working, but I assume this is a fallback feature.

patspam:
Would be nice if the instructions said that you need to install the `oxc-zed` extension to get it to work properly

naokihaba:
@fengmk2 
The underlying technical issue has already been resolved in oxc-zed v0.4.6, and vp create/vp migrate now automatically generates .zed/settings.json as of #832. What's still missing is documentation on the IDE Integration page covering the Zed setup — including installing the oxc-zed extension and the manual config using configPath: "./vite.config.ts". I'd like to open a PR for that.

shinebayar-g:
@naokihaba fyi https://github.com/voidzero-dev/vite-plus/pull/832 generates different config than the example mentioned in the zed oxc extension repository https://github.com/oxc-project/oxc-zed/blob/main/examples/both/.zed/settings.json

[...]

PR #832, opened 2026-03-13
feat(cli): add Zed editor support for project creation

xingwangzhe:
The `.zed/settings.json` is from [oxc-project/oxc-zed](https://github.com/oxc-project/oxc-zed/blob/main/examples/both/.zed/settings.json) with some modifications,like ` configPath: './ox(fmt|lint).jsonc'

#826

### DIFF
--- a/packages/cli/install.ps1
+++ b/packages/cli/install.ps1
@@ -316,13 +316,23 @@ function Main {
     } | ConvertTo-Json -Depth 10
     Set-Content -Path (Join-Path $VersionDir "package.json") -Value $wrapperJson
 
+    # Isolate from user's global package manager config that may block
+    # installing recently-published packages (e.g. pnpm's minimumReleaseAge,
+    # npm's min-release-age) by creating a local .npmrc in the version directory.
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/editor.ts
```diff
@@ -21,6 +21,119 @@ const VSCODE_EXTENSIONS = {
   recommendations: ['VoidZero.vite-plus-extension-pack'],
 } as const;
 
+const ZED_SETTINGS = {
+  lsp: {
+    oxlint: {
+      initialization_options: {
+        settings: {
+          configPath: './.oxlintrc.json',
```
fengmk2, line 29: @xingwangzhe configPath should be `./vite.config.ts`?

xingwangzhe: https://github.com/oxc-project/oxc/blob/main/.zed%2Fsettings.json

写的是oxc那两个工具配置文件的路径，至少官方是这样的

uncvrd: @fengmk2 I think you might be right? a little confused why we're referencing the .json here when that file doesn't exist, this was after running vp create for a monorepo. 

image

PR #834, opened 2026-03-13
fix(cli): show errors when vp install fails and bypass npm & pnpm minimum release age

### LINKED ISSUE #833: `curl -fsSL https://vite.plus | bash` fails silently when `vp install` encounters an error
### Describe the bug

Fresh installation via `curl -fsSL https://vite.plus | bash` fails silently, no error is shown to the user, but the installation is incomplete. 
`node_modules` is never created, ~/.vite-plus/bin/ is never set up, and vp is not added to PATH.

The direct trigger in my case was pnpm's minimumReleaseAge setting (configured in my global `~/.config/pnpm/rc`), which blocked installing the recently-published vite-plus package. However, the underlying issue is that install.sh runs vp install --silent (line 580), which suppresses all error output. 

When this step fails, set -e aborts the script and all subsequent setup steps are skipped, with no feedback to the user about what went wrong

### Reproduction

No URL

### Steps to reproduce

local reproduction

1. Ensure pnpm global config has minimum-release-age=1440 (or any setting that would cause pnpm install to fail for a recently-published package):
```
# Check: cat ~/.config/pnpm/rc
minimum-release-age=144
```
2. Clean any existing installation: `rm -rf ~/.vite-plus`
3. Run: `curl -fsSL https://vite.plus | bash`
4. Observe no error is shown, but installation is incomplete:
```sh
ls ~/.vite-plus/bin/            # No such file or directory
ls ~/.vite-plus/current         # No such file or directory
ls ~/.vite-plus/env             # No such file or directory
ls ~/.vite-plus/*/node_modules/ # No such file or directory
```
5. Running the binary directly shows the consequence:
```
Error: Cannot find module '.../.vite-plus/<version>/node_modules/vite-plus/dist/bin.js'
```

To see the actual error that `--silent` hides, run pnpm manually against the generated wrapper `package.json`:
```
cd ~/.vite-plus/0.1.11
npx pnpm@10.32.1 install
# ERR_PNPM_NO_MATURE_MATCHING_VERSION  Version 0.1.11 (released X hours ago)
# of vite-plus does not meet the minimumReleaseAge constraint
```

### System Info

```shell
OS: macOS (Darwin 24.5.0, arm64)
Shell: zsh
Node.js: v24.14.0
pnpm (auto-selected by installer): v10.32.1
vite-plus version: 0.1.11
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

kazupon:
## Summary
resolve #833 

- Remove `--silent` from `vp install` in `install.sh` so that pnpm/npm errors are visible to the user instead of being silently swallowed
- Add a local `.npmrc` with `minimum-release-age=0` to the version directory to isolate the installer from the user's global pnpm configuration (e.g. `minimumReleaseAge` blocking recently-published packages)

Previously, when `vp install --silent` failed (for any reason), `set -e` would abort `install.sh` with no error output, leaving `~/.vite-plus/` in a partial state: no `node_modules`, no `bin/` symlink, no `env` file, and no PATH configuration.

### DIFF
--- a/packages/cli/install.sh
+++ b/packages/cli/install.sh
@@ -603,9 +603,10 @@ NPMRC_EOF
   # Install production dependencies (skip if VITE_PLUS_SKIP_DEPS_INSTALL is set,
   # e.g. during local dev where install-global-cli.ts handles deps separately)
   if [ -z "${VITE_PLUS_SKIP_DEPS_INSTALL:-}" ]; then
-    if ! (cd "$VERSION_DIR" && CI=true "$BIN_DIR/vp" install 2>&1); then
-      error "Failed to install dependencies. Re-run without piping to see full output:
-    curl -fsSL https://vite.plus -o /tmp/vp-install.sh && bash /tmp/vp-install.sh"
+    local install_log="$VERSION_DIR/install.log"
+    if ! (cd "$VERSION_DIR" && CI=true "$BIN_DIR/vp" install --silent > "$install_log" 2>&1); then
+      error "Failed to install dependencies. See log for details: $install_log"
+      exit 1
     fi
   fi
 
--- a/packages/cli/install.ps1
+++ b/packages/cli/install.ps1
@@ -324,12 +324,13 @@ function Main {
     # Install production dependencies (skip if VITE_PLUS_SKIP_DEPS_INSTALL is set,
     # e.g. during local dev where install-global-cli.ts handles deps separately)
     if (-not $env:VITE_PLUS_SKIP_DEPS_INSTALL) {
+        $installLog = Join-Path $VersionDir "install.log"
         Push-Location $VersionDir
         try {
             $env:CI = "true"
-            & "$BinDir\vp.exe" install
+            & "$BinDir\vp.exe" install --silent *> $installLog
             if ($LASTEXITCODE -ne 0) {
-                Write-Error "Failed to install dependencies. Please check the error output above."
+                Write-Host "error: Failed to install dependencies. See log for details: $installLog" -ForegroundColor Red
                 exit 1
             }
         } finally {


packages/cli/install.sh
```diff
@@ -592,10 +592,21 @@ main() {
...
+  cat > "$VERSION_DIR/.npmrc" <<NPMRC_EOF
+minimum-release-age=0
+min-release-age=0
+NPMRC_EOF
+
   # Install production dependencies (skip if VITE_PLUS_SKIP_DEPS_INSTALL is set,
   # e.g. during local dev where install-global-cli.ts handles deps separately)
   if [ -z "${VITE_PLUS_SKIP_DEPS_INSTALL:-}" ]; then
-    (cd "$VERSION_DIR" && CI=true "$BIN_DIR/vp" install --silent)
+    if ! (cd "$VERSION_DIR" && CI=true "$BIN_DIR/vp" install 2>&1); then
```
fengmk2: Image

Can the logs from vp install be forwarded to a log file? And only print the logs if execution fails. Currently this causes some unwanted execution logs to be exposed, which is also the reason why the `--silent` parameter was added before.

packages/cli/install.sh
```diff
@@ -592,10 +592,21 @@ main() {
...
+      error "Failed to install dependencies. Re-run without piping to see full output:
+    curl -fsSL https://vite.plus -o /tmp/vp-install.sh && bash /tmp/vp-install.sh"
```
fengmk2: this line is for debug?

packages/cli/install.sh
kazupon: I've kept `--silent` to suppress unnecessary logs during normal installation, and redirect all `vp install` output to a log file (`$VERSION_DIR/install.log`). On success, nothing is shown to the user. On failure, we show an error message with the log file path so users (and us) can inspect it for debugging.

packages/cli/install.sh
kazupon: I've removed the `curl` re-run suggestion, showing the log file path is more useful for both users and our bug investigation than asking them to re-run the installer.

Issue #836, opened 2026-03-13
Improve shell configuring for Vite+ installer


CyanChanges:
### Description

I use Nix home-manager to manage my fish config so the config file is read-only.
The installer failed to update the config file. 

<img width="725" height="79" alt="Image" src="https://github.com/user-attachments/assets/6da38f00-0ff8-4670-8224-ee062aae53a2" />

### Suggested solution

For fish, have something like `~/.config/fish/conf.d/vite-plus.fish` would be nice.(https://fishshell.com/docs/current/language.html#configuration, fish read snippets in ~/.config/fish/conf.d/ by default)
No need to update the main config.fish & make things cleaner

I guess it would be better to ask before making any changes to the shell configs too?
for example, ask if the user want the script to update automatically, or if they want to do it manually

### Additional context

<img width="1386" height="49" alt="Image" src="https://github.com/user-attachments/assets/d4c714db-0da6-41e8-9ed8-b6e671b1e5db" />

https://github.com/voidzero-dev/vite-plus/blob/2223726fa8945dc0c71a216284874c97c41608f7/packages/cli/install.sh#L393-L407

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

PR #837, opened 2026-03-13
feat(cli): create a snippet instead of updating config.fish for fish shell in install.sh

CyanChanges:
Use `~/.config/fish/conf.d/vite-plus.fish` instead of updating `config.fish`

ref: https://github.com/voidzero-dev/vite-plus/issues/836

### DIFF
--- a/packages/cli/src/utils/editor.ts
+++ b/packages/cli/src/utils/editor.ts
@@ -21,6 +21,119 @@ const VSCODE_EXTENSIONS = {
   recommendations: ['VoidZero.vite-plus-extension-pack'],
 } as const;
 
+const ZED_SETTINGS = {
+  lsp: {
+    oxlint: {
+      initialization_options: {
+        settings: {
+          configPath: './.oxlintrc.json',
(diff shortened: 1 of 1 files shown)


packages/cli/install.sh
```diff
@@ -389,22 +389,14 @@ configure_shell_path() {
...
-        else
-          echo "" >> "$fish_config"
-          echo "# Vite+ bin (https://viteplus.dev)" >> "$fish_config"
-          echo "source \"$INSTALL_DIR_REF/env.fish\"" >> "$fish_config"
-          result=0
-          SHELL_CONFIG_UPDATED="config.fish"
-        fi
+        result=2
+      else
+        echo "# Vite+ bin (https://viteplus.dev)" >> "$fish_config"
```
fengmk2, line 398: Will the directory `$HOME/.config/fish/conf.d` definitely exist?

cyans-nya: oh, i forget to cover that
1sec

PR #953, opened 2026-03-16
fix(justfile): add Windows support, extract clean-dist

nekomoyi:
Fix Windows compatibility for `just init`.

Add platform-specific recipes for cleaning dist directories:
- Unix: `rm -rf`
- Windows: PowerShell `Remove-Item`


fengmk2:
@nekomoyi Thank you for helping us with our development experience on Windows 🪟

PR #961, opened 2026-03-16
fix(cli): split IDE and shell profile checks in `vp env doctor`

### LINKED ISSUE #881: fix(env): `vp env doctor` misses most shell profiles when checking IDE integration
### Description

The install script (`install.sh`) writes the env sourcing line to multiple shell profiles:
- **zsh**: `.zshenv` + `.zshrc`
- **bash**: `.bash_profile` + `.bashrc` + `.profile`

But `vp env doctor`'s `check_profile_files` only checks:
- **macOS**: `.zshenv`, `.profile`
- **Linux**: `.profile` only

So on Linux with zsh (env in `.zshrc`) or bash (env in `.bashrc`), the doctor reports IDE integration is not set up and shows unnecessary setup guidance — even though everything is correctly configured.

### Expected behavior

`check_profile_files` should check all profiles the install script may have written to: `.zshenv`, `.zshrc`, `.bash_profile`, `.bashrc`, `.profile`. This also matches the set that `vp implode` already cleans up.

### Location

`crates/vite_global_cli/src/commands/env/doctor.rs:460-467`

jong-kyung:
## Summary

- Introduce tri-state `EnvSourcingStatus` (`IdeFound` / `ShellOnly` / `NotFound`) to replace the boolean `check_ide_integration()` return
- Check IDE-visible profiles (`.zshenv`, `.profile`) first, then fall back to all shell profiles (`.bashrc`, `.zshrc`, `.bash_profile`, fish configs)
- Add proper fish shell support with `env.fish` pattern matching via `is_fish` flag
- Parameterize `check_profile_files()` to accept profile list as argument for reuse

This avoids the false positives that caused #878 to be rejected (marking `.bashrc` as IDE-OK) while eliminating the false negatives from #881 (ignoring shell-only sourcing entirely).

Closes #881

### DIFF
--- a/crates/vite_global_cli/src/commands/env/doctor.rs
+++ b/crates/vite_global_cli/src/commands/env/doctor.rs
@@ -578,61 +578,53 @@ fn check_profile_files(
 }
 
 /// Print IDE setup guidance for GUI applications.
+#[cfg(not(windows))]
 fn print_ide_setup_guidance(bin_dir: &vite_path::AbsolutePath) {
-    // On Windows, IDE PATH is handled by System Environment Variables (covered by check_path)
-    #[cfg(windows)]
-    {
-        let _ = bin_dir;
-    }
-
-    #[cfg(not(windows))]
-    {
-        // Derive vite_plus_home display path from bin_dir.parent(), using $HOME prefix
-        let home_path = bin_dir
-            .parent()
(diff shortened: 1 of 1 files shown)


fengmk2 (approved):
Thanks!

PR #974, opened 2026-03-16
feat(cli): add shell completion support

### LINKED ISSUE #950: Shell completion support
### Description

<kbd>TAB</kbd> completion should improve developer experience.

### Examples

- `vp` <kbd>TAB</kbd> for `vp` commands, params and options
- `vp run` <kbd>TAB</kbd> for scripts in the `package.json` file.

### Suggested solution

Since `~/.vite-plus/env` is already present, it can be added directly to the file. In this case, only `vp upgrade` is needed.

### Validations

nekomoyi:
## Summary

Implements shell completion for the vp CLI.

## Changes

Completion scripts for bash, zsh, fish, and PowerShell are generated in `~/.vite-plus/completion/` when running `vp env setup`, and automatically sourced in the respective env files.

## Fixes
Fixed mixed path separators in env files on Windows to use forward slashes consistently for `$HOME`-relative paths
```bash
# Before
__vp_bin="$HOME/.vite-plus\bin"
# After
__vp_bin="$HOME/.vite-plus/bin"
```
Increased stack size to 8MB to handle deep recursion in `clap_complete::generate()` which exceeds Windows' default stack size.

Closes #950

### DIFF
--- a/.claude/skills/bump-vite-task/SKILL.md
+++ b/.claude/skills/bump-vite-task/SKILL.md
@@ -1,0 +1,86 @@
+---
+name: bump-vite-task
+description: Bump vite-task git dependency to the latest main commit. Use when you need to update the vite-task crates (fspy, vite_glob, vite_path, vite_str, vite_task, vite_workspace) in vite-plus.
+allowed-tools: Read, Grep, Glob, Edit, Bash, Agent, WebFetch
+---
+
+# Bump vite-task to Latest Main
+
+Update the vite-task git dependency in `Cargo.toml` to the latest commit on the vite-task main branch, fix any breaking changes, and create a PR.
+
+## Steps
+
+### 1. Get current and target commits
+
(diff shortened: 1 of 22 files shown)


.cargo/config.toml
```diff
@@ -4,10 +4,17 @@ WORKSPACE_DIR = { value = "rolldown", relative = true }
 
 [build]
 rustflags = ["--cfg", "tokio_unstable"] # also update .github/workflows/ci.yml
+
 # fix sqlite build error on linux
 [target.'cfg(target_os = "linux")']
 rustflags = ["--cfg", "tokio_unstable", "-C", "link-args=-Wl,--warn-unresolved-symbols"]
 
+# Increase stack size on Windows to avoid stack overflow
```
fengmk2, line 12: @nekomoyi Is it caused by the newly introduced clap_complete?

nekomoyi: Yes, this is the root cause.However, I believe this is an underlying issue of `clap` on Windows, and the recursive generate process of `clap_complete` exposed it.
clap-rs/clap#5134

wtto00:
There is a bug in the `__vp_bin` path configuration. Please see #1039.

PR #986, opened 2026-03-17
fix(cli): adopt DA1 sandwich technique for robust OSC color query detection

Brooooooklyn:
Replace ad-hoc environment checks (CI, Warp, tmux, Docker) with a
comprehensive `is_osc_query_unsupported()` pre-screen modelled after
`terminal-colorsaurus`, and implement the DA1 sandwich technique to
detect unsupported terminals at runtime without waiting for timeouts.

Key changes:
- Centralized quirks detection covering CI, Docker, devcontainers,
  Kubernetes, Emacs, GNU Screen, Eterm, tmux, and TERM=dumb
- DA1 sentinel query appended after OSC queries; if its response
  arrives first the terminal doesn't support OSC and we bail immediately
- drain_da1() consumes the trailing DA1 response to prevent leaks
- BEL terminator instead of ST to work around urxvt response bug
- SSH-aware timeout (1000ms vs 200ms local)
- Ported regression test from terminal-colorsaurus issue #38
- Extended test coverage for parsing edge cases

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

---

> [!NOTE]
> **Medium Risk**
> Changes low-level `/dev/tty` I/O and escape-sequence parsing to avoid hangs/leaked output across many terminal environments; mistakes could regress header color detection or terminal interaction timing.
> 
> **Overview**
> Makes Vite+ header color probing **more robust across terminal emulators and container/CI environments** by replacing ad-hoc OSC query handling with a centralized `is_osc_query_unsupported()` gate plus the *DA1 sandwich* technique to detect lack of OSC support without waiting for timeouts.
> 
> Reworks `query_terminal_colors` to send **BEL-terminated** OSC queries (urxvt workaround), parse responses **sequentially from the tty stream** via `BufReader` (with a poll-based, deadline-aware reader), and drain trailing DA1 output; also adds SSH-aware timeouts.
> 
> Expands unix test coverage for hex/RGB parsing, BEL termination, mixed-response buffers, `read_until_either`, and includes a regression test ensuring `query_terminal_colors` doesn’t hang in CI.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a2f33995a33cc4bf08dcce61202c5f78588ba8c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #988, opened 2026-03-17
feat(cli): set terminal title to project name

### LINKED ISSUE #978: Set terminal title
### Description

Right now, the terminal window title is just "vp"

Image

### Suggested solution

Set a title based on the projects name (inferred via pkg.json).

### Alternative

_No response_

### Additional context

_No response_

### Validations

zerone0x:
## Summary

- Read the project name from the nearest `package.json` and set it as the terminal window title using the OSC 0 escape sequence (`ESC ] 0 ; <title> BEL`)
- Applies to both the global CLI (`vp` binary) and the local CLI entry point
- Falls back gracefully: no-op when stdout is not a terminal, when running in CI, or when no `package.json` with a `name` field is found

Fixes #978

## Test plan

- [x] `vite_shared` tests pass (31 passed, 1 ignored)
- [ ] Run `vp dev` in a project directory and verify the terminal tab/window title shows the project name instead of "vp"
- [ ] Run `vp` outside any project directory and verify no error occurs (graceful fallback)
- [ ] Verify CI pipeline passes

### DIFF
--- a/crates/vite_shared/src/header.rs
+++ b/crates/vite_shared/src/header.rs
@@ -525,6 +525,21 @@ fn render_header_variant(
     format!("{}{}", bold(&vite_plus, prefix_bold), bold(&suffix, suffix_bold))
 }
 
+/// Set the terminal window title using OSC 0 escape sequence.
+///
+/// Writes `ESC ] 0 ; <title> BEL` to stdout when stdout is a terminal.
+/// This is a no-op when stdout is not a terminal or when running in CI.
+pub fn set_terminal_title(title: &str) {
+    use std::io::Write;
+
+    if !std::io::stdout().is_terminal() || std::env::var_os("CI").is_some() {
+        return;
+    }
+
+    let _ = write!(std::io::stdout(), "\x1b]0;{title}\x07");
+    let _ = std::io::stdout().flush();
+}
+
 /// Render the Vite+ CLI header string with JS-parity coloring behavior.
 #[must_use]
 pub fn vite_plus_header() -> String {
--- a/crates/vite_shared/src/header.rs
+++ b/crates/vite_shared/src/header.rs
@@ -527,17 +527,36 @@ fn render_header_variant(
 
 /// Set the terminal window title using OSC 0 escape sequence.
 ///
-/// Writes `ESC ] 0 ; <title> BEL` to stdout when stdout is a terminal.
-/// This is a no-op when stdout is not a terminal or when running in CI.
+/// Writes `ESC ] 0 ; <title> BEL` only when stdout looks like a terminal with
+/// ANSI/VT escape support. This is a best-effort hint: unsupported terminals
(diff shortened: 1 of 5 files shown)


crates/vite_shared/src/header.rs
```diff
@@ -363,6 +363,21 @@ fn render_header_variant(
...
 }
 
+/// Set the terminal window title using OSC 0 escape sequence.
+///
+/// Writes `ESC ] 0 ; <title> BEL` to stdout when stdout is a terminal.
+/// This is a no-op when stdout is not a terminal or when running in CI.
+pub fn set_terminal_title(title: &str) {
+    use std::io::Write;
+
+    if !std::io::stdout().is_terminal() || std::env::var_os("CI").is_some() {
```
fengmk2: Will there be similar issues to https://github.com/voidzero-dev/vite-plus/pull/986? Under what conditions would title not be supported?

fengmk2:
@codex review

zerone0x (commented):
PR #986 dealt with OSC color queries (OSC 10/11), which many terminals don't respond to and can cause hangs. OSC 0 (set window title) is a write-only escape — it doesn't wait for a terminal response, so there's no risk of blocking or timeout. It's supported by virtually all modern terminal emulators (xterm, iTerm2, Terminal.app, Windows Terminal, GNOME Terminal, etc.).

The only case where it's "not supported" is if a terminal simply ignores the sequence, which is harmless. The current guards (`is_terminal()` + `CI` check) should be sufficient.

crates/vite_shared/src/header.rs
zerone0x: Good call. #986 is different because it is about OSC *queries* (OSC 10/11/4): those write to the tty and then wait for a response, so unsupported/non-forwarding environments can hang or leak responses.

OSC 0 title setting is write-only, so it should not have the same blocking class of issue. The possible failure modes are:
- stdout is not a TTY / CI: no terminal to update
- `TERM=dumb`, missing/disabled ANSI/VT support, or older Windows console hosts without VT processing: escape bytes could be printed literally
- terminal multiplexers/remotes may ignore or rewrite the title sequence
- individual terminal emulators may simply ignore OSC 0

I pushed 8ff69643 to make the fallback more conservative: it now only emits OSC 0 when stdout is a TTY, not CI/GitHub Actions, and the existing `supports-color` terminal capability check says ANSI/VT escapes are supported. I also sanitize control characters from the package name before writing the title. If unsupported, this is a no-op.

fengmk2:
@zerone0x Can you resolve the conflicts and continue completing this PR?

zerone0x:
@fengmk2 Rebased onto main and resolved the conflict. The cli.rs → cli/* module split removed the original injection point; ported the 5-line `set_terminal_title` call to `cli/mod.rs` (equivalent location, same behavior). 5 files +81. PTAL.

PR #1000, opened 2026-03-17
chore(cli): upgrade @oxc-node/core to catalog version (0.0.35)

Arthie:
The `@oxc-node/core` dependency in the cli package stood out as being the only package without the version set to `catalog:`.
Upgrading the version to `catalog:` avoids dependency version drift between workspace packages.


fengmk2:
@Arthie Sorry for handling your PR so late. I will fix the conflicts and then merge it. Thank you for contributing to open source!

fengmk2 (approved):

PR #1001, opened 2026-03-17
chore: replace husky with vp git hooks

Arthie:
Replace husky and lint-staged with the built-in Vite+ hooks, so this repo dogfoods `vp config` and `vp staged`.

- `.vite-hooks/pre-commit` runs `vp staged`
- The lint-staged rules move to the `staged` block in `vite.config.ts`
- `prepare` runs `vp config` instead of `husky`
- Remove husky and lint-staged from the root package.json and husky from the catalog (`packages/cli` keeps lint-staged, it implements `vp staged`)

Hooks stay opt-in: the repo sets `ignoreScripts: true`, so `prepare` does not run on `pnpm install`. To activate, run `vp config` after `pnpm build`. If a clone activated husky before, run `git config --unset core.hooksPath` once first.

Earlier revisions also committed `vp config` agent files (`.claude/settings.json`, a skills symlink, a CLAUDE.md block). The rebase dropped them: `vp mcp` and the shipped skills directory no longer exist, and repo guidance now lives in AGENTS.md.


pnpm-workspace.yaml
```diff
@@ -68,7 +68,6 @@ catalog:
   follow-redirects: ^1.15.6
   fs-extra: ^11.3.2
   glob: ^13.0.0
-  husky: ^9.1.7
   jsonc-parser: ^3.3.1
   lint-staged: ^16.2.6
```
Arthie, line 92: `"lint-staged": "catalog:"` is only used in `packages/cli/package.json` let me know if I should remove it here and update it in the cli package.

fengmk2: it can be removed now.

fengmk2 (commented):
@Arthie Thanks, I will merge after doing local verification with no issues.

pnpm-workspace.yaml
Arthie: ok removed from `pnpm-workspace.yaml` and added to the `packages/cli/package.json`

johnrommeloctaviano (approved):

lyzno1:
Two things I think still need to be adjusted here:

1. `lint-staged` should stay in the catalog. The root repo can drop its `lint-staged` config/dependency, but `packages/cli` still imports `lint-staged` to implement `vp staged`, so `packages/cli/package.json` should keep `"lint-staged": "catalog:"`.

2. This also needs to be wired into contributor setup. Since `pnpm-workspace.yaml` has `ignoreScripts: true`, `prepare: vp config` will not run during `pnpm install`. `just init` should explicitly run `vp config` after install so contributors actually get `core.hooksPath` configured.

fengmk2:
@Arthie I will take over and fix conflicts before merging.

fengmk2 (approved):

PR #1027, opened 2026-03-18
refactor(config): auto-update agent instructions, never create new files

fengmk2:
## Summary

- `vp config` no longer creates new agent files or prompts for agent selection
- Silently updates existing files that contain Vite+ markers (`<!--VITE PLUS START-->`) when content is outdated
- Same behavior for `prepare` lifecycle, interactive, and non-interactive modes
- Skipped with `--hooks-only`
- Delete `config/agent.ts` and its tests (superseded by shared `utils/agent.ts`)

## Behavior

| Scenario | Action |
|----------|--------|
| No agent files | No writes |
| Agent files exist, no Vite+ markers | No writes |
| Vite+ markers, content up to date | No writes |
| Vite+ markers, content outdated | Update marked section silently |
| `--hooks-only` | Skip agent update |

## Test plan

- [x] `pnpm -F vite-plus snap-test-global command-config` — all 9 tests pass
- [x] `pnpm -F vite-plus snap-test` — all tests pass
- [x] New snap test: `command-config-update-agents` — verifies outdated markers are updated, user content preserved
- [x] New snap test: `command-config-no-agent-writes` — verifies no writes when no markers, no new files created


packages/cli/src/utils/agent.ts
```diff
@@ -280,6 +280,40 @@ export function hasExistingAgentInstructions(projectRoot: string): boolean {
...
+ * - Markers present, content outdated → update marked section
+ */
+export function updateExistingAgentInstructions(projectRoot: string): void {
+  const targetPaths = detectExistingAgentTargetPaths(projectRoot);
+  if (!targetPaths) {
+    return;
+  }
+
+  const templatePath = path.join(pkgRoot, 'AGENTS.md');
+  if (!fs.existsSync(templatePath)) {
```
fengmk2, line 297: oh no, only agents.md

fengmk2: If there's only [CLAUDE.md](http://CLAUDE.md), won't it stop auto-updating? Is this problematic?

cpojer: We should support all the files we support already. Keep in mind that we need to also resolve symlinks since these often tend to be symlinked. `vp migrate` handles that already.

(We might want to also just use symlinks for these in the first place and only create `AGENTS.md`)

fengmk2: I reviewed it wrong, the AI's logic is correct, you can ignore me. 😭

PR #1033, opened 2026-03-18
feat(cli): add Kiro agent instructions

ErikCH:
Add Kiro to the AGENTS array so it shows up in the interactive agent selection during `vp create` and `vp migrate`. Kiro uses `AGENTS.md` at the workspace root, which it natively supports.


fengmk2 (approved):
@ErikCH Thanks!

Issue #1014, opened 2026-03-18
reqwest calls in vp fail when running behind a MITM proxy.


watiko:
### Describe the bug

When I ran the installation script in a WSL2 environment where Zscaler certificates were installed, an error occurred.
After checking the error details, it appeared that running `vp install` failed to retrieve the Node.js runtime list from [https://nodejs.org/dist/index.json](https://nodejs.org/dist/index.json?utm_source=copilot.com).
Since TLS communication worked correctly with curl and other programs, I suspected that the certificates were not being referenced properly. While reviewing the repository, I found the following:

1. The reqwest version is 0.12
https://github.com/voidzero-dev/vite-plus/blob/9db93c21726781c7fe0d2d90b7373ad172af2dbc/Cargo.toml#L140 
2. The enabled reqwest features are ["stream", "rustls-tls", "json"]
https://github.com/voidzero-dev/vite-plus/blob/9db93c21726781c7fe0d2d90b7373ad172af2dbc/crates/vite_install/Cargo.toml#L39-L40
3. In reqwest 0.12, specifying rustls-tls does not use the system’s CA certificates
https://github.com/seanmonstar/reqwest/issues/2640
4. In reqwest 0.13, the handling of root certificates has been reorganized
https://seanmonstar.com/blog/reqwest-v013-rustls-default/

As for possible fixes, I believe the options are either adjusting the feature flags while staying on reqwest 0.12, or upgrading to reqwest 0.13.

### Reproduction

To reproduce this, you need to introduce a MITM proxy such as Zscaler on the network side, rather than relying on the contents of a repository.

### Steps to reproduce

Run `vp install` under any MITM proxy.

### System Info

```shell
❯ ~/watiko/.vite-plus/0.1.12/bin/vp env current
VITE+ - The Unified Toolchain for the Web

error: Failed to download Node.js runtime: Failed to download from https://nodejs.org/dist/index.json: error sending request for url (https://nodejs.org/dist/index.json)

❯ ~/watiko/.vite-plus/0.1.12/bin/vp --version
VITE+ - The Unified Toolchain for the Web

vp v0.1.12

Local vite-plus:
  vite-plus  Not found

Tools:
  vite             Not found
  rolldown         Not found
  vitest           Not found
  oxfmt            Not found
  oxlint           Not found
  oxlint-tsgolint  Not found
  tsdown           Not found

Environment:
  Package manager  Not found
  Node.js          Not found



[...]

PR #1068, opened 2026-03-20
fix: upgrade reqwest from 0.12 to 0.13 and rename the rustls-tls feature to rustls

Giorno-Giovana:
closes https://github.com/voidzero-dev/vite-plus/issues/1014

Motivation

vp install fails on macOS with a TLS error when trying to download the Node.js runtime from nodejs.org:

error: Failed to download Node.js runtime: Failed to download from
https://nodejs.org/dist/v22.18.0/SHASUMS256.txt: error sending request for url

The root cause is how reqwest 0.12 handles TLS certificate verification on non-Windows platforms. With the
rustls-tls feature, it uses webpki-roots — a hardcoded bundle of Mozilla root certificates compiled into the
binary. If the server's certificate chain involves a CA not present in that bundle (or the bundle is
outdated), TLS verification fails, even though the system's certificate store (macOS Keychain) trusts the
certificate.

reqwest 0.13 replaces the rustls-tls feature with rustls, which uses rustls-platform-verifier instead of
webpki-roots. This verifier delegates certificate validation to the OS-native certificate store
(Security.framework on macOS, SChannel on Windows), matching the behavior of curl and other system tools.

### DIFF
--- a/packages/cli/src/create/__tests__/builtin.spec.ts
+++ b/packages/cli/src/create/__tests__/builtin.spec.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { executeBuiltinTemplate } from '../templates/builtin.js';
+
+const { mockLogError } = vi.hoisted(() => ({ mockLogError: vi.fn() }));
+
+vi.mock('../templates/remote.js', () => ({
+  runRemoteTemplateCommand: vi.fn(),
+}));
+
+vi.mock('@voidzero-dev/vite-plus-prompts', () => ({
+  log: { error: mockLogError },
+}));
+
+const workspaceInfo = {
+  rootDir: '/tmp/workspace',
+} as any;
+
+const baseTemplateInfo = {
+  packageName: 'wage-meeting',
+  targetDir: 'wage-meeting',
+  args: [],
+  envs: {},
+  type: 'builtin' as any,
+  interactive: false,
+};
+
+describe('executeBuiltinTemplate', () => {
+  it('returns exitCode 1 for unknown vite: template', async () => {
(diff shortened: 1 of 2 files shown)


crates/vite_install/src/request.rs
```diff
@@ -57,6 +65,8 @@ impl HttpClient {
     }
 
     async fn get(&self, url: &str) -> Result<Response, Error> {
+        ensure_tls_provider();
```
fengmk2: windows fails:

```
error[E0425]: cannot find function `ensure_tls_provider` in this scope
  --> crates\vite_install\src\request.rs:68:9
   |
68 |         ensure_tls_provider();
   |         ^^^^^^^^^^^^^^^^^^^ not found in this scope

    Checking rolldown_plugin_vite_build_import_analysis v0.1.0 (D:\a\vite-plus\vite-plus\rolldown\crates\rolldown_plugin_vite_build_import_analysis)
For more information about this error, try `rustc --explain E0425`.
    Checking rolldown_plugin_vite_resolve v0.1.0 (D:\a\vite-plus\vite-plus\rolldown\crates\rolldown_plugin_vite_resolve)
error: could not compile `vite_install` (lib) due to 1 previous error
```

crates/vite_js_runtime/src/download.rs
```diff
@@ -15,6 +15,14 @@ use vite_str::Str;
 
 use crate::{Error, provider::ArchiveFormat};
 
+#[cfg(not(target_os = "windows"))]
+fn ensure_tls_provider() {
```
fengmk2: can we move this function to `vite_shared` crate? it should be reused by others.

Giorno-Giovana: Done

PR #1097, opened 2026-03-22
feat(create): support creating project in current directory

### LINKED ISSUE #1077: Feature request: Add way to create project in cwd
### Description

I tried to create project using `vp create`, but it does not seem like there is "create project in this directory" option.
As I use [ghq](https://github.com/x-motemen/ghq) for managing my projects, I create a directory that is empty but git repository, then do `pnpm init`,`cargo init`, `bundle init`, or whatever.
So I tried to do so in vp, but it looks like there is no "Create here" option:
```
❯ : npx --package vite-plus vp create --directory .


Parsed package name "." is invalid: name cannot start with a period
The --directory option is invalid
```
```
❯ : npx --package vite-plus vp create

  Vite+ Application

▲ Package name:
  .█
 name cannot start with a period

```

### Suggested solution

Support `--directory .`, which creates project in current directory. It would be nice if it asks project name, as someone may make a directory that is not "project name"-friendly. (I sometimes encounter "Your directory's name, foo.bar, cannot be project name. Please specify project name via `--lib`" in Cargo)

### Alternative

Add `vp init`, which creates project in current directory.

### Additional context

Having both solution would be nice I guess.

### Validations

jong-kyung:
## Summary
- Support `vp create --directory .` to scaffold a project in the current directory without creating a new folder
- Auto-derive package name from the current directory name (no interactive prompt), similar to `cargo init` / `npm init -y`
- Fix `"./"` input to be treated equivalently to `"."` in `formatTargetDir`
- Extract `deriveDefaultPackageName` to `utils.ts` for testability
- Log the auto-derived package name when using `--directory .`

Closes #1077

### DIFF
--- a/packages/cli/snap-tests-global/create-from-monorepo-subdir/snap.txt
+++ b/packages/cli/snap-tests-global/create-from-monorepo-subdir/snap.txt
@@ -23,3 +23,15 @@ Created at apps/custom-app with --directory
 > mkdir -p apps/dot-test && cd apps/dot-test && vp create --no-interactive vite:application --directory . # --directory . from monorepo subdir
 > test -f apps/dot-test/package.json && echo 'Created at apps/dot-test with --directory .' || echo 'NOT at apps/dot-test'
 Created at apps/dot-test with --directory .
+
+> vp create --no-interactive vite:application --directory . 2>&1 || true # --directory . from monorepo root should fail
+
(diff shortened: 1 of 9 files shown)


fengmk2 (approved):
@jong-kyung LGTM, please take a look at CI fails.

PR #1102, opened 2026-03-22
refactor(dev): simplify local dev workflow

Salnika:
## Summary

This PR simplifies local development of the CLI.

It adds a repo-local setup and bootstrap flow so a fresh clone can be prepared with:

```bash
pnpm install:dev
pnpm bootstrap:dev
```

It also keeps `bootstrap-cli` for the global install flow.

## What changed

- added `pnpm install:dev` to prepare local upstream checkouts and install dependencies
- added `pnpm bootstrap:dev` to prepare the repo and build the local CLI
- updated local CLI and test scripts to use repo-local artifacts
- updated global snap test execution to use the repo-local binary instead of relying on `~/.vite-plus`
- updated `just init` and contributor docs to reflect the new workflow

## Why

The local development workflow required extra manual setup before the CLI could be built and tested.

This PR makes the common development path more direct while keeping the existing global install flow available.


fengmk2:
@Salnika After I changed the snap test to sharding, there should be some incompatible changes. Can you continue to fix it?

alexis-salnikoff-wmx:
@fengmk2 I finaly got time to clean this pr :)

crates/vite_global_cli/src/commands/global/install.rs
```diff
@@ -364,8 +364,14 @@ async fn install_one(
 
     // 4. Run npm install with prefix set to staging directory
     //    Pipe stdout/stderr so npm output is hidden on success, shown on failure
+    let mut install_args = vec!["install", "-g", "--no-fund"];
+    if is_local_package_spec(package_spec) {
```
fengmk2, line 519: Are these changes unrelated to the PR content?

Salnika: No, --install-links is required for local installs to work correctly

But I had to do some other changes to fix the test/CI and could use a little help if something isn't right 😅

PR #1103, opened 2026-03-22
fix: support yarn pnp cli delegation

Salnika:
## Summary

This PR adds Yarn PnP support to local CLI delegation.

It makes delegated JS commands run through a Yarn-aware local execution path for Yarn PnP projects, adds a delegate bootstrap that resolves the local `vite-plus` CLI from the project context, and keeps Vite config loading compatible with the current parent-directory config traversal.

It also adds regression coverage for the delegate bootstrap and for the Yarn PnP monorepo create flow.


fengmk2:
only support nodeLinker mode.

PR #1109, opened 2026-03-23
fix(editor): correct Vue.js language key in Zed settings

CoutinhoTTS:
image


packages/cli/src/utils/editor.ts
```diff
@@ -122,7 +122,7 @@ const ZED_SETTINGS = {
       prettier: { allowed: false },
       formatter: [{ language_server: { name: 'oxfmt' } }],
     },
-    Vue: {
+    'Vue.js': {
```
fengmk2, line 125: Can you provide a reference source for me to see why the field is named `Vue.js`?

CoutinhoTTS: > Can you provide a reference source for me to see why the field is named `Vue.js`?

here👉🏻：https://github.com/zed-extensions/vue/blob/main/extension.toml

PR #1116, opened 2026-03-23
docs: add manual package-manager install path for vp

yash27-lab:
## Summary

Add a manual package-manager-based installation path for `vp` to the docs.

## Changes

- add `npm install -g vite-plus` and `pnpm add -g vite-plus` to `/guide/`
- add follow-up notes for `PATH`, `pnpm setup`, `vp env setup`, and `vp env off`
- add matching npm/pnpm upgrade commands to `/guide/upgrade`

## Verification

- verified the docs gap still exists on current `main`
- tested `npm install -g vite-plus` in an isolated temp prefix
- tested `pnpm add -g vite-plus` in an isolated temp global environment
- tested upgrade with:
  - `npm install -g vite-plus@latest`
  - `pnpm add -g vite-plus@latest`
- ran `cd docs && pnpm install && pnpm build`

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -24,6 +24,27 @@ After installation, open a new shell and run:
 vp help
 ```
 
+### Install with npm or pnpm
+
+If you prefer not to pipe a remote script into your shell, you can install the published `vite-plus` package globally with a package manager instead:
+
+```bash
+npm install -g vite-plus
+# or
+pnpm add -g vite-plus
+```
+
+After installing, open a new shell and run `vp help`.
+
+If your shell cannot find `vp`, make sure your package manager's global bin directory is on your `PATH`. For `pnpm`, you may need to run `pnpm setup` first.
+
+This installation path is more manual than the install script:
+
+- It does not modify your shell profile for you.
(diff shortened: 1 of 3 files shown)


docs/guide/index.md
```diff
@@ -24,6 +24,27 @@ After installation, open a new shell and run:
 vp help
 ```
 
+### Install with npm or pnpm
+
+If you prefer not to pipe a remote script into your shell, you can install the published `vite-plus` package globally with a package manager instead:
```
fengmk2, line 29: the `vp` cli from `vite-plus`'s package we call it `Local CLI`. It is not recommended to install this `Local CLI` globally through a package manager, as its features differ from those of the `Global CLI`, such as the lack of node runtime manager(e.g.: `vp env` command is missing) and package manager functionalities.

yash27-lab: Thanks, that makes sense.

I misunderstood the published `vite-plus` package as a supported global installation path for the full `vp` CLI. I’ll avoid documenting it that way.

Would you prefer this PR to be reshaped to clarify that:
- the install script is the supported way to install the global `vp` CLI
- the published `vite-plus` package is the project-local CLI and should not be installed globally

PR #1124, opened 2026-03-24
fix(migrate): warn on uncoercible husky version instead of false positive

### LINKED ISSUE #1114: vp migrate false-positive husky version check when version is a dist-tag or non-semver string ("latest", "next", etc.)
### Describe the bug

vp migrate reports "Detected husky <9.0.0 — please upgrade to husky v9+ first, then re-run migration." even when husky v9 is installed. The check reads the version string from package.json devDependencies and uses semver.coerce() on it. When the version is "latest", semver.coerce("latest") returns null, which falls back to "0.0.0" and incorrectly satisfies < 9.0.0.

### Reproduction

https://github.com/mildlyfunctional/vp-migrate-repro

### Steps to reproduce

Steps to reproduce:
1. Clone the reproduction repo
2. Run vp migrate
3. See: Detected husky <9.0.0 — please upgrade to husky v9+ first, then re-run migration.

### System Info

```shell
Environment:
  Version  24.14.0
  Source   lts

Tool Paths:
  node  /Users/mrwokkel/.vite-plus/js_runtime/node/24.14.0/bin/node
  npm   /Users/mrwokkel/.vite-plus/js_runtime/node/24.14.0/bin/npm
  npx   /Users/mrwokkel/.vite-plus/js_runtime/node/24.14.0/bin/npx


vp v0.1.13

Local vite-plus:
  vite-plus  Not found

Tools:
  vite             Not found
  rolldown         Not found
  vitest           Not found
  oxfmt            Not found
  oxlint           Not found
  oxlint-tsgolint  Not found
  tsdown           Not found

Environment:
  Node.js          v24.14.0
```

### Used Package Manager

npm

### Logs

```shell

```

### Validations

jong-kyung:
## Summary

- Fix false positive "husky <9.0.0" warning in `vp migrate` when husky version is a dist-tag (`"latest"`, `"next"`) or other non-semver string
- When `semver.coerce()` returns `null`, warn the user to specify a semver-compatible version instead of silently falling back to `'0.0.0'`
- Change `checkUnsupportedHuskyVersion` return type from `boolean` to `string | null` to co-locate message logic with version checking

Closes #1114

### DIFF
--- a/packages/cli/src/migration/migrator.ts
+++ b/packages/cli/src/migration/migrator.ts
@@ -1510,18 +1510,25 @@ function rewriteAllImports(projectPath: string, silent = false, report?: Migrati
 /**
  * Check if the project has an unsupported husky version (<9.0.0).
  * Uses `semver.coerce` to handle ranges like `^8.0.0` → `8.0.0`.
- * Accepts pre-loaded deps to avoid re-reading package.json when called
- * from contexts that already parsed it.
+ * Returns a reason string if hooks migration should be skipped, or null
+ * if husky is absent or compatible.
  */
 function checkUnsupportedHuskyVersion(
   deps: Record<string, string> | undefined,
   prodDeps: Record<string, string> | undefined,
-): boolean {
+): string | null {
   const huskyVersion = deps?.husky ?? prodDeps?.husky;
   if (!huskyVersion) {
-    return false;
+    return null;
   }
-  return semver.satisfies(semver.coerce(huskyVersion) ?? '0.0.0', '<9.0.0');
+  const coerced = semver.coerce(huskyVersion);
+  if (coerced == null) {
+    return `Could not determine husky version from "${huskyVersion}" — please specify a semver-compatible version (e.g., "^9.0.0") and re-run migration.`;
+  }
+  if (semver.satisfies(coerced, '<9.0.0')) {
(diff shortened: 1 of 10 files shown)


packages/cli/src/migration/migrator.ts
```diff
@@ -1510,18 +1510,25 @@ function rewriteAllImports(projectPath: string, silent = false, report?: Migrati
...
-): boolean {
+): string | null {
   const huskyVersion = deps?.husky ?? prodDeps?.husky;
   if (!huskyVersion) {
-    return false;
+    return null;
   }
-  return semver.satisfies(semver.coerce(huskyVersion) ?? '0.0.0', '<9.0.0');
+  const coerced = semver.coerce(huskyVersion);
+  if (coerced == null) {
```
fengmk2, line 1528: Under this condition, should we try to read the version of husky installed in node_modules? It's very likely that the user has already executed npm install in the current project, and in this case, we can read the actual version.

jong-kyung: > Under this condition, should we try to read the version of husky installed in node_modules? It's very likely that the user has already executed npm install in the current project, and in this case, we can read the actual version.

Good point! Will use the `detectPackageMetadata` function from `package.ts` to read the installed version from node_modules as a fallback when semver.coerce fails, keeping the current warning only when node_modules isn't available

fengmk2 (approved):

PR #1127, opened 2026-03-24
fix(create): suggest `vp run` for all templates

### LINKED ISSUE #1123: Running `vp dev` on a new Astro project doesn't work
### Describe the bug

After creating an astro project with `vp create astro` the cli recommends to run

`Next: cd astro-project && vp dev`

But running `vp dev` starts the development server at 5173, which is not what astro does.

Running `vp run dev` works correctly.

It might be better to rather recommend running `vp run dev` instead of `vp dev` for astro projects.

This might be related to #912 

### Reproduction

https://stackblitz.com/~/github.com/joknoll/viteplus-astro

### Steps to reproduce

_No response_

### System Info

```shell

vp v0.1.13

Local vite-plus:
  vite-plus  v0.1.13

Tools:
  vite             v8.0.1
  rolldown         v1.0.0-rc.10
  vitest           v4.1.0
  oxfmt            v0.41.0
  oxlint           v1.56.0
  oxlint-tsgolint  v0.17.1
  tsdown           v0.21.4

Environment:
  Node.js          v24.14.0 (engines.node)
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

jong-kyung:
## Summary
- Fix `vp create` to suggest `vp run` instead of `vp dev` for non-Vite-native templates (Astro, Nuxt, etc.)
- Also fix the monorepo path: non-builtin templates now suggest `cd <dir> && vp run` instead of `vp dev <dir>`
- Extract `getCreateNextCommand` and `getNextCommand` to `utils.ts` for testability
- Add unit tests covering all template type × environment (standalone/monorepo) combinations
- Add snap tests for monorepo next command verification

Closes #1123


packages/cli/src/create/utils.ts
```diff
@@ -139,3 +140,31 @@ export function deriveDefaultPackageName(
...
+ */
+export function getCreateNextCommand(
+  projectDir: string,
+  selectedTemplateName: string,
+  isMonorepo: boolean,
+): string {
+  const isBuiltinTemplate = selectedTemplateName.startsWith('vite:');
+  const hasBuiltinDevServer = isBuiltinTemplate && selectedTemplateName !== BuiltinTemplate.library;
+
+  if (isMonorepo) {
```
fengmk2: I suddenly thought that it could be simpler and more direct to just prompt `vp run` without `dev`, as it will prompt the user to choose which command to run. 
This way, it can adapt to all scenarios, whether or not the user has configured scripts.dev, and it doesn't need to distinguish between monorepos.

jong-kyung: > I suddenly thought that it could be simpler and more direct to just prompt `vp run` without `dev`, as it will prompt the user to choose which command to run. This way, it can adapt to all scenarios, whether or not the user has configured scripts.dev, and it doesn't need to distinguish between monorepos.

Thanks for the suggestion! That makes sense — using just `vp run` is much simpler and covers all cases without needing to distinguish template types or monorepo setups. I'll make the changes and push an update shortly.

fengmk2 (approved):

Issue #1158, opened 2026-03-26
`vp migrate` handles migration from `.nvmrc`


TheAlexLichter:
### Description

As a dev using Vite+, I want my old nvm files being migrated over to the Vite+ way.

### Suggested solution

Ask them if they want to migrate them over when being used.

### Alternative

Also adhere to .nvmrc (meh)

### Additional context

Same for other node version managers!

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

PR #1159, opened 2026-03-26
feat(migrate): migrate .nvmrc to .node-version during `vp migrate`

naokihaba:
### 🔗 Related Issues

Resolves: #1158

### 📚 Description

Vite+ previously retrieved the Node.js version from .node-version, but projects using nvm typically use .nvmrc.

With this change, automatic detection of .nvmrc and a migration feature to .node-version have been added as part of the vp migrate command execution process.

During migration, if a v prefix is present at the beginning, it is removed, while aliases such as lts/* are carried over as is.

Unit tests have confirmed that it works correctly regardless of the presence of the v prefix.

On the other hand, if unsupported aliases like node or stable are included, the system is designed to display a warning and skip the migration rather than failing.

### Manual Test: .nvmrc Migration

```bash
$ mkdir /tmp/vp-migrate-test
$ echo '{ "name": "test-migrate", "version": "1.0.0" }' > package.json
$ echo 'v20.5.0' > .nvmrc
```

#### Command

```bash
$ cd /tmp/vp-migrate-test && vp migrate --no-interactive
```

#### Output (Excerpt)

```
.nvmrc detected. Auto-migrating to .node-version...
...
• Node version manager file migrated to .node-version
```

#### Final State

```
$ cat /tmp/vp-migrate-test/.node-version
20.5.0
```

- .nvmrc (v20.5.0) -> Deleted
- .node-version -> 20.5.0 (With the v prefix removed)

### DIFF
--- a/packages/cli/src/migration/migrator.ts
+++ b/packages/cli/src/migration/migrator.ts
@@ -726,6 +726,7 @@ export function rewriteStandaloneProject(
   }
   mergeViteConfigFiles(projectPath, silent, report);
   injectLintTypeCheckDefaults(projectPath, silent, report);
+  injectFmtDefaults(projectPath, silent, report);
   mergeTsdownConfigFile(projectPath, silent, report);
   // rewrite imports in all TypeScript/JavaScript files
   rewriteAllImports(projectPath, silent, report);
@@ -771,6 +772,7 @@ export function rewriteMonorepo(
   }
   mergeViteConfigFiles(workspaceInfo.rootDir, silent, report);
   injectLintTypeCheckDefaults(workspaceInfo.rootDir, silent, report);
+  injectFmtDefaults(workspaceInfo.rootDir, silent, report);
   mergeTsdownConfigFile(workspaceInfo.rootDir, silent, report);
   // rewrite imports in all TypeScript/JavaScript files
   rewriteAllImports(workspaceInfo.rootDir, silent, report);
@@ -1335,6 +1337,21 @@ export function injectLintTypeCheckDefaults(
     'lint',
     '.vite-plus-lint-init.oxlintrc.json',
     JSON.stringify({ options: { typeAware: true, typeCheck: true } }),
+    silent,
+    report,
+  );
+}
+
+export function injectFmtDefaults(
+  projectPath: string,
+  silent = false,
+  report?: MigrationReport,
+): void {
+  injectConfigDefaults(
+    projectPath,
+    'fmt',
+    '.vite-plus-fmt-init.oxfmtrc.json',
+    JSON.stringify({}),
     silent,
     report,
   );
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7193,6 +7193,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "clap",
+ "cow-utils",
  "fspy",
  "napi",
  "napi-build",
--- a/crates/vite_migration/src/vite_config.rs
+++ b/crates/vite_migration/src/vite_config.rs
@@ -1013,6 +1013,27 @@ export default defineConfig({
     }
 
     #[test]
+    fn test_merge_json_config_content_empty_object() {
+        let vite_config = r#"import { defineConfig } from 'vite-plus';
+
+export default defineConfig({
+  lint: { options: { typeAware: true, typeCheck: true } },
+});"#;
+
+        let result = merge_json_config_content(vite_config, "{}", "fmt").unwrap();
+        assert!(result.updated);
+        assert_eq!(
+            result.content,
+            r#"import { defineConfig } from 'vite-plus';
+
+export default defineConfig({
+  fmt: {},
+  lint: { options: { typeAware: true, typeCheck: true } },
+});"#
+        );
+    }
+
+    #[test]
     fn test_merge_tsdown_config_content_simple() {
         let vite_config = r#"import { defineConfig } from 'vite-plus';
 
--- a/packages/cli/binding/Cargo.toml
+++ b/packages/cli/binding/Cargo.toml
@@ -10,6 +10,7 @@ rolldown = ["dep:rolldown_binding"]
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
+cow-utils = { workspace = true }
 fspy = { workspace = true }
 rustc-hash = { workspace = true }
 napi = { workspace = true }
--- a/packages/cli/binding/src/cli.rs
+++ b/packages/cli/binding/src/cli.rs
@@ -12,6 +12,7 @@ use clap::{
     Parser, Subcommand,
     error::{ContextKind, ContextValue, ErrorKind},
 };
+use cow_utils::CowUtils;
 use owo_colors::OwoColorize;
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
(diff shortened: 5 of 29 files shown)


packages/cli/src/migration/bin.ts
```diff
@@ -168,6 +171,21 @@ async function promptPrettierMigration(
   return true;
 }
 
+async function confirmNodeVersionFileMigration(interactive: boolean): Promise<boolean> {
+  if (interactive) {
+    const confirmed = await prompts.confirm({
+      message: 'Migrate .nvmrc to .node-version?',
```
naokihaba, line 177: In Vite+, the Node.js version is determined based on the following priority order:

The sequence is ".node-version", followed by "engines.node" in package.json, and then "devEngines.runtime" in package.json.
For this reason, we are treating the migration of the ".node-version" file as the highest priority in this migration process.


https://github.com/voidzero-dev/vite-plus/blob/d2e4dfd821500580bbd976a95112978f02944b64/crates/vite_global_cli/src/commands/env/config.rs#L197-L207

naokihaba:
@codex review

packages/cli/src/migration/migrator.ts
```diff
@@ -1977,3 +1977,82 @@ function setPackageManager(
...
+    return null;
+  }
+
+  // LTS aliases (lts/*, lts/iron, etc.) pass through as-is
+  if (version.startsWith('lts/')) {
+    return version;
+  }
+
+  // Strip optional 'v' prefix
+  return version.startsWith('v') ? version.slice(1) : version;
```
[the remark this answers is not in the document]
naokihaba: Fixed parseNvmrcVersion. It now strips the v prefix and performs validation using semver.validRange, returning null if it is not a valid SemVer version or range. The process to remove .nvmrc is now executed only if the parsed version is non-null.


db3da18

packages/cli/src/migration/migrator.ts
```diff
@@ -1977,3 +1977,86 @@ function setPackageManager(
...
+  }
+  return undefined;
+}
+
+/**
+ * Parse a version string from a .nvmrc file.
+ * Returns null for unsupported aliases like "node", "stable", "system".
+ */
+export function parseNvmrcVersion(content: string): string | null {
+  const version = content.split('\n')[0]?.trim();
```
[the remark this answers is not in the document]
naokihaba: The existing process for reading .node-version (parse_node_version_content) is also designed to read only the first line without excluding comments.

Therefore, this implementation follows the same limitations to maintain consistency across the codebase.
If necessary, I can revisit this and make further adjustments.

packages/cli/src/migration/migrator.ts
```diff
@@ -1977,3 +1977,86 @@ function setPackageManager(
...
+
+  if (!version) {
+    return null;
+  }
+
+  // Unsupported nvm aliases that have no direct version equivalent
+  if (['node', 'stable', 'iojs', 'system', 'default'].includes(version)) {
```
fengmk2: we can convert all these to `lts/*` ?

naokihaba: Thank you for the suggestion!

As you suggested, I would like to modify the implementation to map aliases such as node and stable to lts/*.

One thing I would like to ensure is that this conversion does not happen silently in the background. I want to explicitly inform the user through a log message, for example: "node" is not a specific version, so it will be automatically mapped to "lts/*".

On a side note, there was a reason why I initially avoided this mapping. Under the current release schedule, odd-numbered versions like Node 25 do not become LTS (Long Term Support). I was concerned that an automatic conversion might result in an unintended version downgrade for the user.

However, I read the announcement that starting from Node 27, all releases will become LTS. This makes it clear that the gap between node/stable and lts/* will narrow in the future. Therefore, I am now fully on board with proceeding with the approach you proposed.

https://nodejs-org-git-fork-ulisesgascon-release-announcement-openjs.vercel.app/en/blog/announcements/evolving-the-nodejs-release-schedule

I will update the implementation accordingly.

[93ae0f5](https://github.com/voidzero-dev/vite-plus/pull/1159/commits/93ae0f5f67025cd47d43ac85f846cc6a369c6dff)

packages/cli/src/migration/migrator.ts
```diff
@@ -1977,3 +1977,102 @@ function setPackageManager(
...
+    return false;
+  }
+
+  // TODO: remove this log once Node 27+ makes all releases LTS, at which point
+  // "node"/"stable" and "lts/*" will be effectively equivalent.
+  if (version === 'lts/*' && (originalAlias === 'node' || originalAlias === 'stable')) {
+    prompts.log.info(
+      `"${originalAlias}" in .nvmrc is not a specific version; automatically mapping to "lts/*"`,
+    );
+  }
```
naokihaba, line 2087: Starting from Node 27, every release will be an LTS (Long Term Support) version, meaning there will no longer be a distinction between node/stable and lts/*. We plan to stop displaying this log as soon as this transition is complete.

packages/cli/src/migration/migrator.ts
```diff
@@ -1977,3 +1977,102 @@ function setPackageManager(
...
+export function migrateNodeVersionManagerFile(
+  projectPath: string,
+  _detection: NodeVersionManagerDetection,
+  report?: MigrationReport,
+): boolean {
+  const sourcePath = path.join(projectPath, '.nvmrc');
+  const nodeVersionPath = path.join(projectPath, '.node-version');
+  const content = fs.readFileSync(sourcePath, 'utf8');
+  const originalAlias = content.split('\n')[0]?.trim();
+  const version = parseNvmrcVersion(content);
```
fengmk2: Passing originalAlias to parseNvmrcVersion eliminates the need to repeat content processing in parseNvmrcVersion.

packages/cli/snap-tests-global/migration-nvmrc-lts/snap.txt
```diff
@@ -0,0 +1,14 @@
+> vp migrate --no-interactive # migration should detect .nvmrc with lts alias and auto-migrate
+VITE+ - The Unified Toolchain for the Web
+
+
+.nvmrc detected. Auto-migrating to .node-version...
```
fengmk2: Is this log prompt unnecessary? It seems a bit redundant with the log below.

packages/cli/src/migration/migrator.ts
naokihaba: The processing content.split('\n')[0]?.trim() was being called twice in total—once within parseNvmrcVersion and once within migrateNodeVersionManagerFile—to extract the originalAlias. I have fixed this as it was a redundant repetition of the same logic.

[d3997a3](https://github.com/voidzero-dev/vite-plus/pull/1159/commits/d3997a3867a350606e99b2408df728f0d5dc74a7)

[d8c1ed4](https://github.com/voidzero-dev/vite-plus/pull/1159/commits/d8c1ed4943d8588e49f92726604c024f9a9d2dc8)

packages/cli/snap-tests-global/migration-nvmrc-lts/snap.txt
naokihaba: You are right. Both logs convey the same information, which is that .nvmrc is being migrated to .node-version.

Originally, I added the pre-migration log to notify users running in non-interactive mode (--no-interactive) that the process was about to happen, since they wouldn't see a confirmation prompt. However, the single summary line, "• Node version manager file migrated to .node-version," is sufficient to communicate the outcome.

I will go ahead and remove the pre-migration log.

[085b9df](https://github.com/voidzero-dev/vite-plus/pull/1159/commits/085b9df522d6a0f91c0094d5d189c7e99582e361)

PR #1162, opened 2026-03-27
feat(cli): expose bundled tool versions via `vite-plus/versions` export

### LINKED ISSUE #1149: Provide a JSON file on the vendored libraries in vite plus
### Description

Hi :wave: 

Currently, `vitest` is being vendored by `vite-plus`. Third party tools canno't really retrieve the version being vendored by `vite-plus`.
Some needs to do this for compatibility check with tools being aliased by vite-plus

### Suggested solution

have vendored tools provide a json, could be a copy of the package.json from the bundled tool so third party libraries can do

```ts
import { versions } from 'vite-plus'
// or
import { version} from 'vite-plus/test'
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

kazupon:
resolve #1149


packages/cli/build.ts
```diff
@@ -425,6 +429,67 @@ async function syncTestPackageExports() {
   console.log(`\nSynced ${Object.keys(generatedExports).length} exports from test package`);
 }
 
+/**
+ * Read version from a dependency's package.json in node_modules.
+ * Uses readFile because these packages don't export ./package.json.
+ */
+async function readDepVersion(packageName: string): Promise<string | null> {
```
kazupon, line 444: Unfortunately, oxlint and oxfmt packages is not exported `pacakge.json` as `exports` filed. /cc @leaysgur @camc314

leaysgur: If needed, would you please create issue for oxc repo with detailed context? 🙏🏻

kazupon: I've just push the PR
https://github.com/oxc-project/oxc/pull/20784

PR #1164, opened 2026-03-27
refactor(cli): simplify `vp --version` by using generated versions module

kazupon:
ref: https://github.com/voidzero-dev/vite-plus/pull/1162#issuecomment-4139872021

This branch is derrived from #1162


fengmk2:
flaky snap test will fix in https://github.com/voidzero-dev/vite-plus/pull/1209

PR #1181, opened 2026-03-28
feat(completion): implement dynamic shell completion with `vp run` task support

### LINKED ISSUE #1131: Shell completion support for `vp run` command
### Description

Currently, I need to run `vp run` without additional params to see all available commands in the `package.json` file. 

Having `vp run` <kbd>TAB</kbd> for scripts in the `package.json` file would be nice to have.

### Additional context

Related to #950 

### Validations

nekomoyi:
## Summary
- Switch to dynamic shell completion system
- Add task name completion for `vp run` command, supporting workspace packages and tasks
### Dynamic Completion System
Replaced the static completion files with a dynamic completion system that can provide more flexible suggestions based on the current workspace state.

### Task Name Completion for `vp run`
Implemented dynamic task completion for the `vp run` command, supporting task format like `task` or `package#task`

## Known Limitations
- Only Category A command arguments are supported in completions. Category B and C subcommand parameters are pending implementation.
- `vp run` only completes tasks; parameters like `-r`, `-t` are not yet supported

Awaiting future work to sync local CLI parameters to global CLI
close #1131


crates/vite_global_cli/src/commands/env/setup.rs
```diff
@@ -41,9 +40,6 @@ pub async fn execute(refresh: bool, env_only: bool) -> Result<ExitStatus, Error>
     // Ensure home directory exists (env files are written here)
     tokio::fs::create_dir_all(&vite_plus_home).await?;
 
-    // Generate completion scripts
-    generate_completion_scripts(&vite_plus_home).await?;
```
fengmk2, line 45: @nekomoyi Does the old file ~/.vite-plus/completion/ that is no longer being used need to be deleted?

nekomoyi: @fengmk2 The files are no longer needed. Is it necessary to add deletion logic, or just leave them there?

fengmk2: @nekomoyi I think we should still add the auto-cleanup logic. We can add a todo code comment: remove this cleanup logic before the beta release.

PR #1185, opened 2026-03-28
fix: get vite_plus_home from node wrapper's path

### LINKED ISSUE #1184: `$HOME` env overriding makes Vite+ stuck stuck in an infinite loop, especially for Agents use
### Describe the bug

The current vp would find executable node with `~/.vite-plus` dir. However, if we override `$HOME` environment var to a empty / no-vp dir, vp would download another node version.

And it causes problems, Vite+ also has an `VITE_PLUS_TOOL_RECURSION` env var, and if it is greater than 0, vp would pass through commands to `node` installed by vp including the downloading node program mentioned above. And an infinite loop happens here (as run `node` will trigger download node program over and over again).

The conditions that triggered it sounded very extreme but actually quite common. `$HOME` overriding usually happens for e2e tests in global cli projects, and if we use agents like `Claude Code`, `Open Code` which based on Node.js, and try to run tests inside, it's easy to trigger this case.

A possible solution is to find `node` by the getting the grandparent directory from the `node` command wrapper's path (`~/.vite-plus/bin/node`)

### Reproduction

https://github.com/liang-demos/vite-plus-node-hanging.git

### Steps to reproduce

Follow the document in the repo. ***Do not forget to kill the process to protect your computer!***

### System Info

```shell
% vp env current && vp --version

Environment:
  Version       22.18.0
  Source        .node-version
  Source Path   /Users/liangmi/code/voidzero-dev/vite-plus/.node-version
  Project Root  /Users/liangmi/code/voidzero-dev/vite-plus

Tool Paths:
  node  /Users/liangmi/.vite-plus/js_runtime/node/22.18.0/bin/node
  npm   /Users/liangmi/.vite-plus/js_runtime/node/22.18.0/bin/npm
  npx   /Users/liangmi/.vite-plus/js_runtime/node/22.18.0/bin/npx

vp v0.1.14

Local vite-plus:
  vite-plus  v0.0.0

Tools:
  vite             v8.0.3
  rolldown         v1.0.0-rc.12
  vitest           v4.1.2
  oxfmt            v0.42.0
  oxlint           v1.57.0
  oxlint-tsgolint  v0.17.4
  tsdown           v0.21.5

Environment:
  Package manager  pnpm v10.28.0
  Node.js          v22.18.0 (.node-version)
```

### Validations

liangmiQwQ:
Close #1184

### DIFF
--- a/crates/vite_shared/src/home.rs
+++ b/crates/vite_shared/src/home.rs
@@ -22,6 +22,7 @@ pub fn get_vite_plus_home() -> std::io::Result<AbsolutePathBuf> {
     }
 
     // Get from `node` executable file's grandparent directory (~/.vite-plus/bin/node)
+    // For the case where `$HOME` is overridden
     if let Ok(path) = which("node")
         && let Some(parent) = path.parent()
         && let Some(grandparent) = parent.parent()


liangmiQwQ:
Done! 

This PR used `which` crate to find `vp`'s executable file and it's grandparent directory is expected to be `vite_plus_home`.

The reason not to use `std::env::current_exe()` is that `node` is essentially a symlink, and this function would get `vp` binary file's path. It makes both tests and debugs harder and less stable than a certain path in `$PATH` var. (`$HOME` reading logic is also kept if the `node` is not controlled by `vp`) 

As `which` crate is also used in `vite_command` crate, so there is no cost for new dependencies.

Thanks!

PR #1198, opened 2026-03-29
fix(cache): add Package-scoped negative globs to prevent cache miss in monorepos

fengmk2:
In monorepos, each package has its own node_modules/. When vp build/test/pack
run, Vite creates .vite-temp/ config files and dist/ outputs that are both read
and written during execution. The existing !node_modules/.vite-temp/** glob at
Workspace scope only matched the workspace root, not package-level paths,
causing perpetual cache misses due to read-write overlap detection.

- Build: add Package-scoped !node_modules/.vite-temp/** and !dist/**
- Test: add Package-scoped !node_modules/.vite-temp/** and !node_modules/.vite/**/results.json
- Pack: add Package-scoped !node_modules/.vite-temp/** and !dist/**

### DIFF
--- a/packages/cli/binding/src/cli.rs
+++ b/packages/cli/binding/src/cli.rs
@@ -332,13 +332,7 @@ impl SubcommandResolver {
                     cache_config: UserCacheConfig::with_config(EnabledCacheConfig {
                         env: Some(Box::new([Str::from("VITE_*")])),
                         untracked_env: None,
-                        input: Some(vec![
-                            UserInputEntry::Auto(AutoInput { auto: true }),
-                            UserInputEntry::GlobWithBase(GlobWithBase {
-                                pattern: Str::from("!node_modules/.vite-temp/**"),
-                                base: InputBase::Workspace,
-                            }),
-                        ]),
+                        input: Some(build_pack_cache_inputs()),
                     }),
                     envs: merge_resolved_envs_with_version(envs, resolved.envs),
                 })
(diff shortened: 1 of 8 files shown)


packages/cli/binding/src/cli.rs
```diff
@@ -338,6 +338,14 @@ impl SubcommandResolver {
                                 pattern: Str::from("!node_modules/.vite-temp/**"),
                                 base: InputBase::Workspace,
                             }),
+                            UserInputEntry::GlobWithBase(GlobWithBase {
+                                pattern: Str::from("!node_modules/.vite-temp/**"),
```
fengmk2: After testing, still need to add the blacklist under the package directory.

fengmk2: cc @branchseer

packages/cli/binding/src/cli.rs
```diff
@@ -338,6 +338,14 @@ impl SubcommandResolver {
...
+                                base: InputBase::Package,
+                            }),
+                            UserInputEntry::GlobWithBase(GlobWithBase {
+                                pattern: Str::from("!dist/**"),
```
fengmk2: The dist directory filtering is currently hardcoded. Later, it should first read the vite.config.ts and command line parameter configurations to make the determination. It would be better to have the actual tool report this, otherwise it would be difficult to implement completely.

@branchseer Do you have any thoughts? Or should we not add this dist filtering rule for now? Let users configure it proactively first?

packages/cli/binding/src/cli.rs
```diff
@@ -504,6 +499,22 @@ impl SubcommandResolver {
...
+}
+
+/// Common cache input entries for build/pack commands.
+/// Excludes .vite-temp config files and dist output files that are both read and written.
+fn build_pack_cache_inputs() -> Vec<UserInputEntry> {
+    vec![
+        UserInputEntry::Auto(AutoInput { auto: true }),
+        exclude_glob("!node_modules/.vite-temp/**", InputBase::Workspace),
+        exclude_glob("!node_modules/.vite-temp/**", InputBase::Package),
+        exclude_glob("!dist/**", InputBase::Package),
```
fengmk2, line 516: The dist directory filtering is currently hardcoded. Later, it should first read the vite.config.ts and command line parameter configurations to make the determination. It would be better to have the actual tool report this, otherwise it would be difficult to implement completely.

@branchseer Do you have any thoughts? Or should we not add this dist filtering rule for now? Let users configure it proactively first?

PR #1215, opened 2026-03-30
feat: add lazyPlugins() helper for conditional plugin loading

fengmk2:
## Summary

Add `lazyPlugins()` helper that conditionally loads plugins based on the `VP_COMMAND` environment variable. Plugins load for vite commands (`dev`, `build`, `test`, `preview`) and when `VP_COMMAND` is unset (e.g., running vitest directly or via VS Code extension). Skipped for non-vite commands (`lint`, `fmt`, `check`, etc.).

Refs https://github.com/vitejs/vite/issues/22085

### Usage

```ts
import { defineConfig, lazyPlugins } from 'vite-plus'

// sync
export default defineConfig({
  plugins: lazyPlugins(() => [fooPlugin()]),
})

// async with dynamic import
export default defineConfig({
  plugins: lazyPlugins(async () => {
    const { default: heavyPlugin } = await import('vite-plugin-heavy');
    return [heavyPlugin()];
  }),
})
```

### How it works

- `bin.ts` sets `VP_COMMAND` env var from the CLI command argument
- Rust resolver injects `VP_COMMAND` into child process envs for synthesized subcommands (fixes `vp run build` case)
- `lazyPlugins` checks `VP_COMMAND` — executes callback for vite commands, returns `undefined` otherwise
- Async callbacks have their Promise wrapped in an array internally for Vite's `asyncFlatten`

## Test plan

- [x] Unit tests: 30 tests — lazyPlugins behavior, type compatibility with `plugins` field, defineConfig compatibility, vitest plugin hooks
- [x] Snap tests: 6 cases — sync build, async build, vitest plugin, skip-on-lint, `vp run build`, `vp run lint`
- [x] `vp check` passes (format + lint)
- [x] Windows CI: forward-slash path normalization fix for snap test stability


docs/config/index.md
```diff
@@ -29,3 +29,66 @@ Vite+ extends the basic Vite configuration with these additions:
...
+export default defineConfig(async () => ({
+  lazy: async () => {
+    const { default: heavyPlugin } = await import('vite-plugin-heavy');
+    return { plugins: [heavyPlugin()] };
+  },
+}));
+```
+
+::: info
+The `lazy` field is a temporary Vite+ extension. We plan to support this in upstream Vite in the future.
```
leaysgur: Is this something listed in the roadmap or an issue on the Vite side?
(Sorry, I might just not be aware of)

leaysgur (commented):
I have no objection to recommending this method as a current workaround. 👌🏻 

(However, I cannot judge whether this is the best approach, as I do not know the background behind why this `lazy` was introduced.)

docs/config/index.md
fengmk2: Not yet, @sapphi-red will Vite consider this problem? Or do I need to submit an issue first?

sapphi-red: Yes, you'll need to submit an issue first.

BTW, wouldn't the following config have the same effect?
```ts

export default defineConfig({
  plugins: [async () => {
    const { default: heavyPlugin } = await import('vite-plugin-heavy');
    return [heavyPlugin()];
  }],
});
```

fengmk2: Thank you very much for the reminder. If this works, then let's not expose lazy for now, since it's a temporary solution specific to Vite+ and not suitable for public use.

docs/config/index.md
```diff
@@ -29,3 +29,66 @@ Vite+ extends the basic Vite configuration with these additions:
...
+```ts
+lazy?: () => Promise<{
+  plugins?: Plugin[];
+}>;
+```
+
+### Merging with Existing Plugins
+
+Plugins returned from `lazy` are appended after any plugins already in the `plugins` array. This lets you keep lightweight plugins inline and defer only the expensive ones:
+
```
wan9chi: How about: supporting all vite config fields in the returned object, and making `lazy` mutuably exclusive from all the top-level vite config fields?

I propose this because I'm concerned about evaluating other config fields also being slow. Forcing them all to be lazy makes performance more predictable. Also, I don't see a use case where both `lazy` and eager loading are needed simultaneously.

docs/config/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+When `vite.config.ts` imports heavy plugins at the top level, every `import` is evaluated eagerly, even for commands like `vp lint` or `vp fmt` that don't need those plugins. This can make config loading noticeably slow.
+
+Vite supports promises in the `plugins` array, so you can use dynamic `import()` to defer heavy plugin loading:
+
+```ts
+import { defineConfig } from 'vite-plus';
+
+export default defineConfig({
+  plugins: [
+    import('vite-plugin-heavy').then((m) => m.default()),
```
fengmk2: @leaysgur @cpojer @branchseer I referenced @sapphi-red's suggestion and changed to using Vite's standard solution to resolve the slow configuration loading issue caused by plugins. The slow loading problem of other configuration items is temporarily not what this PR is meant to discuss and solve.

sapphi-red: Won't this still cause the plugin to be executed eagerly? I think this should be:
```suggestion
    () => import('vite-plugin-heavy').then((m) => m.default()),
```

fengmk2: You are right that bare `import().then()` is evaluated eagerly.

However, wrapping it in `() => import(...)` does not work because Vite treats functions in the `plugins` array as plugin objects, not as factories to be called. It fails with:

```
Cannot assign to read only property 'name' of object '[object Object]'
```

Same issue with `async () => { ... }` (fails with `[object AsyncFunction]`).

Vite's `PluginOption` type is `Plugin | false | null | undefined | PluginOption[] | Promise<PluginOption>`, which does not include functions. So there is no native way to truly lazy-load plugins in Vite today. The `lazy` field in vite-plus was created specifically to fill this gap.

sapphi-red: Ah, true...

fengmk2: @sapphi-red 😢 let me revert to `lazy` field for now

docs/config/troubleshooting.md
```diff
@@ -0,0 +1,42 @@
...
+Use the `vitePlugins()` helper to conditionally load plugins. It checks which `vp` command is running and skips plugin loading for commands that don't need them (like `lint`, `fmt`, `check`):
+
+```ts
+import { defineConfig, vitePlugins } from 'vite-plus';
+
+import myPlugin from 'vite-plugin-foo';
+
+export default defineConfig({
+  plugins: [
+    vitePlugins(() => [myPlugin()]),
```
TheAlexLichter: The `vitePlugins` naming is a bit confusing to me. If we can somehow indicate that they are skipped for vp lint/fmt/check in the helper name that'd be nice.

TheAlexLichter: Closest I come is `lazyPlugins` but that might be too ambiguous. Another one was `viteOnlyPlugins`.

cpojer:
I don't really follow, why do we need to add a new function to wrap plugins instead of redefining the plugin type from `plugins: Array<Plugin>` to `plugins: Array<Plugin> | () => Promise<Array<Plugin>>`?

fengmk2:
@codex review

packages/cli/src/define-config.ts
```diff
@@ -25,75 +24,74 @@ declare module '@voidzero-dev/vite-plus-core' {
...
+    return [result];
+  }
+  return result;
+}
+
+function transformConfig(config: VPUserConfig): UserConfig {
+  if (config.plugins && typeof config.plugins === 'function') {
+    return { ...config, plugins: resolvePlugins(config.plugins) } as UserConfig;
+  }
+  return config as UserConfig;
```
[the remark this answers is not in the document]
fengmk2: lazy has never been publicly exposed, so compatibility is not a consideration

docs/.vitepress/config.mts
```diff
@@ -166,6 +166,7 @@ export default extendConfig(
                 { text: 'Build', link: '/config/build' },
                 { text: 'Pack', link: '/config/pack' },
                 { text: 'Staged', link: '/config/staged' },
+                { text: 'Troubleshooting', link: '/config/troubleshooting' },
```
cpojer: I'm not a huge fan of adding a section called "Troubleshooting" to the config section on the website because we already have a troubleshooting page for Vite+ in the guide section. We should either merge them into one, or call this one "Config Troubleshooting".

cpojer: I prefer a single page for troubleshooting any type of issue.

docs/config/troubleshooting.md
```diff
@@ -0,0 +1,38 @@
...
+});
+```
+
+::: warning
+The plugins factory requires `defineConfig` from `vite-plus`, not from `vite`. Vite's native `defineConfig` does not support factory functions for the `plugins` field.
+:::
+
+::: info
+The plugins factory works by checking the `VP_COMMAND` environment variable, which is automatically set by `vp` for every command.
+:::
```
cpojer: Is this relevant enough to be in the docs? As a user, what am I supposed to do with this information?

cpojer (approved):
It's beautiful. From a UX perspective, I love it. I don't know if there are any remaining concerns from the Vite side, cc @sapphi-red if you don't mind signing off on this 🙇‍♂️

packages/cli/src/define-config.ts
```diff
@@ -25,75 +24,74 @@ declare module '@voidzero-dev/vite-plus-core' {
...
+    }
+    return viteDefineConfig(transformConfig(config));
+  }
+  return viteDefineConfig(config as UserConfig);
+}
+
+const VITE_COMMANDS = new Set(['dev', 'build', 'test', 'preview']);
+
+function isViteCommand(): boolean {
+  return VITE_COMMANDS.has(process.env.VP_COMMAND ?? '');
```
sheremet-va: Does it make sense to default to `true` if `VP_COMMAND` is not defined? For example, if I directly run `vitest` or `node ./node_modules/vite-plus/test/bin` (I don't remember the full path). The vscode extension runs vite-plus like this:

```shell
/Users/vladimir/.nvm/versions/node/v24.13.0/bin/node /Users/vladimir/Projects/vscode/dist/worker.js
```

Then the worker file just imports `vite-plus/test/node` and calls `createVitest` from it

Unless `vite-plus/test/node` sets `VP_COMMAND`, it won't be able to see the plugin. (Alternatively, we could also set `VP_COMMAND`, I guess)

fengmk2: Makes sense, if VP_COMMAND doesn't exist it should return `true`, which can ensure maximum compatibility. I will change it. 👍

packages/cli/src/define-config.ts
```diff
@@ -25,75 +24,78 @@ declare module '@voidzero-dev/vite-plus-core' {
...
+  }
+  return viteDefineConfig(config as UserConfig);
+}
+
+const VITE_COMMANDS = new Set(['dev', 'build', 'test', 'preview']);
+
+function shouldLoadPlugins(): boolean {
+  const cmd = process.env.VP_COMMAND;
+  if (!cmd) {
+    return true;
```
fengmk2: @sheremet-va fixed

sheremet-va: Thank you

fengmk2:
@cursor review

fengmk2:
@cursor review

packages/cli/snap-tests/vite-plugins-async-test/src/index.test.ts
```diff
@@ -0,0 +1,11 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+import { expect, test } from '@voidzero-dev/vite-plus-test';
+
+test('async plugin factory should load vitest plugin with configureVitest hook', () => {
+  const markerPath = path.join(import.meta.dirname, '..', '.vitest-plugin-loaded');
+  expect(fs.existsSync(markerPath)).toBe(true);
+  expect(fs.readFileSync(markerPath, 'utf-8')).toBe('configureVitest hook executed');
+  fs.unlinkSync(markerPath);
```
sheremet-va: If test fails, this line won't be reached. Use `afterEach` or [`onTestFinished`](https://vitest.dev/api/hooks.html#ontestfinished), or `using`

sapphi-red (approved):
It should be ok as a workaround

PR #1252, opened 2026-04-01
fix(init): use bin.js entry point for TypeScript loader in justfile

### LINKED ISSUE #1251: just init ERR_MODULE_NOT_FOUND
### Describe the bug

`just init` fails with `ERR_MODULE_NOT_FOUND` because `justfile` line 21 runs `node packages/tools/src/index.ts sync-remote` directly. Node.js cannot resolve `.js` extension imports (e.g. `import('./sync-remote-deps.js')`) to `.ts` files without a TypeScript loader.

The correct entry point is `packages/tools/src/bin.js`, which registers `@oxc-node/core/register` before importing `index.ts`.

### Reproduction

Clone the repo and run: ``` just init ```

### Steps to reproduce

1. `git clone` the repo
2. Run `just init`
3. Observe `ERR_MODULE_NOT_FOUND` for `sync-remote-deps.js`

### System Info

```shell
Node.js v22.18.0
just 1.48.1
```

### Used Package Manager

pnpm

### Logs

```shell
node packages/tools/src/index.ts sync-remote
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/.../packages/tools/src/sync-remote-deps.js' imported from /Users/.../packages/tools/src/index.ts
```

### Validations

jong-kyung:
## Summary
Closes #1251 

- `just init` fails with `ERR_MODULE_NOT_FOUND` because `node packages/tools/src/index.ts` runs without a TypeScript loader, so Node.js cannot resolve `.js` imports to `.ts` files
- Changed to `node packages/tools/src/bin.js` which registers `@oxc-node/core/register` before importing `index.ts`, enabling proper `.js` → `.ts` resolution

### DIFF
--- a/justfile
+++ b/justfile
@@ -18,7 +18,7 @@ _clean_dist:
 
 init: _clean_dist
   cargo binstall watchexec-cli cargo-insta typos-cli cargo-shear dprint taplo-cli -y
-  node packages/tools/src/index.ts sync-remote
+  node packages/tools/src/bin.js sync-remote
   pnpm install
   pnpm -C docs install
 


fengmk2 (approved):

justfile
```diff
@@ -18,7 +18,7 @@ _clean_dist:
 
 init: _clean_dist
   cargo binstall watchexec-cli cargo-insta typos-cli cargo-shear dprint taplo-cli -y
-  node packages/tools/src/index.ts sync-remote
+  node packages/tools/src/bin.js sync-remote
```
fengmk2, line 21: @jong-kyung This change will cause problems, as the "@oxc-node/core" dependency is not installed during the initialization phase. Since we declared that the node version supports running ts directly, the original `node packages/tools/src/index.ts sync-remote` is correct. I will revert this change first.

fengmk2: I know the reason for your modification. Tools should be changed to a pure TypeScript project. Let me refactor it.

jong-kyung: > I know the reason for your modification. Tools should be changed to a pure TypeScript project. Let me refactor it.

@fengmk2 Thanks for the explanation! I didn't consider that `@oxc-node/core` isn't available during the init phase. Looking forward to the refactor.

fengmk2: https://github.com/voidzero-dev/vite-plus/pull/1274

PR #1254, opened 2026-04-01
docs: add custom Node.js mirror docs and rename `VITE_NODE_DIST_MIRROR` to `VP_NODE_DIST_MIRROR`

### LINKED ISSUE #1245: Add note about using a custom Node Mirror
### Documentation is

- [x] Missing
- [ ] Outdated
- [ ] Confusing
- [ ] Not sure

### Explain in Detail

When attempting to use Vite+ Environment management feature with my work's internal artifactory (which is enforced), I'm unable to do most features due to not having access to the default Node mirror (https://nodejs.org/dist). I was able to dig through the source to find that I could utilize `VITE_NODE_DIST_MIRROR` to override the default. 

### Your Suggestion for Changes

I suggest mentioning the ability to override with a custom Node mirror within the environment page. 

### Relevant Page

https://viteplus.dev/guide/env

### Reproduction (Optional)

_No response_

### Steps to reproduce (Optional)

_No response_

kazupon:
resolve #1245

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -15,14 +15,13 @@
       "enabled": false
     },
     {
-      "matchPackageNames": [
+      "matchDepNames": [
         "fspy",
         "vite_glob",
         "vite_path",
         "vite_str",
         "vite_task",
-        "vite_workspace",
-        "https://github.com/voidzero-dev/vite-task"
+        "vite_workspace"
       ],
       "enabled": false
     }
--- a/packages/cli/src/__tests__/resolve-lint.spec.ts
+++ b/packages/cli/src/__tests__/resolve-lint.spec.ts
@@ -1,0 +1,24 @@
+import { existsSync } from 'node:fs';
+
+import { describe, expect, it } from '@voidzero-dev/vite-plus-test';
+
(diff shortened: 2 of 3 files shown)


docs/guide/env.md
```diff
@@ -82,3 +82,18 @@ vp env list-remote --lts      # List only LTS versions
 vp env exec --node lts npm i  # Execute npm with latest LTS
 vp env exec node -v           # Use shim mode with automatic version resolution
 ```
+
+## Custom Node.js Mirror
+
+By default, Vite+ downloads Node.js from `https://nodejs.org/dist`. If you're behind a corporate proxy or need to use an internal mirror (e.g., Artifactory), set the `VITE_NODE_DIST_MIRROR` environment variable:
```
fengmk2: Would it be more appropriate to rename `VITE_NODE_DIST_MIRROR` to `VP_NODE_DIST_MIRROR`? We’ve already switched all other Vite+ related environment variable names to use the `VP_` prefix.

kazupon: I've just fixed this issue.

PR #1255, opened 2026-04-01
feat(env): make `vp env off` disable Node.js management for all vp commands

### LINKED ISSUE #977: Support system Node.js for `vp` commands (not just shims)
## Summary

I'd like an option to use system-installed Node.js for **all** parts of `vp` — not just shims, but also `vp create`, `vp install`, `vp build`, and every other command that delegates to JS. Currently there's no way to do this.

## Problem

`vp` commands that delegate to JS **always download a managed Node.js binary**, even when a compatible `node` is already installed on the system. The existing `system-first` shim mode (`vp env off`) only applies to shimmed tools (`node`/`npm`/`npx` wrappers via `dispatch.rs:679-703`), not to `vp` commands themselves.

### Who's affected

- **NixOS** and **GNU Guix** — downloaded binaries are dynamically linked against `/lib64/ld-linux-x86-64.so.2` which doesn't exist on these distros. `vp create` fails immediately:
  ```
  Could not start dynamically linked executable: ~/.vite-plus/js_runtime/node/24.14.0/bin/node
  NixOS cannot run dynamically linked executables intended for generic linux environments out of the box.
  ```
- **Distro packages** (Nix, Homebrew, AUR, Fedora COPR, etc.) — packagers want to use the distro-managed Node.js
- **Air-gapped environments** — no network access to download from nodejs.org
- **Container images** — Node.js is already installed; downloading a second copy wastes time and space

### Code path showing the gap

```
vp create
  → commands/delegate.rs:execute()
  → JsExecutor::delegate_to_local_cli()
  → js_executor.rs:ensure_project_runtime()
  → download_runtime_for_project()           ← always downloads
  → download_runtime_with_provider()         ← no system check here
```

The shim dispatch has `find_system_tool()` which locates system binaries — but `JsExecutor` never calls it. It goes directly to `download_runtime()` which only checks its own cache at `~/.vite-plus/js_runtime/node/{version}/`.

After resolution, the downloaded Node.js is only used for two things:
- `get_binary_path()` → passed to `Command::new()` to execute JS scripts
- `get_bin_prefix()` → prepended to `PATH` so child processes find `npm`/`npx`

A system Node.js satisfies both of these identically.

## What I'd like

A way to tell `vp` "use my system Node.js everywhere" — for shims **and** for all `vp` commands. Ideally this would be:

1. **An environment variable** like `VITE_PLUS_USE_SYSTEM_NODE=1` that makes all Node.js resolution prefer the system `node` in PATH before downloading
2. **Or extending `system-first` mode** so that `vp env off` applies to `JsExecutor` too — if a user opts into system-first, they expect *all* node usage to prefer system node, not just shims

### Suggested implementation

The simplest approach: add a system check in `download_runtime_with_provider()` (the single bottleneck all download paths funnel through), before the cache check at line 121:

```rust
use std::process::Command as StdCommand;

if runtime_type == JsRuntimeType::Node {
    if std::env::var_os("VITE_PLUS_USE_SYSTEM_NODE").is_some() {
        if let Ok(output) = StdCommand::new("node").arg("--version").output() {
            if output.status.success() {
                if let Ok(version_str) = String::from_utf8(output.stdout) {
                    let version_str = version_str.trim().trim_start_matches('v');
                    if let Some(path_var) = std::env::var_os("PATH") {
                        for dir in std::env::split_paths(&path_var) {
                            let candidate = dir.join("node");
                            if candidate.exists() {
                                if let Some(install_dir) = dir.parent() {
                                    if let Some(install_dir) =
                                        AbsolutePathBuf::new(install_dir.to_path_buf())
                                    {
                                        tracing::info!(
                                            "Using system Node.js v{version_str} at {:?}",
                                            candidate
                                        );
                                        return Ok(JsRuntime {
                                            runtime_type,
                                            version: version_str.into(),
                                            install_dir,
                                            binary_relative_path,
                                            bin_dir_relative_path,
                                        });
                                    }
                                }
                                break;
                            }
                        }
                    }
                }
            }
        }
    }
}
```

This covers all call paths at once:
- `JsExecutor::ensure_project_runtime()` → `download_runtime()` / `download_runtime_for_project()`
- `JsExecutor::ensure_cli_runtime()` → `download_runtime_for_project()`
- Shim dispatch runtime resolution
- `vp env install` / `vp env use`

### Distro packager workaround (current)

Nix currently patches `download_runtime_with_provider()` unconditionally (no env var gate) and wraps the binary with `nodejs` in PATH. This works but diverges from upstream. An official option would let us drop the patch.

fengmk2:
Previously, `vp env off` only affected shim dispatch (node/npm/npx
invoked directly). All vp commands (create, build, install, etc.)
still downloaded managed Node.js via JsExecutor, ignoring the
system-first mode setting.

Now when `vp env off` is active, JsExecutor checks ShimMode before
downloading and uses the system-installed Node.js found in PATH.
This fixes NixOS/Guix (where downloaded binaries fail), air-gapped
environments, and users managing Node.js via other tools (mise, nvm).

- Add JsRuntime::from_system() constructor for system binary paths
- Add system-first checks to ensure_cli_runtime/ensure_project_runtime
- Expose find_system_tool as pub(crate) and deduplicate doctor.rs copy
- Update vp env off/on messaging to reflect broader scope
- Add CI E2E tests for system-first mode on all platforms

Closes #977

### DIFF
--- a/crates/vite_global_cli/src/commands/env/doctor.rs
+++ b/crates/vite_global_cli/src/commands/env/doctor.rs
@@ -7,7 +7,7 @@ use vite_path::{AbsolutePathBuf, current_dir};
 use vite_shared::{env_vars, output};
 
 use super::config::{self, ShimMode, get_bin_dir, get_vp_home, load_config, resolve_version};
-use crate::error::Error;
+use crate::{error::Error, shim};
 
 /// IDE-relevant profile files that GUI-launched applications can see.
 /// GUI apps don't run through an interactive shell, so only login/environment
@@ -114,7 +114,7 @@ pub async fn execute(cwd: AbsolutePathBuf) -> Result<ExitStatus, Error> {
 
     // Section: Configuration
     print_section("Configuration");
-    check_shim_mode().await;
+    let (shim_mode, system_node_path) = check_shim_mode().await;
 
     // Check env sourcing: IDE-relevant profiles first, then all shell profiles
     #[cfg(not(windows))]
@@ -128,7 +128,7 @@ pub async fn execute(cwd: AbsolutePathBuf) -> Result<ExitStatus, Error> {
 
     // Section: Version Resolution
     print_section("Version Resolution");
-    check_current_resolution(&cwd).await;
+    check_current_resolution(&cwd, shim_mode, system_node_path).await;
 
     // Section: Conflicts (conditional)
     check_conflicts();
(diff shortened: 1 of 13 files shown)


cpojer:
Question regarding `vp env off`'s expected behavior. I'm trying to figure out if we need three different modes here. Does it still shim node.js inside of a project that uses Vite+?

I imagine that there might be use-cases for three things:
* Vite+ does not manage Node.js → Always use system Node.js
* Vite+ only manages Node.js inside of projects that use `vite-plus` → otherwise fall back to system Node.js
* Vite+ always manages Node.js

Can you walk me through what our plan is exactly?

PR #1268, opened 2026-04-02
fix(check): do not treat lint warnings as errors in exit code

fengmk2:
When oxlint reports only warnings (0 errors), override the raw exit code
to SUCCESS instead of propagating oxlint's non-zero exit code. This
fixes ecosystem-ci failures where type-aware lint rules (e.g.
typescript-eslint/await-thenable) produce warnings that cause
`vp check --fix` to exit with code 1.


cpojer (approved):
Should this maybe also be fixed upstream in Oxlint?

camc314 (commented):
note oxlint has a `denyWarnings` option that should probably be respected.


Ah i just saw #1424

PR #1269, opened 2026-04-02
fix(lint): resolve tsgolint on Windows with bun package manager

fengmk2:
Bun stores packages in .bun/ cache dirs with symlinks, so the
hardcoded node_modules/.bin paths don't contain tsgolint.exe.
Add a fallback that resolves the real path of vite-plus and looks
for tsgolint in the sibling .bin directory.


cpojer (approved):
😬

PR #1270, opened 2026-04-02
feat(cli): add explanations to migration prompts

### LINKED ISSUE #1030: Add explanation when configs are merged/migrated/...
### Description

> Would recommend some in-terminal explanations of why merging/keeping some of the different configs is being offered as an option, but that's also small details.

### Suggested solution

Add info where possible (e.g. why it is prompted to skip or merge .vscode files, same for agents etc.)

### Alternative

_No response_

### Additional context

https://x.com/sachasayan/status/2034302785777193428?s=20

### Validations

hakshu25:
## Summary

Closes #1030

When running `vp migrate` interactively, users are prompted with choices (which agents to use, whether to merge editor configs, whether to migrate ESLint/Prettier, etc.) but are given no context for why. This PR adds a brief gray explanation embedded directly in each prompt message so users understand what they're agreeing to before making a choice.

## Demo

スクリーンショット 2026-04-02 17 05 23

## Test plan

- [x] `vp check --fix` passes
- [x] `pnpm bootstrap-cli && pnpm test` passes (snap tests unaffected — all new messages are inside `interactive` guards)
- [x] GPG-signed commits


packages/cli/src/utils/editor.ts
```diff
@@ -181,7 +184,12 @@ export async function selectEditor({
...
     };
     const selectedEditor = await prompts.select({
-      message: 'Which editor are you using?',
+      message:
+        'Which editor are you using?\n  ' +
+        styleText(
+          'gray',
+          'Writes editor config files to enable recommended extensions and Oxlint/Oxfmt integrations.',
+        ),
       options: [...editorOptions, noneOption],
```
TheAlexLichter: One more change here: Let's reword the `noneOption` to `Other` as some editors don't need or have any recommendations.

hakshu25: Thanks for the suggestion! Renamed to Other (also updated the variable name from
  noneOption to otherOption).
  [63aa46c](https://github.com/voidzero-dev/vite-plus/pull/1270/commits/63aa46c0bcc9089e0aee6dc6fa7dd7d912a7d3a7)
[3cd4dad](https://github.com/voidzero-dev/vite-plus/pull/1270/commits/3cd4dad23dae89e083a88cae43f6c8033d664792)

PR #1274, opened 2026-04-02
refactor(tools): use .ts import extensions

fengmk2:
Switch tools package imports from .js to .ts extensions, update entry
point from bin.js to index.ts, and enable allowImportingTsExtensions in
tsconfig. Also fix potential null reference on wasiBindingError in the
NAPI binding loader.

https://github.com/voidzero-dev/vite-plus/pull/1252#discussion_r3027858446

### DIFF
--- a/packages/cli/binding/index.cjs
+++ b/packages/cli/binding/index.cjs
@@ -725,13 +725,17 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       wasiBindingError = err;
     }
   }
-  if (!nativeBinding) {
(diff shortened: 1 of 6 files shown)


packages/cli/binding/index.cjs
```diff
@@ -725,13 +725,17 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       wasiBindingError = err;
     }
   }
-  if (!nativeBinding) {
+  if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
```
fengmk2, line 728: auto gen by napi-rs

PR #1310, opened 2026-04-05
fix: correctly resolve tsgolint in yarn monorepo packages

### LINKED ISSUE #1296: `vp check` fails in Yarn monorepo nested directory
### Describe the bug

Yarn monorepos do not create a `node_modules` under every package, unlike pnpm. When running `vp check` from any nested directory, it fails with:

```
error: Linting could not start
Failed to find tsgolint executable: OXLINT_TSGOLINT_PATH points to '.\node_modules\.bin\tsgolint.cmd' which does not exist
```

This works correctly for `pnpm` monorepos, as pnpm symlinks `node_modules` in every package.

### Steps to reproduce

1. Create a new Yarn monorepo, note this currently fails due to #1295, so setting `nodeLinker: node-modules` is necessary after running `vp create`
2. Run `vp check` - works
3. `cd packages/utils && vp check` - fails

### System Info

```shell

Environment:
  Version       24.4.0
  Source        engines.node

Tool Paths:
  node  C:\Users\Matei\.vite-plus\js_runtime\node\24.4.0\node.exe
  npm   C:\Users\Matei\.vite-plus\js_runtime\node\24.4.0\npm.cmd
  npx   C:\Users\Matei\.vite-plus\js_runtime\node\24.4.0\npx.cmd

Local vite-plus:
  vite-plus  v0.1.15

Tools:
  vite             v8.0.3
  rolldown         v1.0.0-rc.12
  vitest           v4.1.2
  oxfmt            v0.43.0
  oxlint           v1.58.0
  oxlint-tsgolint  v0.18.1
  tsdown           v0.21.7

Environment:
  Package manager  yarn v4.13.0
  Node.js          v24.4.0 (engines.node)
```

### Used Package Manager

yarn

### Validations

rChaoz:
Closes #1296

The `oxlint-tsgolint` executable resolution algorithm assumes that there is a `node_modules` directory in the current working directory (current package directory). However, this is only true for `pnpm`, `yarn` only creates a single `node_modules` in the workspace root, unlike `pnpm` which creates one under every package.

Because of this, `vp lint`/`vp check` fail to find the `tsgolint` executable. Updates the algorithm to extract the `node_modules` directory location based on `resolve()`, rather than `process.cwd()`.

Additionally improves the error message to list all tried paths, rather than just one:

**Before**
```

Failed to find tsgolint executable: OXLINT_TSGOLINT_PATH points to '.\node_modules\.bin\tsgolint.cmd' which does not exist
```

**Now**
```

error: Failed to resolve lint command: GenericFailure, Error: Unable to resolve oxlint-tsgolint executable, tried:
- C:\Users\Matei\projects\agent-lab\node_modules\vite-plus\node_modules\.bin\tsgolint.exe
- C:\Users\Matei\projects\agent-lab\node_modules\vite-plus\node_modules\.bin\tsgolint.cmd
- C:\Users\Matei\projects\agent-lab\node_modules\.bin\tsgolint.exe
- C:\Users\Matei\projects\agent-lab\node_modules\.bin\tsgolint.cmd
```

### DIFF
--- a/packages/cli/install.ps1
+++ b/packages/cli/install.ps1
@@ -374,8 +374,9 @@ function Main {
         $installLog = Join-Path $VersionDir "install.log"
         Push-Location $VersionDir
         try {
-            $env:CI = "true"
-            & "$BinDir\vp.exe" install --silent *> $installLog
+            # Use cmd /c so CI=true is scoped to the child process only,
+            # avoiding leaking it into the user's shell session.
+            cmd /c "set CI=true && `"$BinDir\vp.exe`" install --silent" *> $installLog
             if ($LASTEXITCODE -ne 0) {
                 Write-Host "error: Failed to install dependencies. See log for details: $installLog" -ForegroundColor Red
                 exit 1
--- a/packages/cli/snap-tests-global/new-vite-monorepo-bun/snap.txt
+++ b/packages/cli/snap-tests-global/new-vite-monorepo-bun/snap.txt
@@ -66,7 +66,7 @@ website
     "preview": "vp preview"
   },
   "devDependencies": {
-    "typescript": "~5.9.3",
+    "typescript": "~6.0.2",
     "vite": "catalog:",
     "vite-plus": "catalog:"
   }


packages/cli/src/resolve-lint.ts
```diff
@@ -67,7 +68,10 @@ export async function lint(): Promise<{
       }
     }
     if (!oxlintTsgolintPath) {
-      oxlintTsgolintPath = join(cwdBinDir, 'tsgolint.cmd');
```
rChaoz, line 70: Note: this default value was guaranteed not to exist as `existsSync()` returned `false` for it a few lines above

PR #1272, opened 2026-04-02
fix(upgrade): bypass package manager release age gates during `vp upgrade`

kazupon:
## Summary

resolves #1260

`vp upgrade` fails when a package manager's global config enforces a minimum release age (e.g. pnpm's `minimumReleaseAge`), because newly published vite-plus versions are blocked during dependency installation.

## Changes

- Write local config overrides to the version directory before `vp install`, bypassing release age gates for all package managers:
  - `.npmrc` — pnpm (`minimum-release-age=0`) and npm (`min-release-age=0`)
  - `.yarnrc.yml` — yarn (`npmMinimalAgeGate: "0m"`)
  - `bunfig.toml` — bun (`minimumReleaseAge = 0`)
- Write `upgrade.log` on install failure with stdout+stderr for debugging

Issue #1260, opened 2026-04-01
`vp upgrade` fails when `minimumReleaseAge` is configured


mizdra:
### Describe the bug

When pnpm's `minimumReleaseAge` option is set globally, the `vp upgrade` command may fail.

### Reproduction

This bug can be reproduced with the `vp upgrade` command alone, so no reproduction repository is provided.

### Steps to reproduce

```bash
# Downgrade previous version
vp upgrade 0.1.14

# Remove the cached version directory
rm -rf ~/.vite-plus/0.1.15

# Set minimumReleaseAge to a large value
pnpm config set --location=global minimumReleaseAge 1000000

# Upgrade to the latest version
vp upgrade
```

Running `vp upgrade` produces the following output and the command fails.

```
info: checking for updates...
info: found vite-plus@0.1.15 (current: 0.1.14)
info: downloading vite-plus@0.1.15 for darwin-arm64...
info: installing...
error: Upgrade error: Failed to install production dependencies (exit code: 1)
```

### System Info

```shell
$ vp env current
VITE+ - The Unified Toolchain for the Web

Environment:
  Version  24.14.1
  Source   lts

Tool Paths:
  node  /Users/mizdra/.vite-plus/js_runtime/node/24.14.1/bin/node
  npm   /Users/mizdra/.vite-plus/js_runtime/node/24.14.1/bin/npm
  npx   /Users/mizdra/.vite-plus/js_runtime/node/24.14.1/bin/npx
```
```shell
$ vp --version
VITE+ - The Unified Toolchain for the Web

vp v0.1.14

Local vite-plus:
  vite-plus  Not found

Tools:
  vite             Not found
  rolldown         Not found
  vitest           Not found
  oxfmt            Not found
  oxlint           Not found
  oxlint-tsgolint  Not found
  tsdown           Not found

Environment:
  Package manager  Not found
  Node.js          v24.14.1
```
```shell
$ pnpm -v
10.28.0
```

### Used Package Manager

pnpm

### Logs

```shell
VITE_LOG=debug vp upgrade
2026-04-01T12:39:18.045254Z DEBUG vp: argv0: vp
2026-04-01T12:39:18.064133Z DEBUG vp::commands::upgrade: Platform: darwin-arm64
info: checking for updates...
2026-04-01T12:39:18.064155Z DEBUG vp::commands::upgrade::registry: Fetching main package metadata: https://registry.npmjs.org/vite-plus/latest
[...]

PR #1338, opened 2026-04-08
fix: more safer Vite+ global install and `vp upgrade`

kazupon:
## Summary

related issues and PRs
- #1260 
- #1272 
- #833 
- #834 

This updates the release-age handling for the Vite+ global install/upgrade path so we no longer silently bypass package manager protections.

Users who configure pnpm `minimumReleaseAge` are explicitly trying to reduce supply-chain risk from newly published compromised packages.

Instead of always writing `minimum-release-age=0`, Vite+ now first runs the wrapper install normally.

If pnpm blocks the install with a release-age error, Vite+ only writes the local override and retries after an interactive, default-No confirmation.

The same behavior is applied to the standalone install scripts, including `install.ps1` and `install.sh`.

## Changes

- Remove unconditional `minimum-release-age=0` override from `vp upgrade` and standalone installers.
- Detect pnpm release-age failures from `ERR_PNPM_NO_MATURE_MATCHING_VERSION`, `minimumReleaseAge` messages, and guarded `ERR_PNPM_NO_MATCHING_VERSION` cases.
- Prompt users with a default-No warning before disabling release-age protection for this Vite+ install only.
- Keep non-interactive environments blocked instead of adding a bypass flag or env var.
- Preserve `install.log` / `upgrade.log` visibility for failure diagnosis.
- Avoid passing `--silent` to the inner captured `vp install`, because pnpm suppresses the release-age error body in silent mode.
- Add comments with pnpm source references explaining the release-age detection signals.


fengmk2 (approved):

liangmiQwQ:
Hey guys, I am handling the package manager refactor (#2406) these days and noticed that. In my opinion, this is quite odd, and we should probably reconsider about that.

First, I want to argue that the reason we call `pnpm` here is not to download a new package, it is install the `vite-plus`, a part of our global cli, vp setup itself. Using `pnpm` and other pm here is just to reuse the functions, as Vite+ is totally able to download the package itself. Most users even don't know there is an internal package manager call, they will feel strange when receiving this prompt.

Second, this doesn't make sense for Vite+ itself's internal installation semantically. Unlike `pnpm install`, who needs `--latest` to upgrade the latest version, `vp upgrade`'s default target is always the newest one, users who run that already decided to use the new Vite+ version.

Third, if we indeed want to add a safety guard, the timing is not right. `vite-plus` package download happens after the corresponding `vp` binary downloads, and the layout is already set. We should either ask users if they want to install the latest version before anything happens; the current approach requires rolling back the installation state later, and the next installation also needs to consider the residual state, which greatly increases the complexity.

kazupon:
Thanks for the comment!

I take the core point to be: this is not a user-project `pnpm install`, yet we overlay the internal pnpm `minimumReleaseAge` gate after the fact, and only prompt after we have already created a partial install.

As long as we use pnpm here, we still have to handle the cooldown. Dropping pnpm and fetching `vite-plus` dependencies ourselves would mean Vite+ also owns dependency resolution and cooldown, which is out of scope for now.

So I think it is enough to fix only the timing of the prompt, rather than removing pnpm or writing our own resolver.

I also agree we should not carry this after-the-fact prompt, retry, and partial rollback into the #2611 self-setup. Asking before setup starts fits that unification better.

liangmiQwQ:
@kazupon Thank you!

I don't think replacing `pnpm` with a custom resolver is the right direction, either. What I mean is that `pnpm` is used here mainly for convenience and performance, and the user is completely unaware of it.

For this question, I suggest directly bypassing pnpm's `minimumReleaseAge` gate, and removing this users prompt.

On the one hand, `vp upgrade` is semantically different from `pnpm install`, which I've mentioned in https://github.com/voidzero-dev/vite-plus/pull/1338#issuecomment-5552112073.

> Second, this doesn't make sense for Vite+ itself's internal installation semantically. Unlike `pnpm install`, who needs `--latest` to upgrade the latest version, `vp upgrade`'s default target is always the newest one, users who run that already decided to use the new Vite+ version.

On the other hand, if we move this prompt before everything starts, we have to read users' settings or set a random number, as `vp` binary itself is downloaded without any package manager. We have to care about `pnpm11+`'s default setting and other edge cases. So `vite-plus` downloading, who will really trigger `minimumReleaseAge`, has still to be bypassed.

Would it be okay?

PR #1348, opened 2026-04-09
rfc: `vp hooks`

ubugeeei:


cpojer:
@ubugeeei My primary question for this RFC is: what is the motivation for adding this? It raises the complexity of Vite+ significantly, and I'm not sure what it allows us to do that you cannot already do with Vite Task.

ubugeeei:
@cpojer 

The motivation was to maintain interfaces like `vp check`, `vp build`, and `vp dev` while incorporating user-specific pipelines.

However, when workflows combine multiple components, it felt counterproductive that users still had to write tasks like `vp run all-check` themselves.

ubugeeei:
Having Vite+'s own plugin system, allowing for extensions using the same interface, seems like a natural motivation.

cpojer:
I think the right "primitive" here is still Vite Task, that is executed via `vp run`. I don't really see how hooks are adding value beyond that.

ubugeeei:
For example, are you suggesting that we should use commands like `sh pre-build.sh && vp dev`, `vp exec cspell && vp check`, `vp run my-build`, or `vp run my-check`? 

I didn’t want to be constantly bothered by inconsistent naming and interfaces, so I thought that an API allowing us to embed custom pipelines into `vp dev` and `vp check` (and others) would solve this problem.

cpojer:
Thanks for explaining! Let's discuss some of these proposals on Discord first before proposing them as RFCs so we make sure we are aligned. For now, we don't plan on adding hooks and we'll be revisiting a potential plugin system in Q3 or Q4 this year.

PR #1360, opened 2026-04-11
feat(js-executor): block vp commands when Node.js version is incompatible

### LINKED ISSUE #1358: JS based tools in vp stop working after setting an incompatible node version with `vp env`
### Describe the bug

According to `js_executor.rs`, almost all JS based vp commands will use `project_runtime` set by `vp env`

So, if users set an incompatible node version out of the scope of `"^20.19.0 || >=22.12.0"` which set in "vite-plus"'s package.json, some vp commands will be unusable because of missing utils, unknown extensions or other issues.

---

I think this is not completely a bug but a design discussion. And I don't know whether is just an expected behavior.

Only vp has the problem because other node version managers do not have JS part so there are no responsibility for them to make sure other js tools run correctly. 

But vp has its own node part and integrated some js tools, so it became hard to decide whether we should ensure other vp tools to run no matter node version is set.

I've tried making fixes for it, I previously thought perhaps a feasible solution is to switch a compatible node versions for vp commands, but I found it really complex as there are commands like `vp test`, `vp run` which runs users' code in codebase, and `vp install` which should not only care vp's node version scope but also respect package manger's. So maybe more discussions are needed on this topic.

Thanks.

### Reproduction

https://github.com/liang-demos/vp-node-version

### Steps to reproduce

https://github.com/liang-demos/vp-node-version#readme

### Validations

liangmiQwQ:
Closes #1358

This PR added check logic to block running / downloading node with unsupported node versions.

The check logic is added in `ensure_project_runtime` fn to make sure all `vp` commands run locally node be checked, not only for Categories C cmds. While direct `node` won't be influenced.

Generated with [Claude Code](https://claude.com/claude-code)


liangmiQwQ:
Done! There are a few test snapshots got updated as they are using node version out of vp's engines scope defined in package.json

fengmk2:
@codex review

crates/vite_global_cli/src/error.rs
```diff
@@ -54,4 +54,9 @@ pub enum Error {
 
     #[error("{0}")]
     Setup(#[from] vite_setup::error::Error),
+
+    #[error(
+        "Node.js {version} does not meet vite-plus requirements ({requirement}).\nRun `vp env use lts` to switch to a compatible version."
```
fengmk2: Run `vp env pin lts` should be better?

fengmk2:
@liangmiQwQ Can we add a snap test for block error reporting? To ensure this functionality remains effective in future changes.

Others LGTM ❤️

liangmiQwQ:
Got it, I would add it soon

crates/vite_global_cli/src/error.rs
liangmiQwQ: umm... perhaps a reason use `vp env use` here is mainly for the situation that user has global default node version and runs commands like `vp create` / `vp migrate` in a no-project directory. (i'm not sure exactly as it is generated by claude)

crates/vite_global_cli/src/js_executor.rs
```diff
@@ -192,17 +204,54 @@ impl JsExecutor {
                 // At least one valid project source exists — delegate to
                 // download_runtime_for_project for cache-aware range resolution
                 // and intra-project fallback chain
-                download_runtime_for_project(project_path).await?
+                let runtime = download_runtime_for_project(project_path).await?;
+                self.check_runtime_compatibility(&runtime.version).await?;
```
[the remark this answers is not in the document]
liangmiQwQ: We can only get the version after running `download_runtime_for_project` as we only have `project_path` as that time.

We need to divide download_runtime_for_project into two funcs for this change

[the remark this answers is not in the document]
fengmk2: no need for now.

crates/vite_global_cli/src/error.rs
fengmk2: I think `vp env pin lts` is the better suggestion for project-local cases, but `vp env use lts` still makes sense in no-project directories for commands like `vp create` or `vp migrate`.

So instead of hardcoding a single remediation, could we make the error message context-aware?

It may also be clearer to say “incompatible with the Vite+ CLI” instead of “does not meet vite-plus requirements”, since this check is based on the CLI’s own `engines.node`, not necessarily the project’s runtime requirement.

If possible, including where the version was resolved from, such as `.node-version` or `VITE_PLUS_NODE_VERSION`, would also make the remediation much easier to understand.

For example:

Project-local case:
```text
Node.js v18.20.0 is incompatible with the Vite+ CLI.
Required by Vite+: ^20.19.0 || >=22.12.0
Resolved from: .node-version

Fix this project: vp env pin lts
Temporary override: vp env use lts
```

No-project case:
```text
Node.js v18.20.0 is incompatible with the Vite+ CLI.
Required by Vite+: ^20.19.0 || >=22.12.0

Use a compatible version for this shell: vp env use lts
```

This keeps the guidance accurate for both cases and makes it clearer that the requirement comes from Vite+ itself.

liangmiQwQ: What about suggesting `vp env default lts` for no project cases?

btw thanks for your explanation

fengmk2: `vp env default lts` is better, let's do it.

fengmk2 (approved):
@liangmiQwQ Thanks again!

fengmk2:
@liangmiQwQ I will roll back this feature and not implement version enforcement validation interception for now to avoid affecting the usage of some older versions. Although the declared older versions are not supported, older versions of Node can actually continue to use Vite+ without issues.

PR #1362, opened 2026-04-12
fix(install): filter KeyEventKind on Windows for arrow key navigation

fengmk2:
On Windows, crossterm emits both Press and Release events for each
key press. Without filtering for Press-only, arrow keys in the
`vp i` package manager prompt were processed twice per press (move
then move back), making them appear non-functional.

Closes #1361

---

> [!NOTE]
> **Low Risk**
> Low risk: small, localized change to interactive terminal input handling, primarily affecting Windows key navigation behavior.
> 
> **Overview**
> Fixes the interactive package manager selection menu to **ignore non-press keyboard events** by filtering `crossterm` `KeyEventKind` to `Press` before handling navigation/selection keys.
> 
> This prevents double-handling of arrow keys on Windows (where both Press and Release events are emitted), making ↑/↓ navigation behave correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a94deea2fc68b5a6a2e2f2868df32b6cc7bf7ec8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1364, opened 2026-04-12
fix(tracing): avoid SetGlobalDefaultError panic when rolldown devtools is enabled

### LINKED ISSUE #1356: SetGlobalDefaultError panic when `@vitejs/devtools` is imported alongside `@voidzero-dev/vite-plus-core`
## Summary

Importing `@vitejs/devtools` in a Vite+ project causes a Rolldown panic at build time:

```
Rolldown panicked. This is a bug in Rolldown, not your code.

thread '<unnamed>' panicked at tracing-subscriber-0.3.23/src/util.rs:94:14:
failed to set global default subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

The panic fires before any plugin runs — simply having `import { DevTools } from "@vitejs/devtools"` at the top of `vite.config.ts` is enough to trigger it, even if `DevTools()` is never added to `plugins`.

## Root cause (as far as I can tell)

`@vitejs/devtools` transitively depends on `@rolldown/debug`, which ships its own Rolldown napi binary. `@voidzero-dev/vite-plus-core` statically bundles Rolldown into a single napi binary (per @fengmk2's note in #1285: _"we packaged rolldown with vite-plus's napi-rs into a single binary"_).

Both binaries load into the same Node process and each tries to install a global `tracing_subscriber`. The second one loses and panics.

This looks like exactly the class of problem discussed in:
- #1285 — panic-hook duplication between Vite+ and bundled Rolldown
- rolldown/rolldown#9023 — gate Rolldown's panic/tracing hook behind a feature flag

…but specifically manifesting through the `@rolldown/debug` → `@vitejs/devtools` dependency chain, rather than through Vite+'s own bundled copy alone.

## Reproduction

1. Fresh project on Vite+ `0.1.16`:
   ```json
   "devDependencies": {
     "@vitejs/devtools": "0.1.13",
     "vite": "npm:@voidzero-dev/vite-plus-core@0.1.16",
     "vite-plus": "0.1.16",
     "vitest": "npm:@voidzero-dev/vite-plus-test@0.1.16"
   },
   "overrides": {
     "vite": "npm:@voidzero-dev/vite-plus-core@0.1.16",
     "vitest": "npm:@voidzero-dev/vite-plus-test@0.1.16"
   }
   ```

2. `vite.config.ts`:
   ```ts
   import { DevTools } from "@vitejs/devtools";
   import { defineConfig } from "vite-plus";

   export default defineConfig({
     plugins: [DevTools()],
     build: {
       rolldownOptions: {
         devtools: {},
       },
     },
   });
   ```

3. `vp build` → panic above.

Commenting out `DevTools()` and `build.rolldownOptions.devtools` but leaving the `import` in place still panics — confirming it's a side-effect of loading the napi binary, not plugin registration.

Fully removing the `import` (and the package) lets the build succeed.

## Dependency tree (relevant slice)

```
@voidzero-dev/vite-plus-core@0.1.16   (bundled rolldown rc.13)
@vitejs/devtools@0.1.13
 └── @vitejs/devtools-rolldown@0.1.13
      └── @rolldown/debug@1.0.0-rc.15   (second rolldown napi)
```

## Environment

- `vp` v0.1.16
- Reported tools: vite v8.0.5, rolldown v1.0.0-rc.13, vitest v4.1.2
- Node.js v24.14.1
- Bun 1.3.11
- macOS (darwin-arm64)
- Framework: React Router 7 framework mode

## Questions

1. Is there a supported path today to use `@vitejs/devtools` with Vite+, or is the expectation that users drop it until the panic-hook gating in rolldown/rolldown#9023 lands and Vite+ consumes it?
2. Would it make sense for `vite-plus-core` to alias/shim `@rolldown/debug` to the bundled Rolldown instance (similar to how `vite` and `vitest` are aliased via npm overrides), so there's only ever one napi binary in the process?
3. Should the DevTools getting-started guide at https://devtools.vite.dev/guide/ note the Vite+ incompatibility until this is fixed?

Happy to test patches / prereleases.

fengmk2:
## Summary

Fixes a panic that occurs when `build.rolldownOptions.devtools` is set in `vite.config.ts` (commonly via `@vitejs/devtools`):

```
Rolldown panicked. This is a bug in Rolldown, not your code.
thread '<unnamed>' panicked at tracing-subscriber-0.3.23/src/util.rs:94:14:
failed to set global default subscriber:
SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

## Root cause

The vite-plus NAPI binary bundles `rolldown_binding` (including `rolldown_devtools`) into a single `.node` file. Two paths inside this same binary each try to install a global tracing subscriber:

1. **NAPI module load** → `vite_shared::init_tracing()` — previously installed a no-op subscriber (empty `Targets` filter) even when `VITE_LOG` was unset, occupying the global default slot.
2. **Build with devtools enabled** → `DebugTracer::init()` — tries to install its own subscriber and panics because the slot is already taken.

## Fix

- Skip subscriber installation entirely when `VITE_LOG` is not set, leaving the global default slot free for `DebugTracer::init()` (the common case — no one pays for tracing they didn't ask for).
- Use `.try_init().ok()` instead of `.init()` so a double-init still fails gracefully rather than panicking.

## Test plan
- [x] Reproduced the panic at `/tmp/repro-1356` with `@vitejs/devtools@0.1.13` + `build.rolldownOptions.devtools: {}`.
- [x] Rebuilt with `RELEASE_BUILD=1 pnpm bootstrap-cli`; `npx vp build` now completes successfully.
- [x] `VITE_LOG=debug vp build` still prints traces (subscriber is installed when requested).

## Related

Upstream rolldown also makes the same `.init()` → `.try_init()` change as defense-in-depth: rolldown/rolldown#9078. That upstream fix is not required for this patch — the vite-plus fix here is sufficient on its own.

Closes #1356

### DIFF
--- a/crates/vite_shared/src/tracing.rs
+++ b/crates/vite_shared/src/tracing.rs
@@ -1,6 +1,6 @@
 //! Tracing initialization for vite-plus
 
-use std::sync::OnceLock;
+use std::{str::FromStr, sync::OnceLock};
 
 use tracing_subscriber::{
     filter::{LevelFilter, Targets},
@@ -14,25 +14,28 @@ use crate::env_vars;
 /// Uses `OnceLock` to ensure tracing is only initialized once,
 /// even if called multiple times.
 ///
+/// Only sets the global default subscriber when `VITE_LOG` is set.
+/// When unset, the global default slot is left free so that other
+/// subscribers (e.g., rolldown devtools) can claim it without panicking.
+///
 /// # Environment Variables
 /// - `VITE_LOG`: Controls log filtering (e.g., "debug", "vite_task=trace")
 pub fn init_tracing() {
     static TRACING: OnceLock<()> = OnceLock::new();
     TRACING.get_or_init(|| {
+        let Ok(env_var) = std::env::var(env_vars::VITE_LOG) else {
+            return;
+        };
+
         tracing_subscriber::registry()
             .with(
-                std::env::var(env_vars::VITE_LOG)
-                    .map_or_else(
-                        |_| Targets::new(),
-                        |env_var| {
-                            use std::str::FromStr;
-                            Targets::from_str(&env_var).unwrap_or_default()
-                        },
-                    )
+                Targets::from_str(&env_var)
+                    .unwrap_or_default()
                     // disable brush-parser tracing
                     .with_targets([("tokenize", LevelFilter::OFF), ("parse", LevelFilter::OFF)]),
             )
             .with(tracing_subscriber::fmt::layer())
-            .init();
+            .try_init()
+            .ok();
     });
 }


fengmk2:
* **#1364** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1364?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1362** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1362?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/voidzero-dev?ref=merge-queue-instructions-in-stack-comment&prId=1364).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

fengmk2:
https://github.com/rolldown/rolldown/pull/9078

hyf0 (approved):

PR #1375, opened 2026-04-14
feat(deps): upgrade upstream dependencies

Brooooooklyn:
Automated daily upgrade of upstream dependencies:
- rolldown (latest tag)
- vite (latest tag)
- vitest (latest npm version)
- tsdown (latest npm version)

Build status: success

---

> [!NOTE]
> **Medium Risk**
> Mostly dependency bumps and snapshot updates, but upgrading `oxlint`/`oxfmt` can subtly change CLI behavior and output (including lint exit conditions), which may affect CI or user workflows.
> 
> **Overview**
> Upgrades upstream tooling versions (notably `oxfmt` `0.45.0`, `oxlint` `1.60.0`, and `tsdown` `0.21.8`, plus related lockfile updates) and aligns `packages/core` peer/bundled versions accordingly.
> 
> Updates CLI snapshot tests to match new outputs/help text (including the new `--no-error-on-unmatched-pattern` option and updated lint/no-files messaging) and applies small lint-suppression comments for generic JSON/YAML/package helpers to satisfy stricter rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8249925af9974473664969eddf55ac4e2742e9fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1378, opened 2026-04-14
fix(cli)!: stop injecting -c config so oxlint/oxfmt pick up nested vite.config.ts

### LINKED ISSUE #997: Vp lint ignores nested package `ignorePattern`
### Describe the bug

The nested vite-plus config in `packages/some-package/vite.config.ts` defines i`gnorePatterns: ["tests/fixtures/**/*"]` to exclude test fixtures from linting. The root vite.config.ts defines lint rules including import/no-commonjs. When vite-plus merges these configs, it applies the root's import/no-commonjs rule to the sub-package but does not respect the sub-package's ignorePatterns. As a result, tests/fixtures/should-ignore-this.cjs gets linted and fails with an import/no-commonjs error, even though it should be ignored.

Expected: Files matching ignorePatterns in the nested config are excluded from linting.

Actual: ignorePatterns from the nested config is not applied (or is overridden during config merging), causing fixture files to be linted against inherited rules.

### Reproduction

https://github.com/rajzik/vite-plus-repro

### Steps to reproduce

1. Create a monorepo with vite-plus and turbo as dev dependencies.
2. Define a root `vite.config.ts` with a lint rule:
```ts
import { defineConfig } from "vite-plus";
export default defineConfig({
  lint: {
    plugins: ["import"],
    rules: {
      "import/no-commonjs": "error",
    },
  },
});
```
3. Create a nested package at `packages/some-package/ `with its own `vite.config.ts` that sets ignorePatterns:
```ts
import { defineConfig } from "vite-plus";
export default defineConfig({
  lint: {
    options: {
      denyWarnings: true,
      reportUnusedDisableDirectives: "error",
      typeAware: true,
      typeCheck: true,
    },
    ignorePatterns: ["tests/fixtures/**/*"],
  },
});
```
4. Add a CommonJS fixture file at `packages/some-package/tests/fixtures/should-ignore-this.cjs` that uses module.exports.
5. Run `pnpm install`.
6. Run `pnpm run lint`.

### System Info

```shell

Environment:
  Version  24.14.0
  Source   lts

Tool Paths:
  node  /Users/jansilhan/.vite-plus/js_runtime/node/24.14.0/bin/node
  npm   /Users/jansilhan/.vite-plus/js_runtime/node/24.14.0/bin/npm
  npx   /Users/jansilhan/.vite-plus/js_runtime/node/24.14.0/bin/npx

vp v0.1.12

Local vite-plus:
  vite-plus  v0.1.12

Tools:
  vite             v8.0.0
  rolldown         v1.0.0-rc.9
  vitest           v4.1.0
  oxfmt            v0.40.0
  oxlint           v1.55.0
  oxlint-tsgolint  v0.17.0
  tsdown           v0.21.3

Environment:
  Node.js          Not found
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

fengmk2:
## Breaking change

`vp lint` and `vp fmt` no longer inject `-c <workspace-root vite.config.ts>` into the underlying oxlint / oxfmt invocation.

Before this PR, every invocation was pinned to the workspace-root config regardless of cwd. That defeated oxlint 1.60 / oxfmt 0.45's new nested-config auto-discovery (gated on `VP_VERSION`, which vp already sets — see oxc-project/oxc#20416 / oxc-project/oxc#20417).

After this PR, config selection is delegated entirely to oxlint / oxfmt:

- Running from a subpackage cwd uses that subpackage's nearest `vite.config.ts` (or `.oxlintrc*` for lint) — nested overrides apply.
- Running from the workspace root uses the root config — but oxlint auto-discovery also matches `.oxlintrc.json` / `.oxlintrc.jsonc` in intermediate directories. If any such file sits between cwd and the file being linted, its config wins over the root's.

### Migration

Repos relying on the old "root config wins everywhere" behavior need one of:
- Remove stray nested `.oxlintrc*` / `vite.config.ts` files from subpackages that shouldn't have their own lint/fmt scope.
- Update nested configs to `extends` the root.
- Restructure so top-level commands always run from the root cwd with no intermediate configs.

Refs oxc-project/oxc#20416


fengmk2:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1378?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1378** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1378?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1378?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1380** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/voidzero-dev?ref=merge-queue-instructions-in-stack-comment&prId=1378).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

fengmk2:
wait for https://github.com/voidzero-dev/vite-plus/pull/1375

fengmk2:
@cursor review

leaysgur (approved):
LGTM for snapshots

cpojer:
Direction makes sense, deferring to @leaysgur for approving the code changes.

fengmk2:
@cpojer @leaysgur Since this change would result in a breaking update, I need more time for testing and validation as well as user documentation. Therefore, I've decided not to include this feature in version 0.1.17 and will release it next week instead.

package.json
```diff
@@ -9,7 +9,7 @@
     "bootstrap-cli:ci": "pnpm install-global-cli",
     "install-global-cli": "tool install-global-cli",
     "tsgo": "tsgo -b tsconfig.json",
-    "lint": "vp lint --type-aware --type-check --threads 4",
+    "check": "vp check --disable-nested-config",
```
fengmk2, line 12: @leaysgur @camc314 can we support disable nested config in `vite.config.ts` too?

leaysgur: If you're asking whether that applies even when using `vite.config.ts` in VP mode, I think the answer is yes.

docs/guide/nested-config.md
```diff
@@ -0,0 +1,103 @@
+# Nested Configuration
+
+Vite+ supports multiple `vite.config.ts` files in the same repository, so packages in a monorepo can have their own lint and format settings while sharing a baseline.
+
+`vp lint` and `vp fmt` resolve configuration from the **current working directory** (cwd), but with one subtle difference:
+
+- **`vp lint`** — cwd-only. Uses `<cwd>/vite.config.ts` if it exists. If not, falls back to Oxlint's built-in defaults — it does **not** walk up to find an ancestor config.
+- **`vp fmt`** — cwd walk-up. Walks up from cwd and uses the first `vite.config.ts` it finds. If none is found anywhere up to the filesystem root, falls back to Oxfmt defaults.
```
fengmk2: @leaysgur Curious why the logic of oxlint and oxfmt is different. 😢

leaysgur: I feel there is no need to write about the internals of oxlint and oxfmt in the user guide, since it's a constraint due to the specific requirements of each tool?

The key points here are:

- 1: Regardless of the cwd, both use the config closest to the file
  - NOTE: oxlint also does finds-up
- 2: When using `-c`, the specified config is used for all files
- 3: However, using `--disable-nested-config` changes how the config is determined based on the cwd:
  - oxlint: refers only to the cwd; if not found, it uses the default
  - oxfmt: finds-up from the cwd; if not found, it uses the default

With regard to 3, the behavior of Oxfmt is aligned with Prettier, dprint, and Biome.

But I don't know about Oxlint, it follows ESLint v9...?

cc: @camc314

Sysix: I want to put it out here, why I think `oxlint` "base" config search should also search upwards the `cwd`, like `oxfmt` does and `oxlint --lsp` ❗ 

Users in a monorepo setup will execute the command in their current package/app. Something like `/project/packages/package oxlint`. With good luck there is already a `.oxlintrc.json` or another config file, which it will detect. 
If not it should imo still find the `.oxlintrc.json` from `/project/` directory.

This was specially problem in LSP with nested [multi root workspaces](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces). 
The issue & fix: https://github.com/oxc-project/oxc/issues/19061 & https://github.com/oxc-project/oxc/pull/19062

I would argue that CLI will hit the same problem when integrated with other tools like turborepo, where the cwd is directly controlled by the user?

camc314: > Curious why the logic of oxlint and oxfmt is different. 😢

@leaysgur I think this goes back to the discussion yesterday - we should align the two!

> I want to put it out here, why I think oxlint "base" config search should also search upwards the cwd, like oxfmt does and oxlint --lsp ❗

Is there an issue for this? lets make one and change the behaviour?

Sysix: @camc314 could not find one, here it is: https://github.com/oxc-project/oxc/issues/21533

camc314: Thanks 🙏

PR #1387, opened 2026-04-15
feat(check): add `--type-check-only` flag

### LINKED ISSUE #1275: Add `vp typecheck` command
### Description

### Suggested solution

### Alternative

### Additional context

  - `vp lint`: runs lint only

### Validations

jong-kyung:
## Summary

Closes #1275

Adds `vp check --type-check-only`, which skips format and lint rules and runs only type checking via oxlint's `--type-check-only` (v1.60.0, upstream oxc-project/oxc#21184). Useful after large-scale changes where fixing type errors first resolves most downstream lint errors.

```bash
vp check --type-check-only
vp check --type-check-only src/
```

## Behavior

- Skips the fmt phase entirely and forwards `--type-check-only` to oxlint in the lint phase.
- Forces type checking regardless of `lint.options.typeCheck` in `vite.config.ts` (oxlint internally enables both `type_aware` and `type_check`).
- New `LintMessageKind::TypeCheckOnly` produces type-specific output ("Found no type errors" / "Type errors found").
- Rejects `--type-check-only --fix` (type diagnostics have no auto-fixer; oxlint also rejects) and `--type-check-only --no-lint` (type checking runs in the lint phase) before spawning any subprocess.
- `--type-check-only --no-fmt` is allowed (redundant no-op).

## Tests

Five snap cases cover happy path, error path, `typeCheck: false` override, and both conflict flags, plus the global `vp check -h` snap is refreshed. Two unit tests cover the new `LintMessageKind` variant.

### DIFF
--- a/crates/vite_global_cli/src/help.rs
+++ b/crates/vite_global_cli/src/help.rs
@@ -760,6 +760,10 @@ fn delegated_help_doc(command: &str) -> Option<HelpDoc> {
                         row("--fix", "Auto-fix format and lint issues"),
                         row("--no-fmt", "Skip format check"),
                         row("--no-lint", "Skip lint check"),
+                        row(
+                            "--type-check-only",
+                            "Run only type checking, skipping format and lint rules",
+                        ),
                         row("-h, --help", "Print help"),
                     ],
                 ),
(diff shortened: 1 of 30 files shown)


jong-kyung:
@codex review

jong-kyung:
@cursor review

jong-kyung:
If there are no issues with this PR, I'll resolve the conflicts in the main branch.

fengmk2:
@jong-kyung please fix conflicts first, I will start review when I have time.

jong-kyung:
Closing in favor of the direction confirmed in #1275 — reusing existing flags

PR #1402, opened 2026-04-17
feat(workflow): generate diff-accurate upgrade-deps PR descriptions via Claude

fengmk2:
Previously the `Upgrade Upstream Dependencies` workflow shipped a generic
template commit message and PR body (see #1401) that didn't reflect what
actually changed. This PR rewires the workflow so the commit and PR
description are generated from the real diff on every run.

### Diff-accurate descriptions
- `.github/scripts/upgrade-deps.mjs` records old → new for every dep it
  touches (including rolldown/vite tag + short SHA) and writes
  `versions.json`, `commit-message.txt`, and `pr-body.md` to
  `$UPGRADE_DEPS_META_DIR` (in `$RUNNER_TEMP`, so they aren't committed).
- `.github/workflows/upgrade-deps.yml`:
  - New `Set up metadata directory` step exports `UPGRADE_DEPS_META_DIR`
    via `$GITHUB_ENV`.
  - New `Enhance PR description with Claude` step reads the baseline
    files plus `git diff` and overwrites them with a Summary, a
    dependency table, a Code-changes list, and a Build-status block.
  - New `Read generated PR content` step exposes the files as multi-line
    step outputs (with a trailing-newline guard so the heredoc
    terminator always lands on its own line).
  - `peter-evans/create-pull-request` now consumes those outputs instead
    of a static template body.
- If the enhancement step fails, `continue-on-error: true` keeps the
  workflow going and the baseline content from the Node script ships
  instead of a generic message.

### Tightened `Check upgrade dependencies` prompt
- Single authoritative checklist: Background → Fixups (in order) →
  Final validation → Commit rule. The previous prompt had three
  overlapping sections describing the same checks.
- Final validation requires BOTH `just build` AND
  `pnpm bootstrap-cli:ci && pnpm test` to pass, plus a manual snap-test
  diff inspection (because `pnpm test` always exits 0 even on snapshot
  drift, so the agent has to look at the diff itself).
- New `Running long commands` rule forbids backgrounding (`&`, `nohup`,
  `disown`, …) and polling (`ps`, `pgrep`, `sleep` loops, repeated `ls`
  on build artifacts). Run 24545325671 spent 30+ minutes spinning in a
  `ps aux | grep "just build"` loop before this rule landed.

### Safety bounds on the Claude session
- `--max-turns 200` caps total tool calls so a runaway agent can't
  consume an entire job budget.
- `timeout-minutes: 180` on the step is a belt-and-suspenders cap on
  wall-clock time.
- Both `anthropics/claude-code-action` pins bumped to v1.0.99
  (Claude Code 2.1.112).

### Script hygiene
- `getLatestTag` uses `?per_page=1` and the rolldown + vite fetches run
  in parallel via `Promise.all`.
- `updatePnpmWorkspace` is now a single-pass `String.replace` callback
  with one capture group per pattern (avoids the foot-gun where the
  replace callback's positional `suffix` arg silently received the match
  offset and corrupted versions). Throws explicitly if a pattern goes
  stale instead of silently recording a no-op change.
- `updateCorePackage` early-exits when `@vitejs/devtools` isn't present,
  skipping the no-op JSON rewrite.

### DIFF
--- a/.github/scripts/upgrade-deps.mjs
+++ b/.github/scripts/upgrade-deps.mjs
@@ -134,6 +134,12 @@ async function updatePnpmWorkspace(versions) {
 
   for (const { name, pattern, replacement, newVersion } of entries) {
     const oldVersion = content.match(pattern)?.[1];
+    if (!oldVersion) {
+      throw new Error(
+        `Failed to match ${name} in pnpm-workspace.yaml — the pattern ${pattern} is stale, ` +
+          `please update it in .github/scripts/upgrade-deps.mjs`,
+      );
+    }
     content = content.replace(pattern, replacement);
     recordChange(name, oldVersion, newVersion);
   }
--- a/.github/scripts/upgrade-deps.mjs
+++ b/.github/scripts/upgrade-deps.mjs
@@ -84,47 +84,67 @@ async function updatePnpmWorkspace(versions) {
   const filePath = path.join(ROOT, 'pnpm-workspace.yaml');
   let content = fs.readFileSync(filePath, 'utf8');
 
-  // The capture regex puts the current version in $1 and matches exactly what
-  // the new string will replace. oxlint's trailing \n disambiguates from oxlint-tsgolint.
+  // Each pattern has exactly one capture group (the current version) so the replace
+  // callback signature is `(match, captured, offset, string)` — no positional ambiguity.
+  // oxlint's trailing \n disambiguates from oxlint-tsgolint.
   const entries = [
     {
       name: 'vitest',
-      pattern: /(vitest-dev: npm:vitest@\^)([\d.]+(?:-[\w.]+)?)/,
+      pattern: /vitest-dev: npm:vitest@\^([\d.]+(?:-[\w.]+)?)/,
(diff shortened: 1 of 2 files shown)


fengmk2:
* **#1402** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1402?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/voidzero-dev?ref=merge-queue-instructions-in-stack-comment&prId=1402).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

fengmk2:
@cursor review

fengmk2:
@cursor review

fengmk2:
ci works https://github.com/voidzero-dev/vite-plus/pull/1405

fengmk2:
@cursor review

Brooooooklyn (approved):

PR #1408, opened 2026-04-17
feat(cli): add oxlint rule to prefer vite-plus imports

### LINKED ISSUE #1301: Enforce certain vite-plus imports via lint rules instead of Markdown files
### Description

I noticed that the generated AGENTS.md/CLAUDE.md file includes this line:

````
- **Import JavaScript modules from `vite-plus`:** Instead of importing from `vite` or `vitest`, all modules should be imported from the project's `vite-plus` dependency. For example, `import { defineConfig } from 'vite-plus';` or `import { expect, test, vi } from 'vite-plus/test';`. You must not install `vitest` to import test utilities.
````

This is fine, but it would be better to enforce this properly - even better if it can get auto-fixed.

### Suggested solution

We could add an `oxc/vite-plus-imports` lint rule in Oxlint to enforce and auto-fix these imports, potentially. And have it enabled by default as part of generating the vp config and/or running `migrate`.

It could alternatively be provided as a JS Plugin rule by vite-plus, if we don't want to ship it _in_ Oxlint directly.

### Alternative

Continue to rely on the AI agents to respect this and probably get it wrong sometimes/a lot of the time, especially while models haven't been trained on vite-plus yet.

### Additional context

Right now this feels like an unnecessarily-painful papercut, because AI agents will continue to just use `vitest` or `vite` imports until, hopefully, they're re-trained on updated info, and a new gen of models is released in 6ish months. And human developers won't necessary know that this is a problem either until they run into issues because of it.

### Validations

### LINKED ISSUE #1457: snap-test-global should model global CLI module resolution for built-in oxlint plugin
### Describe the bug

The global snapshot test harness currently cannot verify generated projects with the default Vite+ lint config intact.

`create-framework-shim-astro` and `create-framework-shim-vue` currently need to delete the generated `lint` config before running `vp check --fix`:

    sed -i.bak -e '/jsPlugins/d' -e '/rules:/d' -e '/options:/d' vite.config.ts

Without this workaround, the global snapshot test cannot run `vp check --fix` with the default generated lint config because oxlint cannot load the built-in Vite+ plugin from:

    jsPlugins: [{ name: "vite-plus", specifier: "vite-plus/oxlint-plugin" }]

This looks like a snapshot harness/environment issue rather than an Astro/Vue product behavior issue. The global snap tests run with `--bin-dir ~/.vite-plus/bin`, while the snap runner also creates temporary `node_modules` symlinks for local package resolution. That means the test environment mixes several module resolution roots:

- the test case temp cwd
- the snap runner temp `node_modules`
- the generated project `node_modules`
- the global CLI layout under `~/.vite-plus`

As a result, `vite-plus/oxlint-plugin` may fail to resolve in `vp check` even though the generated config is valid and should be tested as-is.

I intend to submit a PR for this issue.

### Reproduction

https://github.com/voidzero-dev/vite-plus

### Steps to reproduce

1. Remove the `sed` workaround from `packages/cli/snap-tests-global/create-framework-shim-astro/steps.json`.

2. Optionally remove the same workaround from `packages/cli/snap-tests-global/create-framework-shim-vue/steps.json`.

3. Run the global snap test:

       pnpm -F vite-plus snap-test-global -- --filter=create-framework-shim-astro

4. Observe that `vp check --fix` fails when the generated `vite.config.ts` keeps the default `lint.jsPlugins` entry for `vite-plus/oxlint-plugin`.

### System Info

```shell
N/A for now. This issue tracks the repository's own `snap-test-global` harness behavior.
```

### Used Package Manager

pnpm

### Logs

```shell
he current workaround is visible in:

- `packages/cli/snap-tests-global/create-framework-shim-astro/steps.json`
- `packages/cli/snap-tests-global/create-framework-shim-vue/steps.json`

Both remove these generated config fields before `vp check --fix`:

- `jsPlugins`
- `rules`
- `options`

Suggested fix direction:

The `snap-test-global` harness should create an isolated global Vite+ installation layout that more closely matches the real global CLI layout, instead of relying on the developer/user `~/.vite-plus` state.

In particular:

- isolate `VP_HOME` under the snap test temp directory
- ensure the test global CLI root has a consistent `bin` and `node_modules/vite-plus` layout
- ensure `vite-plus/oxlint-plugin` resolves from the global CLI package when running `vp check`
- remove the `sed` workaround from the framework shim snap tests so the generated default lint config is covered by the snapshot

This would make the product more robust by testing the intended global CLI behavior directly, rather than weakening the generated config during tests.
```

### Validations

Han5991:
## Summary
- add a bundled Oxlint JS plugin rule that rewrites `vite`/`vitest` imports to `vite-plus`
- enable the rule by default in generated and migrated `lint` config
- cover the new wiring with unit tests and snapshot tests

## Testing
- pnpm exec tsc --noEmit --pretty false
- pnpm exec tsc --ignoreConfig --noEmit --pretty false --module esnext --moduleResolution bundler --target es2022 --skipLibCheck packages/cli/snap-tests/lint-vite-plus-imports/vite.config.ts
- local snap tests for `command-init-inline-config` and `lint-vite-plus-imports`

Closes #1301
Closes #1457


packages/cli/src/oxlint-plugin.ts
```diff
@@ -0,0 +1,159 @@
...
+  });
+}
+
+export const preferVitePlusImportsRule = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Prefer vite-plus module specifiers over vite and vitest packages.',
+      recommended: true,
+      url: 'https://github.com/voidzero-dev/vite-plus/issues/1301',
```
fengmk2, line 101: Could we consider adding a new document to replace the issue link? cc @TheAlexLichter

packages/cli/snap-tests/lint-vite-plus-imports/steps.json
```diff
@@ -0,0 +1,8 @@
+{
+  "commands": [
+    "vp lint --fix src/index.ts src/types.ts # rewrite vite and vitest imports via the vite-plus oxlint plugin",
```
fengmk2, line 6: add `vp lint` first to make sure lint fails 😄

fengmk2:
@codex review

packages/cli/src/oxlint-plugin.ts
Han5991: Makes sense to me. I can add a short rule doc and replace the issue URL with that. Do you have a preferred location for it, or should I add it under `docs/`? Also, would you like me to include it in this PR or follow up in a separate PR? If there is already a better link we should use instead, I’m happy to point the rule to that as well.

fengmk2: yes, can follow up in a separate PR.

fengmk2:
Awesome, the import rewrite logic can be completely replaced with the oxlint plugin.

packages/cli/src/oxlint-plugin.ts
Han5991: Yes, we can follow up in a separate PR. If you can point me in the right direction for the docs update, I can take care of it.

Han5991:
> Awesome, the import rewrite logic can be completely replaced with the oxlint plugin.

Makes sense. I focused this PR on the unsafe rewrite fix first, and I can move the remaining import rewrite logic into the oxlint plugin in a follow-up.

packages/cli/src/oxlint-plugin.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+interface ReportContext {
+  report(diagnostic: {
+    data: Record<string, string>;
+    fix: (fixer: Fixer) => unknown;
+    messageId: 'preferVitePlusImports';
+    node: StringLiteralLike;
+  }): void;
+}
```
camc314: can't we use oxlint's real types rather than this?

packages/cli/src/oxlint-plugin.ts
```diff
@@ -0,0 +1,175 @@
...
+      description: 'Prefer vite-plus module specifiers over vite and vitest packages.',
+      recommended: true,
+      url: 'https://github.com/voidzero-dev/vite-plus/issues/1301',
+    },
+    fixable: 'code',
+    messages: {
+      preferVitePlusImports: "Use '{{to}}' instead of '{{from}}' in Vite+ projects.",
+    },
+  },
+  create(context: any) {
```
camc314: it would be ideal to use `createOnce` over create due to its perf advantages

packages/cli/src/oxlint-plugin.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (node.global || !isStringLiteralLike(node.id)) {
+          return;
+        }
+        maybeReportLiteral(context, node.id);
+      },
+    };
+  },
+};
+
+const plugin = {
```
camc314: same, this should probably use `definePlugin` and `defineRule` from oxlint

camc314 (changes requested):
This looks good! but we should definitely use oxlint's types rather than handrolling it, it duplicates code, and might cause issues in future!

packages/cli/src/oxlint-plugin.ts
Han5991: `definePlugin` / `defineRule` don't appear to be publicly available in the `oxlint@1.60.0` version used here, so I don't think we can switch this to those helpers yet.

camc314: Ah we'll need to use https://npmx.dev/package-code/@oxlint/plugins/v/1.60.0/index.js

Han5991: @camc314 
`@oxlint/plugins` isn't installed in this repo right now, so we can't use `eslintCompatPlugin` here yet. If we want to switch to that API, we'd need to add it as a runtime dependency first.

packages/cli/src/oxlint-plugin.ts
fengmk2: BTW: @camc314 We need an oxlint plugin rule to guide us on how to write good oxlint plugins 😃

packages/cli/src/oxlint-plugin.ts
fengmk2: https://viteplus.dev/guide/migrate A section can be added here to describe content related to migrating from vite and vitest, which can include an explanation of this oxlint plugin.

packages/cli/src/oxlint-plugin.ts
camc314: lets add the dependency? any reason why we can't?

Han5991: I asked because I am not yet familiar with this project. If it is okay to add, I can modify it by adding.

packages/cli/src/oxlint-plugin.ts
```diff
@@ -0,0 +1,159 @@
...
+function isStringLiteralLike(value: unknown): value is ESTree.StringLiteral {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    'type' in value &&
+    value.type === 'Literal' &&
+    'value' in value &&
+    typeof value.value === 'string'
+  );
+}
```
camc314: this cleck looks overkill - `value` will be a union of types - i think you just need a `value.type === 'Literal'` check?

Han5991:
@fengmk2 

The failure appears to come from a version skew in the global snap test.

In this test, `vp create` and `vp check` are executed by the locally built Vite+ CLI under test. However, after `vp install`, the generated application installs `vite-plus@latest` from the npm registry into its own `node_modules`.

The local CLI generates a lint config that references `vite-plus/oxlint-plugin`. When oxlint loads that config, it resolves `vite-plus/oxlint-plugin` from the generated application's `node_modules`, not from the local CLI package.

That means oxlint sees the registry-installed `vite-plus@latest`. That published package does not currently expose the new `./oxlint-plugin` subpath, so resolution fails.

This also explains why the other `vite-plus/*` imports appear to work: those subpaths already exist in the published package, while `vite-plus/oxlint-plugin` only exists in the local build being tested.

Han5991:
I think there are three possible ways to address this.

1. Install the local Vite+ artifact in the generated app

The failure is caused by mixing the locally built CLI under test with `vite-plus@latest` from the registry. We could make the generated app install the same local Vite+ artifact set as the CLI under test, for example via a packed local tarball.

This avoids the version skew and keeps the test focused on the current build. A packed tarball seems better than `link:` because it is closer to a real package install and still verifies the published package shape, including `package.json#exports`.

2. Resolve the built-in oxlint plugin from the CLI package

Another option is to make `vp check` resolve Vite+’s built-in oxlint plugin from the CLI package itself before passing the lint config to oxlint.

That would keep the generated app’s registry-installed dependencies intact, but it changes production resolution behavior. I would be careful with this because it may be harder to justify unless built-in Vite+ plugins are explicitly meant to be provided by the running CLI rather than by the project’s installed `vite-plus`.

3. Treat this as a registry/local version skew exposed by the snap

We could also leave the test behavior as-is and consider the failure expected until the package containing `./oxlint-plugin` is published.

This is the most conservative option, but it means the global snap test can fail whenever the local CLI starts generating config that depends on an unreleased `vite-plus` subpath.

packages/cli/src/oxlint-plugin.ts
Han5991: I added `@oxlint/plugins` and tried switching the rule to `definePlugin` / `defineRule`, but this changes the plugin from a self-contained plain object into a plugin module with a runtime dependency.

That seems to be where the regression starts. The snapshot was still passing at `a0e94c9`, where `vite-plus/oxlint-plugin` already existed but the plugin was still a plain object. After switching to `@oxlint/plugins`, the global snapshot fails while loading `vite-plus/oxlint-plugin`.

So I think adding the dependency is probably fine, but using it inside the published JS plugin path needs a bit more care. Since `vite-plus/oxlint-plugin` is loaded by oxlint as a JS plugin from a generated project, we need to make sure oxlint can resolve both the package subpath export and the new runtime dependency in that loading context.

For now, the safest fix may be one of these:

1. Add `@oxlint/plugins` as a dependency, but keep the plugin implementation as a plain object until we confirm oxlint can reliably load helper-based plugins from package subpath exports.
2. Keep the `definePlugin` / `defineRule` change, but add a regression test that runs the global snapshot path and proves `vite-plus/oxlint-plugin` can be loaded from the generated project's installed `vite-plus`.
3. If oxlint expects a specific export condition or module shape for JS plugins, adjust `./oxlint-plugin` exports to match that expectation instead of guessing conditions.

Han5991: @camc314 

Given the regression risk, I think we should split this into a separate PR.

The original change only needed a Vite+ import rule, and the plain-object plugin version was already working at `a0e94c9`. Switching the implementation to `@oxlint/plugins` adds a new runtime dependency and changes how `vite-plus/oxlint-plugin` is loaded by oxlint from generated projects. That makes the scope larger than just adding/fixing the rule.

I suggest we keep this PR focused on the working plain-object implementation, then open a follow-up PR to:

1. add `@oxlint/plugins` as a runtime dependency,
2. migrate the rule to `definePlugin` / `defineRule`,
3. verify the global snapshot path where oxlint loads `vite-plus/oxlint-plugin` from an installed generated project.

That should make the first PR easier to review and give the helper-based migration its own test coverage and discussion.

packages/cli/snap-tests-global/create-framework-shim-astro/steps.json
```diff
@@ -10,6 +10,8 @@
       "command": "cd my-astro-app && vp install -- --no-frozen-lockfile # install dependencies",
       "ignoreOutput": true
     },
-    "cd my-astro-app && vp check --fix # fix generated formatting and ensure no errors"
+    {
+      "command": "cd my-astro-app && sed -i.bak -e '/jsPlugins/d' -e '/rules:/d' -e '/options:/d' vite.config.ts && vp check --fix # fix generated formatting and ensure no errors"
```
Han5991, line 14: Sync oxlint-plugin to core to ensure 'vite-plus/oxlint-plugin' is resolvable when core is aliased as 'vite-plus'.

packages/tools/src/snap-test.ts
```diff
@@ -92,6 +91,214 @@ function selectShard<T>(items: T[], index: number, total: number): T[] {
...
+      (newest, entry) => Math.max(newest, newestMtimeMs(path.join(filePath, entry))),
+      stats.mtimeMs,
+    );
+}
+
+function fileContentsEqual(a: string, b: string): boolean {
+  return fs.readFileSync(a).equals(fs.readFileSync(b));
+}
+
+function assertGlobalCliBinaryMatchesCheckout(binDir: string, casesDir: string): void {
```
fengmk2, line 212: @Han5991 Can we create a new GitHub Action to test the oxlint plugin logic? This way we won't need to test through snap test. Currently, because we have to test through snap test, the modification logic for snap test has become too complex.

Han5991: @fengmk2 

Yes, I think we can add a dedicated GitHub Action/job for the oxlint plugin integration test. That should be a better place to verify that oxlint can load `vite-plus/oxlint-plugin` through package exports and that the rule/fix works, instead of relying on snap-test output.

One note: the snap-test changes here were added because the current global snap-test environment can still execute against the previously installed global `vite-plus` package. In that setup, even if this PR adds the `./oxlint-plugin` export, oxlint may resolve `vite-plus/oxlint-plugin` from the stale installed package instead of the checkout and fail to find the plugin. That is why I forced the global CLI JS entry and relinked installed checkout packages.

I can split this so the oxlint plugin behavior is covered by a focused CI job, and keep only the minimal snap-test fix needed to ensure global snapshots run against the current checkout.

Han5991:
@fengmk2 @camc314

Started splitting plugin verification into a dedicated CI workflow and hit a tradeoff:

- `migration-*` snaps don't run `vp lint`; the diff is just captured lint-config output. Removing the override infra → still pass.
- `create-framework-shim-astro` / `vue` run `vp install` + `vp check`, which actually resolves `vite-plus/oxlint-plugin`. They only pass today because `checkout-packages.json` swaps `vite-plus` for the local pack. Drop that override and they fail until a `vite-plus` version exposing `./oxlint-plugin` is on npm.

Options:
1. Drop the lint step from framework-shim snaps; rely on the new CI for plugin behavior.
2. Gate auto-injection of the plugin into generated configs until the export is published.
3. Keep a minimal `vite-plus` → local-pack override; new CI on top.
4. Land this PR after publishing `vite-plus` with `./oxlint-plugin`, then drop the override entirely.

I'd lean (1). Which do you prefer?

fengmk2:
@Han5991 Please continue with what you think is the most appropriate plan. I hope to complete the first version of the oxlint plugin within the next 2 weeks.

Han5991:
Thanks @fengmk2! Here's my plan:
Merge this PR as-is
Complete the first version within the 2-week window

Han5991:
## Update: musl workaround reapplied with root cause notes

I tried removing the Alpine/musl skip workaround to see what was actually breaking and reproduce it locally, but the failure turned out to be a deeper native binding interaction — not a fixable scope for this PR. So I've reapplied the skip in `dac9da1b` to keep this PR landable.

### What was actually failing

In the Alpine `linux-x64-musl` E2E job, every `vp lint` command in `lint-vite-plus-imports` exits 1 with no output, and `vp check --fix` in `create-framework-shim-vue` exits 1 after only the fmt pass line.

### Root cause

When `oxlint` loads `vite.config.ts` (via the `loadVitePlusConfigs` path triggered by `VP_VERSION`), the dynamic import chain pulls in `vite-plus` → `@voidzero-dev/vite-plus-core` → rolldown's native binding. On musl, **loading the rolldown NAPI binding into the same Node.js process that already has the oxlint NAPI binding loaded crashes with SIGSEGV.**

Reproduced locally in `node:22-alpine3.21`:

| Step | Result |
| --- | --- |
| `oxlint src/index.ts` (no config) | exit 0, 0 errors (plugin not loaded — no error visible) |
| `oxlint -c vite.config.ts` (with `VP_VERSION` set) | **exit 139 (SIGSEGV)** |
| Loading the rolldown binding alone via `node -e` | exit 0 — binding itself is fine |

`strace` confirms `+++ killed by SIGSEGV +++` across the worker pool. SIGSEGV is consistent with the CI symptoms: stderr buffer never flushes before the process dies, so the snap test sees `[1]>` with no body.

### Why it's out of scope for this PR

This is a binary-level interaction between two unrelated NAPI modules (`@oxlint/binding-*` and `@rolldown/binding-*`) on musl, not anything specific to the new `prefer-vite-plus-imports` rule. Likely culprits — needs further investigation:

- Memory allocator conflict (mimalloc / jemalloc statically linked into both)
- musl static TLS slot exhaustion when two large NAPI modules co-load
- Shared transitive Rust crate (e.g. `swc_*`, `oxc_*`) ABI collision

### Suggested follow-up (separate issue)

1. Reproduce on `linux/amd64` to confirm CI parity (I tested on `linux/arm64-musl` locally)
2. Capture a coredump / gdb backtrace to identify the exact symbol crashing
3. Fix upstream (build-flag tweak in oxlint or rolldown) or isolate plugin loading to a subprocess

Happy to file the follow-up issue once this lands.

Han5991:
@fengmk2 @camc314 
Please review

pnpm-workspace.yaml
```diff
@@ -9,8 +9,6 @@ catalog:
   '@babel/preset-env': ^7.24.7
   '@babel/preset-typescript': ^7.24.7
   '@clack/core': ^1.0.0
-  '@emnapi/core': ^1.9.2
-  '@emnapi/runtime': ^1.9.2
```
camc314, line 13: I think this needs to be kept?

packages/tools/src/snap-test.ts
```diff
@@ -450,8 +676,11 @@ async function runTestCase(name: string, tempTmpDir: string, casesDir: string, b
       ]);
 
       await outputStream.close();
+      if (exitCode === 0 && globalCliScriptsDir && steps.linkCheckoutPackages) {
+        replaceInstalledCheckoutPackages(caseTmpDir, resolveRepoRoot(casesDir));
+      }
 
-      let output = readFileSync(outputStreamPath, 'utf-8');
+      let output = fs.readFileSync(outputStreamPath, 'utf-8');
```
camc314, line 683:
```suggestion
      let output = readFileSync(outputStreamPath, 'utf-8');
```

this change looks unnecessary

packages/tools/src/snap-test.ts
```diff
@@ -428,7 +654,7 @@ async function runTestCase(name: string, tempTmpDir: string, casesDir: string, b
       // it seems not to have stable ordering of stdout/stderr chunks.
       // To ensure stable ordering, we redirect outputs to a file instead.
       const outputStreamPath = path.join(caseTmpDir, 'output.log');
-      const outputStream = await open(outputStreamPath, 'w');
+      const outputStream = await fsPromises.open(outputStreamPath, 'w');
```
camc314, line 657:
```suggestion
      const outputStream = await open(outputStreamPath, 'w');
```

packages/cli/src/oxlint-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+import {
+  PREFER_VITE_PLUS_IMPORTS_RULE_NAME,
+  VITE_PLUS_OXLINT_PLUGIN_NAME,
+} from './oxlint-plugin-config.ts';
+
+function isStringLiteralLike(
+  value: ESTree.Expression | ESTree.TSModuleDeclaration['id'],
+): value is ESTree.StringLiteral {
+  return value.type === 'Literal';
+}
```
camc314: this function seems unecessary to me

I also think if you've got a numeric literal, `value.type === 'Literal'` would be true, but it woudn't be a `StringLiteral`

camc314 (approved):
LGTM - the plugin code all looks good to me.

As for the testing infra/bundling. @fengmk2 is probably best placed to review.

Thanks!

pnpm-workspace.yaml
Han5991: These two catalog entries are unreferenced.

fengmk2: In `rolldown/packages/browser/package.json:64-65`, they are directly dependent on it. This can be removed here, as it seems to have been mistakenly introduced by the auto deps action.

packages/tools/src/snap-test.ts
fengmk2: All fs operations have been unified under fs and fsPromises, keeping this change for now.

packages/cli/src/oxlint-plugin.ts
fengmk2: @Han5991 Can you take a look at this review feedback? I'll do manual testing today, and if there are no issues, I'll merge it.

Han5991: @fengmk2 @camc314 

Edit completed!

fengmk2:
@codex review

fengmk2:
@Han5991 The VSCode extension is working normally, thank you for your contribution! Great job!


https://github.com/user-attachments/assets/f57e2155-53ec-4f71-bfaf-42068a13fb04

image

fengmk2 (approved):

Han5991:
> @Han5991 The VSCode extension is working normally, thank you for your contribution! Great job!
> 
>  Screen.Recording.2026-05-16.at.13.15.05.mov 
> image

Wow, it's working really well! That's awesome!

Han5991:
@fengmk2 

It seems like the CI failure is unrelated to the current PR. What should I do?

fengmk2:
> @fengmk2 
> 
> 
> 
> It seems like the CI failure is unrelated to the current PR. What should I do?

Similar to the problem with musl, you can skip it first.

PR #1412, opened 2026-04-18
fix: embed node version requirement at compile time

liangmiQwQ:
There is for an edge case about `vp upgrade`.

If you are using an incompatible node version and you run `vp upgrade`, it will not block your upgrading and install a broken vp cli (Node version outside the scope of `engine.node` would prevent package manager from downloading binary bindings)

Theoretically, the logic added in #1360 should resolve this problem, but as the code below, the node version is read at runtime not compile time.
https://github.com/voidzero-dev/vite-plus/blob/2e2c4c6d6f56a8f42efa59ebbf25d02a0ef41c00/crates/vite_global_cli/src/js_executor.rs#L123-L128

As `vp upgrade` would delegate the cli installing task to the new downloaded vp binary, who has no valid package.json to read, it skips the node version's requirement check and causes this bug.

---

A solution is to inject the node version at compile time as a enhancement for #1360.

This PR simply used `include_str!()` to read package.json and the parsing is still at runtime. It is not the best way for performance. `build.rs` is required if we need even better runtime performance for it.

### DIFF
--- a/crates/vite_global_cli/src/js_executor.rs
+++ b/crates/vite_global_cli/src/js_executor.rs
@@ -13,7 +13,7 @@ use vite_js_runtime::{
 };
 use vite_path::{AbsolutePath, AbsolutePathBuf};
 use vite_shared::{
-    PrependOptions, PrependResult,
+    PackageJson, PrependOptions, PrependResult,
     env_vars::{self, VP_NODE_VERSION},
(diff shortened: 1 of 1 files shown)


liangmiQwQ:
@codex review

fengmk2 (approved):

PR #1417, opened 2026-04-18
feat(cli): add --no-type-check to vp check

### LINKED ISSUE #1275: Add `vp typecheck` command

jong-kyung:
## Summary

Implements the 3-phase skip-flag matrix for `vp check` agreed on in #1275.
Each of format / lint rules / type check can now be skipped independently:

| Flag | Skips |
| ---- | ----- |
| `--no-fmt` | Format check |
| `--no-lint` | Lint rules (type check still runs when enabled in config) |
| `--no-type-check` | Type check |

`vp check --fix --no-lint` on a `typeCheck: true` project is rejected up front to avoid leaving the working tree partially formatted inside pre-commit hooks.

## Request for feedback: `sidecar.rs`

`--no-type-check` is implemented via a per-invocation `.mjs` sidecar that overrides `lint.options.typeCheck`. It leans on oxlint's `loadVitePlusConfigs` route loading `-c <path>` via dynamic `import()`, so we can toggle the flag without mutating `vite.config.ts`.

**This is the part of the PR I'd especially appreciate reviewer input on.** 
One approach I also considered was a wrapper `.mjs` that imports the original `vite.config.ts` and patches `typeCheck` in place — this would preserve non-serializable lint values but complicates relative-path resolution from the temp dir. Happy to switch to that, or to another pattern if the project has a preferred direction for per-invocation config overrides in the Rust binding.

## Related

- Fixes #1275
- Replaces #1387

### DIFF
--- a/packages/cli/binding/src/check/mod.rs
+++ b/packages/cli/binding/src/check/mod.rs
@@ -10,7 +10,8 @@ use vite_task::ExitStatus;
 
 use self::analysis::{
     LintMessageKind, analyze_fmt_check_output, analyze_lint_output, format_count, format_elapsed,
-    print_error_block, print_pass_line, print_stdout_block, print_summary_line,
+    lint_config_type_check_enabled, print_error_block, print_pass_line, print_stdout_block,
+    print_summary_line,
 };
 use crate::cli::{
     CapturedCommandOutput, SubcommandResolver, SynthesizableSubcommand, resolve_and_capture_output,
@@ -22,7 +23,7 @@ pub(crate) async fn execute_check(
     fix: bool,
     no_fmt: bool,
     no_lint: bool,
-    _no_type_check: bool,
+    no_type_check: bool,
     no_error_on_unmatched_pattern: bool,
     paths: Vec<String>,
     envs: &Arc<FxHashMap<Arc<OsStr>, Arc<OsStr>>>,
@@ -46,6 +47,31 @@ pub(crate) async fn execute_check(
     let mut fmt_fix_started: Option<Instant> = None;
     let mut deferred_lint_pass: Option<(String, String)> = None;
     let resolved_vite_config = resolver.resolve_universal_vite_config().await?;
+
+    // Per-phase enabled booleans derived from the raw flags plus the resolved
+    // config's `typeCheck` setting. Currently only the guard immediately below
+    // reads these; the fmt/lint phase conditions further down still consult the
+    // raw `no_fmt`/`no_lint` flags.
+    let config_type_check_enabled =
+        lint_config_type_check_enabled(resolved_vite_config.lint.as_ref());
+    let type_check_enabled = !no_type_check && config_type_check_enabled;
+    let lint_enabled = !no_lint;
+
+    // Reject `--fix --no-lint` when the project enables type-check. With lint
+    // rules skipped, oxlint would take the type-check-only path which it
+    // itself refuses to combine with `--fix`. Running fmt first and then
+    // hitting that rejection would leave the working tree partially formatted
+    // (a real hazard inside lint-staged). Failing up-front keeps the
+    // invocation transactional.
+    if fix && !lint_enabled && type_check_enabled {
+        output::error(
+            "`vp check --fix --no-lint` cannot be combined with type-check enabled in config",
+        );
+        print_summary_line(
+            "type-check diagnostics are read-only and cannot be auto-fixed. Add `--no-type-check` to format-only fix, or drop `--no-lint` to run lint fixes.",
+        );
+        return Ok(ExitStatus(1));
+    }
 
     if !no_fmt {
         let mut args = if fix { vec![] } else { vec!["--check".to_string()] };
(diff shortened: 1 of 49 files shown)


jong-kyung:
@codex review

packages/cli/binding/src/check/sidecar.rs
```diff
@@ -0,0 +1,120 @@
...
+        )));
+    }
+
+    let pid = std::process::id();
+    let nanos = SystemTime::now().duration_since(UNIX_EPOCH).map(|d| d.as_nanos()).unwrap_or(0);
+    let counter = SIDECAR_COUNTER.fetch_add(1, Ordering::Relaxed);
+    let filename = vite_str::format!("vite-plus-no-type-check-{pid}-{nanos}-{counter}.mjs");
+    let filename_str: &str = filename.as_ref();
+    let raw_temp_path = std::env::temp_dir().join(filename_str);
+    let sidecar_path = AbsolutePathBuf::new(raw_temp_path).ok_or_else(|| {
```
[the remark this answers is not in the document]
jong-kyung: > **P1 Badge Keep sidecar config in project path scope**
> 
> This writes the `--no-type-check` override config under the system temp directory, which changes how Oxlint resolves config-file-relative entries. Oxlint resolves `extends` and JS plugin specifiers relative to the `-c` file path (for example, see https://oxc.rs/docs/guide/usage/linter/nested-config and https://oxc.rs/docs/guide/usage/linter/js-plugins), so a project using relative `extends`/`jsPlugins` will break or lint the wrong config when `vp check --no-type-check` points to `/tmp/vite-plus-no-type-check-*.mjs` instead of the project config location.
> 
> Useful? React with 👍 / 👎.

 This path doesn't go through standard oxlint config resolution. With `VP_VERSION` set, [`cli.ts`](https://github.com/oxc-project/oxc/blob/main/apps/oxlint/src-js/cli.ts) uses `loadVitePlusConfigs`, which `import()`s the sidecar and hands `.default.lint` to Rust as a plain JSON object — the sidecar file path isn't a resolution root.

packages/cli/binding/src/check/mod.rs
```diff
@@ -146,10 +200,17 @@ pub(crate) async fn execute_check(
...
+        // before fmt made any changes. Borrow its config here to route oxlint
+        // through the override when present; otherwise use the resolved
+        // config unchanged.
+        let lint_vite_config = sidecar.as_ref().map(|s| &s.config).unwrap_or(&resolved_vite_config);
+
         let captured = resolve_and_capture_output(
             resolver,
             SynthesizableSubcommand::Lint { args },
-            Some(&resolved_vite_config),
+            Some(lint_vite_config),
```
fengmk2, line 213: https://github.com/voidzero-dev/vite-plus/pull/1378 The `-c` parameter specifying oxlint configuration will be removed, so the sidecar configuration approach will have issues. Going forward, the configuration resolution logic for oxlint and oxfmt will no longer be controlled by Vite+, and will be handed back to oxlint/fmt themselves to handle.

cc @leaysgur @camc314 Do you have any better suggestions?

leaysgur: > will be handed back to oxlint/fmt themselves to handle.

I think this is the only option.

Otherwise, Vite+ would end up handling everything, including searching for nested configs.
(That might be a valid decision in itself, but as far as I know, there are currently no plans to do so?)

jong-kyung: > #1378 The `-c` parameter specifying oxlint configuration will be removed, so the sidecar configuration approach will have issues. Going forward, the configuration resolution logic for oxlint and oxfmt will no longer be controlled by Vite+, and will be handed back to oxlint/fmt themselves to handle.
> 
> cc @leaysgur @camc314 Do you have any better suggestions?

Thanks for the heads up on #1378. Since the sidecar + `-c` approach won't survive that change, what would be the recommended direction for implementing `--no-type-check` going forward?

camc314: @fengmk2 what's the use case for the `--no-type-check` flag?

fengmk2: @camc314 Supporting `--no-type-check` is indeed a bit over-engineered. The user's original need is to only perform typeCheck without lint https://github.com/voidzero-dev/vite-plus/issues/1275, which can be achieved through `vp check --no-fmt --no-lint`.

@jong-kyung @cpojer WDYT?

jong-kyung: > @camc314 Supporting `--no-type-check` is indeed a bit over-engineered. The user's original need is to only perform typeCheck without lint #1275, which can be achieved through `vp check --no-fmt --no-lint`.
> 
> @jong-kyung @cpojer WDYT?

@fengmk2 Agreed — `--no-type-check` turned out to be over-engineered.                                                                                                   
Instead, as you suggested, let's make `vp check --no-fmt --no-lint` the entry point for the #1275 use case, and branch on **config.typeCheck: when it's true**,  we'd invoke oxlint with `--type-check-only` (oxc-project/oxc#21184); when false, keep the current hard error. WDYT?

jong-kyung: I'll close this PR and reopen a narrower one along the lines above: drop **--no-type-check/sidecar** entirely, and wire --no-fmt --no-lint through --type-check-only based on the resolved `config.typeCheck`

PR #1422, opened 2026-04-20
fix(cli): prevent TTY hang during command delegation (#1396)

### LINKED ISSUE #1396: vp hangs when spawned by lefthook (opens /dev/tty)
### Describe the bug

`vp` commands hang indefinitely when spawned as child processes by [lefthook](https://github.com/evilmartians/lefthook) git hooks. The same commands complete in ~1 second when run directly in a terminal.

```bash
# Works fine (~1s):
vp run typecheck

# Hangs indefinitely:
git add lefthook.yml && npx lefthook run pre-commit
```

Both `vp run <script>` and `vp test` are affected.

#### Root cause

`lsof` on the hanging `vp` process shows it opens `/dev/tty` directly (fd 9), bypassing stdin:

```
COMMAND   PID    USER   FD   TYPE  DEVICE SIZE/OFF NODE NAME
vp      11794 spalger    0u   CHR   16,0      0t0  895 /dev/ttys000
vp      11794 spalger    1u   CHR   16,0      0t0  895 /dev/ttys000
vp      11794 spalger    2u   CHR   16,0      0t0  895 /dev/ttys000
vp      11794 spalger    9u   CHR    2,0     0t26  335 /dev/tty
```

Lefthook allocates pseudo-TTYs for child processes. When `vp` opens `/dev/tty` and reads from it, it blocks because there's no input on the pseudo-TTY.

This is partly a lefthook issue too — lefthook gives non-interactive child processes a controlling terminal even though it already treats them as non-interactive (null stdin per [evilmartians/lefthook#732](https://github.com/evilmartians/lefthook/pull/732)). But `vp` could also handle the case where `/dev/tty` isn't truly interactive more gracefully, rather than blocking indefinitely.

#### Workaround

Redirect stdin from `/dev/null`:

```yaml
pre-commit:
  commands:
    typecheck:
      run: vp run typecheck </dev/null
    test:
      run: vp test </dev/null
```

### Reproduction

Can't manage to run `curl -fsSL https://vite.plus | bash` in stackblitz sandboxes...

### Steps to reproduce

- Install lefthook npm install --save-dev lefthook
- Write lefthook.yml file

`lefthook.yml`:
```yaml
pre-commit:
  commands:
    typecheck:
      run: vp run typecheck
```

- Run `npx lefthook run pre-commit`

### System Info

```shell

Environment:
  Version  24.14.1
  Source   lts

Tool Paths:
  node  /Users/spalger/.vite-plus/js_runtime/node/24.14.1/bin/node
  npm   /Users/spalger/.vite-plus/js_runtime/node/24.14.1/bin/npm
  npx   /Users/spalger/.vite-plus/js_runtime/node/24.14.1/bin/npx

vp v0.1.16

Local vite-plus:
  vite-plus  Not found

Tools:
  vite             Not found
  rolldown         Not found
  vitest           Not found
  oxfmt            Not found
  oxlint           Not found
  oxlint-tsgolint  Not found
  tsdown           Not found

Environment:
  Package manager  pnpm v10.33.0
  Node.js          v24.14.1
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

Han5991:
Resolves #1396

## Description
Commands like `vp run` or `vp test` running under background/reporter environments (e.g., lefthook with certain Node.js versions) would hang due to `query_terminal_colors()` blocking on `/dev/tty`.

The root cause was `apply_custom_help()` eagerly evaluating the `vite_plus_header()` (and its OSC queries) to build the clap help template during argument parsing, prior to command dispatch.

This commit:
- Removes the eager header evaluation from clap help templates.
- Prints the dynamic `vite_plus_header()` lazily only on `ErrorKind::DisplayHelp`.
- Explicitly passes `dynamic_colors` flags into `print_runtime_header()`, allowing delegate commands to safely use `vite_plus_header_static()` right before execution without blocking on TTY queries.

### DIFF
--- a/packages/cli/snap-tests-global/cli-helper-message/snap.txt
+++ b/packages/cli/snap-tests-global/cli-helper-message/snap.txt
@@ -1,6 +1,5 @@
 > vp -h # show help message
 
-
 Usage: vp [COMMAND]
 
 Start:
--- a/packages/cli/snap-tests-global/cli-helper-message/snap.txt
+++ b/packages/cli/snap-tests-global/cli-helper-message/snap.txt
@@ -1,5 +1,6 @@
 > vp -h # show help message
 
+
 Usage: vp [COMMAND]
 
 Start:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7664,7 +7664,6 @@ name = "vite_shared"
 version = "0.0.0"
 dependencies = [
  "directories",
- "nix 0.30.1",
  "owo-colors",
  "rustls",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6105,9 +6105,9 @@ checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.12"
+version = "0.103.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8279bb85272c9f10811ae6a6c547ff594d6a7f3c6c6b02ee9726d1d0dcfcdd06"
+checksum = "61c429a8649f110dddef65e2a5ad240f747e85f7758a6bccc7e5777bd33f756e"
 dependencies = [
  "ring",
  "rustls-pki-types",
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -1646,7 +1646,7 @@ pub async fn run_command_with_options(
             packages,
             pass_through_args,
         } => {
-            print_runtime_header(render_options.show_header && !silent, true);
+            print_runtime_header(render_options.show_header && !silent);
             // If packages are provided, redirect to Add command
             if let Some(pkgs) = packages
                 && !pkgs.is_empty()
@@ -1972,7 +1972,7 @@ pub async fn run_command_with_options(
             if help::maybe_print_unified_delegate_help("dev", &args, render_options.show_header) {
                 return Ok(ExitStatus::default());
             }
-            print_runtime_header(render_options.show_header, false);
+            print_runtime_header(render_options.show_header);
             commands::delegate::execute(cwd, "dev", &args).await
         }
 
@@ -1980,7 +1980,7 @@ pub async fn run_command_with_options(
             if help::maybe_print_unified_delegate_help("build", &args, render_options.show_header) {
                 return Ok(ExitStatus::default());
             }
-            print_runtime_header(render_options.show_header, false);
+            print_runtime_header(render_options.show_header);
             commands::delegate::execute(cwd, "build", &args).await
         }
 
@@ -1988,7 +1988,7 @@ pub async fn run_command_with_options(
             if help::maybe_print_unified_delegate_help("test", &args, render_options.show_header) {
                 return Ok(ExitStatus::default());
             }
-            print_runtime_header(render_options.show_header, false);
+            print_runtime_header(render_options.show_header);
             commands::delegate::execute(cwd, "test", &args).await
         }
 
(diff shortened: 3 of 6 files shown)


packages/cli/snap-tests-global/cli-helper-message/snap.txt
```diff
@@ -1,6 +1,7 @@
 > vp -h # show help message
 VITE+ - The Unified Toolchain for the Web
 
+
```
fengmk2, line 2: There are extra blank lines here

fengmk2:
@codex review

fengmk2:
@codex review

fengmk2:
@Han5991 The amount of changes currently is a bit too large. Why wasn't the solution mentioned here https://github.com/voidzero-dev/vite-plus/issues/1396#issuecomment-4257247451 considered?

Han5991:
> @Han5991 The amount of changes currently is a bit too large. Why wasn't the solution mentioned here [#1396 (comment)](https://github.com/voidzero-dev/vite-plus/issues/1396#issuecomment-4257247451) considered?

  You're right that the earlier change set was broader than it needed to be.

  I did try the Option 1 approach from #1396 by adding the foreground-process-group check in
  `crates/vite_shared/src/header.rs`, but it did not cover this repro. In the child-PTY setup
  here, lefthook spawns `vp` on a separate TTY, and from `vp`'s perspective it is still
  attached to a terminal and is also the foreground process group for that child TTY. In
  other words, `stdin/stdout` are terminals and `tcgetpgrp(stdin) == getpgrp()`, so the
  process-group guard does not fire even though the OSC `/dev/tty` round-trip can still
  block.

  I also should have evaluated Option 2 from that comment earlier. I did not give that path
  enough weight up front, and instead spent time on CLI-side workaround changes that made the
  diff larger than necessary.

  I've since dropped those CLI-side header workarounds and kept the fix localized to `crates/
  vite_shared/src/header.rs`.

  The current change uses the Option 2-style approach instead: the terminal color query now
  runs in a worker thread, and the main path only waits for a short wall-clock timeout before
  falling back to the default header colors. That way, even if the `/dev/tty` query gets
  stuck under a lefthook PTY, command dispatch is no longer blocked by it.

fengmk2:
@codex review

fengmk2:
@Han5991 Determining terminal color support through `tty` seems too complicated. Is there a more standard solution? How to determine whether a terminal supports colors? cc @cpojer @Brooooooklyn 

I feel that the current fix is not good. Starting a thread to determine terminal color always feels a bit odd.

Han5991:
I think keeping the OSC query with a bounded fallback is still reasonable, but the detached worker-thread timeout is not the right shape.

  The safer approach would be:

  - keep the OSC query as an optional enhancement
  - fall back to the default header colors if the query fails or times out
  - avoid returning while a detached worker may still own `/dev/tty` raw mode
  - make the `/dev/tty` fd non-blocking and keep the polling/read loop bounded by a deadline in the same thread
  - only return after the raw-mode guard has been dropped, so the caller never continues while the terminal may still be in raw mode

  That preserves the intended behavior: use terminal-specific colors when they can be queried safely, otherwise fall back to the default gradient. It also avoids the regression risk introduced by letting command
  dispatch continue while the color-query worker might still be mutating terminal state.

fengmk2:
@Han5991 I will first provide a common method to determine whether color is supported, then you can continue to improve it.

fengmk2:
@Han5991 how about this crate https://github.com/zkat/supports-color

Han5991:
@fengmk2 

Good point. `supports-color` looks like the right standard capability gate here.
We do not actually need to query the terminal palette to decide whether the header can be colorized.
I’ll switch the header to use supports-color for color capability detection and fall back to the default VITE+ gradient, removing the OSC `/dev/tty` query path from normal rendering.

fengmk2:
https://github.com/voidzero-dev/vite-plus/issues/1396#issuecomment-4301258060 close for now

PR #1437, opened 2026-04-22
chore(deps): bump @ast-grep/napi for arm linux

branchv:
On ARM Linux, ast-grep's binary is broken prior to https://github.com/ast-grep/ast-grep/commit/eef144b75764a36057708516657543d42a316e18 and fails with `undefined symbol: le16to` (see https://github.com/ast-grep/ast-grep/issues/2429)


fengmk2 (approved):
@branchv Thanks!

PR #1442, opened 2026-04-22
fix(install): use managed npm for global commands

lyzno1:
## Summary

- Use the resolved managed npm binary for global add/remove/update/outdated commands when the adjacent npm is executable.
- Fall back to npm on PATH when the adjacent managed npm is missing or not executable, with debug logging for diagnostics.
- Keep existing vite_install APIs compatible by falling back to `npm` when no explicit binary is provided.
- Add coverage for absolute binary resolution and managed npm fallback behavior.

## Why

Global npm commands previously depended on PATH being prepared by the caller. This makes the managed npm path explicit when it is safe to do so, while preserving PATH-based fallback for system runtimes and unchanged command arguments.

## Tests

- `cargo test -p vite_command --no-fail-fast`
- `cargo test -p vite_install --no-fail-fast`
- `cargo test -p vite_global_cli --no-fail-fast`

### DIFF
--- a/crates/vite_global_cli/src/commands/add.rs
+++ b/crates/vite_global_cli/src/commands/add.rs
@@ -6,7 +6,7 @@ use vite_install::{
 };
 use vite_path::AbsolutePathBuf;
 
-use super::prepend_js_runtime_to_path_env;
+use super::{managed_npm_bin_for_global_command, prepend_js_runtime_to_path_env};
 use crate::error::Error;
 
 /// Add command for adding packages to dependencies.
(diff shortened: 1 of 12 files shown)


lyzno1:
@codex review

lyzno1:
@codex review

crates/vite_global_cli/src/commands/add.rs
```diff
@@ -53,8 +53,15 @@ impl AddCommand {
 
         // Detect package manager
         let package_manager = PackageManager::builder(&self.cwd).build_with_default().await?;
+        let global_npm_bin_path = managed_npm_bin_for_global_command(global, &node_bin_prefix);
 
-        Ok(package_manager.run_add_command(&add_command_options, &self.cwd).await?)
+        Ok(package_manager
+            .run_add_command_with_global_npm_bin(
```
fengmk2, line 59: Why change this logic? What specific problem does it solve? Could you first submit an issue describing the problem?

lyzno1: The intent is to use the npm that belongs to the Node runtime resolved by Vite+, instead of relying only on PATH. It still falls back to PATH npm when the adjacent npm is missing or not executable.
If this is not the intended direction, I’m happy to close this PR and open an issue first. :)

PR #1444, opened 2026-04-22
feat(check): redefine --no-lint for type-check-only workflow

### LINKED ISSUE #1275: Add `vp typecheck` command

jong-kyung:
## Summary

Redefines `vp check --no-lint` so it runs Oxlint's `--type-check-only` path when
`lint.options.typeCheck` is enabled in `vite.config.ts`, and makes
`vp check --no-fmt --no-lint` the natural type-check-only invocation.

Replaces the abandoned PR #1417 approach (sidecar `.mjs` + `--no-type-check`
CLI flag), which conflicts with upstream PR #1378 handing config resolution
back to oxlint/oxfmt.

Related: closes #1275 for the lint-rule-noise-free type-error triage workflow.

## Behavior matrix

| Command                                         | Behavior                                         |
| ----------------------------------------------- | ------------------------------------------------ |
| `vp check --no-lint` (typeCheck=true)           | Runs `--type-check-only`; lint rules suppressed  |
| `vp check --no-lint` (typeCheck=false/missing)  | Skips lint phase entirely (existing behavior)    |
| `vp check --no-fmt --no-lint` (typeCheck=true)  | Type-check only                                  |
| `vp check --no-fmt --no-lint` (typeCheck=false) | `error: No checks enabled` with remediation hint |
| `vp check --fix --no-lint` (typeCheck=true)     | fmt `--fix` + read-only type-check diagnostics   |

`--no-type-check` CLI flag is **not** introduced — deferred to a follow-up once
oxlint exposes a symmetric OFF-direction flag for type-check.

## Key design notes

- **typeAware prerequisite.** Type-check only runs when both
  `lint.options.typeAware` and `lint.options.typeCheck` are true, matching the
  `--type-check` / `--type-aware` flag dependency in `rfcs/check-command.md`
  ("If `--no-type-aware` is set, `--type-check` is also implicitly disabled").
- **`--fix --no-lint` on a typeCheck-enabled project** runs fmt `--fix` and
  then type-check as read-only diagnostics. On type error, fmt's completion
  pass line is still surfaced so lint-staged / pre-commit flows can tell the
  working tree was mutated before the failure (previously swallowed).
- **No sidecar.** Config resolution is left to oxlint/oxfmt per PR #1378.


jong-kyung:
@codex review

docs/guide/check.md
```diff
@@ -14,9 +14,14 @@ We recommend turning `typeCheck` on so `vp check` becomes the single command for
...
 ```bash
 vp check
-vp check --fix # Format and run autofixers.
+vp check --fix             # Format and run autofixers.
+vp check --no-fmt          # Skip format; run lint (and type-check if enabled).
+vp check --no-lint         # Skip lint rules; keep type-check when enabled.
+vp check --no-fmt --no-lint # Type-check only (requires `typeCheck` enabled).
 ```
 
+When `lint.options.typeCheck` is enabled, `--no-lint` keeps type diagnostics by forwarding Oxlint's `--type-check-only` flag — useful for triaging type errors without lint noise. If `typeCheck` is not enabled, `--no-lint` simply skips the lint phase altogether, and `vp check --no-fmt --no-lint` exits with `No checks enabled` (enable `lint.options.typeCheck` to use the type-check-only invocation).
```
camc314: "by forwarding Oxlint's `--type-check-only` flag — useful for triaging type errors without lint noise." 

This, to me feels like unnecessary info - it doesn't really mater how it works.

Interested to here @ fengmk2 @ cpojer 's thoughts

camc314 (commented):
LGTM 👍

docs/guide/check.md
jong-kyung: > "by forwarding Oxlint's `--type-check-only` flag — useful for triaging type errors without lint noise." 
> 
> This, to me feels like unnecessary info - it doesn't really mater how it works.
> 
> Interested to here @ fengmk2 @ cpojer 's thoughts

Agreed — It's a flag for Vite plus, so I think it would be appropriate to remove it

fengmk2: Let's delete this description, user doc doesn't need to describe the specific implementation.

jong-kyung: > Let's delete this description, user doc doesn't need to describe the specific implementation.

I dropped the `--type-check-only` implementation detail. Thanks! ☺️

fengmk2 (approved):

fengmk2:
@jong-kyung Good job again!

PR #1445, opened 2026-04-22
ci: remove redundant release artifact download

lyzno1:
Remove an unconfigured `download-artifact` step from the release build matrix.

Without `name`, `pattern`, or `path`, the action downloads all current-run artifacts into the workspace. The build job does not consume those artifacts, so this only adds unnecessary API/network/disk work and can leave unrelated files in the checkout.


fengmk2:
@codex review

fengmk2:
I will trigger an alpha test version deployment to verify.

fengmk2 (approved):

PR #1464, opened 2026-04-24
fix(check): remove redundant package tsconfigs to unblock single-file lint-staged

### LINKED ISSUE #1443: bug(check): `vp check --fix <single-file>` spurious TS2591 "Cannot find name 'node:X'" for some files
### Describe the bug

`vp check --fix <path/to/file.ts>` emits spurious `TS2591` errors for `node:*` imports / `process` globals on some TS files when given in isolation, while the same files pass when part of a larger batch or a full-project `vp check --fix`.

This breaks the standard lint-staged pre-commit flow when only a single TS file from the affected directories is staged.

```
x typescript(TS2591): Cannot find name 'node:util'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
   ,-[packages/cli/src/utils/terminal.ts:1:27]
 1 | import { styleText } from 'node:util';
   :                           ^^^^^^^^^^^
```

The root `tsconfig.json` has `"types": ["node"]` and `@types/node` is installed; the import resolves fine at runtime and in the editor.

### Reproduction

On the vite-plus repo itself, current `main`:

```bash
# Fails: TS2591 on the node:util import
vp check --fix packages/cli/src/utils/terminal.ts

# Passes: same file, but as part of a batch of ≥3 files
vp check --fix \
  packages/cli/src/utils/terminal.ts \
  packages/cli/src/version.ts \
  packages/cli/src/config/bin.ts

# Passes: full-project check
vp check --fix
```

### Asymmetry between files

Not every file triggers the bug in isolation. Running `vp check --fix <one-file>` at repo root:

| File                                 | Result |
| ------------------------------------ | ------ |
| `packages/cli/src/utils/terminal.ts` | FAIL   |
| `packages/cli/src/utils/help.ts`     | FAIL   |
| `packages/cli/src/utils/agent.ts`    | FAIL (5 errors) |
| `packages/cli/src/version.ts`        | FAIL (4 errors) |
| `packages/cli/src/bin.ts`            | PASS   |
| `packages/cli/src/config/bin.ts`     | PASS   |
| `packages/cli/src/create/bin.ts`     | PASS   |
| `packages/cli/src/migration/bin.ts`  | PASS   |
| `packages/cli/src/staged/bin.ts`     | PASS   |

All of these files import from `node:*` modules. There is no obvious structural difference from the reporter's side.

`tsgolint --list-files` from the repo root lists every one of the above (both "FAIL" and "PASS" sets) — so the underlying project does include them. The bug appears only when the file is driven into oxlint+tsgolint via a single-file-subset command-line.

### Hypothesis

`vp check --fix <paths>` forwards the paths to `oxlint` which then fans out to `tsgolint`. When a single file subset is passed, `tsgolint` seems to lose the tsconfig's `types: ["node"]` context for some files. The nearest tsconfig to `packages/cli/src/**` is `packages/cli/tsconfig.json`, which has `"files": [], "include": [], "exclude": ["**/*"]` — effectively excluding the whole package. The root `tsconfig.json` (which does provide `types: ["node"]`) is only found if tsgolint walks past the nearest tsconfig when its include set is empty. Whether it walks past seems to depend on which file is passed, which is what makes the behaviour feel like a tsgolint / oxlint-tsgolint project-resolution bug rather than a config issue.

### Impact

The default pre-commit hook runs lint-staged → `vp check --fix <staged-files>`. When a developer stages a single TS file under `packages/cli/src/utils/` (or any of the "FAIL" locations above), the commit is blocked by what is not a real type error.

### Workarounds

- Commit with `--no-verify` (loses all pre-commit validation).
- Stage at least one additional TS file from a "PASS" directory alongside the change.

### Suggested direction

If tsgolint is losing `types: ["node"]` when invoked on a subset of project files, either:

1. In oxlint-tsgolint's integration, always resolve the containing tsconfig chain up to the root before invoking tsgolint (even when explicit files are passed).
2. Or in vp check, if a tsconfig in the resolution chain has `"exclude": ["**/*"]` / empty include, skip past it to the next ancestor tsconfig.
3. Or at minimum, expose a stable workaround users can set in `vite.config.ts` so single-file checks work under lint-staged.

### Steps to reproduce

1. Clone vite-plus and check out current `main`.
2. Build and install the global CLI: `pnpm bootstrap-cli`.
3. From the repo root: `vp check --fix packages/cli/src/utils/terminal.ts`.
4. Observe `TS2591` error on line 1.
5. Re-run: `vp check --fix packages/cli/src/utils/terminal.ts packages/cli/src/config/bin.ts packages/cli/src/create/bin.ts`.
6. Observe the same file passes cleanly.

### System Info

- `vp v0.0.0` (current `main`, commit `aeb40fc4`)
- `oxlint v1.61.0`
- `oxlint-tsgolint v0.21.1`
- Node.js v22.18.0
- macOS arm64 (Darwin 25.2.0)

### Used Package Manager

pnpm

fengmk2:
## Summary

Per @cpojer's [comment](https://github.com/voidzero-dev/vite-plus/issues/1443#issuecomment-4309644867), all three redundant `packages/{cli,core,test}/tsconfig.json` files are removed. They were leftovers from the pre-`oxnode` `tsc` build era — empty-include shells whose only role was setting `experimentalDecorators: true`. The single migration step needed:

- Move `experimentalDecorators: true` to the root `tsconfig.json`.
- For the build path: `@oxc-node/core`'s `OxcTransformer` reads compilerOptions only from `cwd/tsconfig.json` and does not walk up to ancestor tsconfigs (verified empirically + by binary inspection — there's no CLI flag, only the `OXC_TSCONFIG_PATH` env var). Wire it through `cross-env` in `packages/{core,test}` build scripts so oxnode finds the root tsconfig: `cross-env OXC_TSCONFIG_PATH=../../tsconfig.json oxnode -C dev ./build.ts`.
- `packages/core/build.ts:94` also passed `join(projectDir, 'tsconfig.json')` to rolldown's bundler API; repoint that to the root tsconfig too.
- `packages/cli/build.ts` doesn't import `rolldown`, so its build path needs no decorator handling — its tsconfig was the easiest to remove and was the original cause of #1443.

## Why this fixes #1443

`vp check --fix <one-file>` forwards the path to oxlint→tsgolint with no tsconfig. tsgolint walks up from the file looking for the nearest ancestor tsconfig. With the empty-include `packages/{cli,core,test}/tsconfig.json` files in place, that ancestor matched no files and tsgolint built a "loose" program — dropping the parent's `types: ["node"]` and emitting spurious `TS2591: Cannot find name 'node:util'` on `node:*` imports.

After deletion, tsgolint walks all the way up to the project root `tsconfig.json`, which declares `types: ["node"]`, and `@types/node` is loaded as expected. Same fix automatically resolves equivalent failures for files under `packages/core/build-support/` and `packages/{core,test}/__tests__/`.

## Test plan

- [x] `vp check --fix packages/cli/src/utils/{terminal,help,agent}.ts packages/cli/src/version.ts` — was FAIL (TS2591), now PASS
- [x] `vp lint packages/core/build-support/{find-create-require,build-cjs-deps,rewrite-imports,rewrite-module-specifiers}.ts` — all PASS
- [x] `vp lint packages/core/__tests__/build-artifacts.spec.ts` — PASS
- [x] `vp lint packages/test/__tests__/build-artifacts.spec.ts` — was FAIL (TS2591 + TS7006), now PASS
- [x] `vp check` (full project) — PASS (1234 files formatted, 137 lint/type clean)
- [x] `pnpm bootstrap-cli` — PASS

Closes #1443

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -100,7 +100,7 @@
     "./types/internal/*": null
   },
   "scripts": {
-    "build": "oxnode -C dev ./build.ts"
+    "build": "cross-env OXC_TSCONFIG_PATH=../../tsconfig.json oxnode -C dev ./build.ts"
   },
   "dependencies": {
     "@oxc-project/runtime": "catalog:",
@@ -116,6 +116,7 @@
     "@oxc-node/cli": "catalog:",
     "@oxc-node/core": "catalog:",
     "@vitejs/devtools": "^0.1.15",
+    "cross-env": "catalog:",
     "es-module-lexer": "^1.7.0",
     "hookable": "^6.0.1",
     "magic-string": "^0.30.21",
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -1,9 +1,0 @@
-{
-  "extends": "../../tsconfig.json",
-  "compilerOptions": {
-    "experimentalDecorators": true
-  },
-  "files": [],
-  "include": [],
-  "exclude": ["**/*"]
-}
--- a/packages/core/build.ts
+++ b/packages/core/build.ts
@@ -91,7 +91,7 @@ await mergePackageJson();
 
 async function buildVite() {
   const newViteRolldownConfig = viteRolldownConfig.map((config) => {
-    config.tsconfig = join(projectDir, 'tsconfig.json');
+    config.tsconfig = join(projectDir, '..', '..', 'tsconfig.json');
     config.cwd = projectDir;
 
     if (Array.isArray(config.external)) {
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -1,9 +1,0 @@
-{
-  "extends": "../../tsconfig.json",
-  "compilerOptions": {
-    "experimentalDecorators": true
-  },
-  "files": [],
-  "include": [],
-  "exclude": ["**/*"]
-}
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -1,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "experimentalDecorators": true
+  },
+  "files": [],
+  "include": [],
+  "exclude": ["**/*"]
+}
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "experimentalDecorators": true
   },
-  "files": [],
-  "include": [],
-  "exclude": ["**/*"]
+  "include": ["build-support/**/*", "__tests__/**/*"],
+  "exclude": ["dist", "node_modules"]
 }
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -1,8 +1,0 @@
-{
-  "extends": "../../tsconfig.json",
-  "compilerOptions": {
-    "experimentalDecorators": true
-  },
-  "include": ["build-support/**/*", "__tests__/**/*"],
-  "exclude": ["dist", "node_modules"]
-}
--- a/packages/test/package.json
+++ b/packages/test/package.json
@@ -272,7 +272,7 @@
     }
   },
   "scripts": {
-    "build": "oxnode -C dev ./build.ts"
+    "build": "cross-env OXC_TSCONFIG_PATH=../../tsconfig.json oxnode -C dev ./build.ts"
   },
   "dependencies": {
     "@standard-schema/spec": "^1.1.0",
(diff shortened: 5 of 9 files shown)


packages/core/package.json
```diff
@@ -100,7 +100,7 @@
     "./types/internal/*": null
   },
   "scripts": {
-    "build": "oxnode -C dev ./build.ts"
+    "build": "cross-env OXC_TSCONFIG_PATH=../../tsconfig.json oxnode -C dev ./build.ts"
```
fengmk2, line 103: https://github.com/oxc-project/oxc-node/pull/569 cc @Brooooooklyn

camc314:
@fengmk2 you should be able to get away without this, going to do a tsgolint release later today.

fengmk2:
> @fengmk2 you should be able to get away without this, going to do a tsgolint release later today.

@camc314 Even after upgrading tsgolint, the current configuration in `packages/cli/tsconfig.json`/`packages/core/tsconfig.json` looks very strange. Shouldn't them be removed as well?

camc314:
@fengmk2 yes, i think so - or maybe they should he a their "includes" set correctly.


I didn't manage to get a release out yesterday - got stuck on one last tricky issue to fix - apologies.

cpojer (approved):

PR #1467, opened 2026-04-24
fix(install): avoid truncating lock file to prevent sharing violation on Windows

### LINKED ISSUE #1465: Concurrency issue: Running vp install in parallel across multiple directories fails on Windows
When attempting to run `vp install` concurrently in multiple subdirectories, the command frequently fails with an exit code of `1`. This is commonly encountered in monorepos or when using build systems like MSBuild that attempt to build multiple frontend projects in parallel. 

**Steps to Reproduce:**
1. Create multiple frontend directories (e.g. `frontend-a`, `frontend-b`, `frontend-c`) initialized with `vite-plus`.
2. On a Windows machine, trigger `vp install` inside all of these directories at the exact same time (for example, via parallel MSBuild targets or using a tool like `concurrently`). 
3. Observe that one or more of the instances will crash/fail, returning exit code 1.

**Expected Behavior:**
`vp install` should be safe to execute concurrently in multiple directories, gracefully handling any internal `npm`/`pnpm` cache locking or global store access without crashing.

**Current Workaround:**
We are currently wrapping `vp install` and `vp build` in a global OS-level Mutex during our build process to force sequential execution and avoid the race condition.

**Environment:**
* OS: Windows
* Package Manager: pnpm (via `vp` wrapper)
* Vite Plus version: 0.1.19

**Proposed fix**

To fix this cross-platform, vite-plus needs to open the lock file without truncating it, using `OpenOptions`:

```rust
use std::fs::OpenOptions;

let lock_path = parent_dir.join(format!("{version}.lock"));
let lock_file = OpenOptions::new()
    .read(true)
    .write(true)
    .create(true)
    .truncate(false) // CRITICAL: Do not truncate, which conflicts with locked files on Windows
    .open(lock_path.as_path())?;

lock_file.lock()?;
```

rorychatt:
Fixes #1465


fengmk2:
@codex review

fengmk2:
@rorychatt Could you add a corresponding test? Unit tests or snap tests are both fine.

rorychatt:
@fengmk2 sure. I will get back to you in the evening.

rorychatt:
To get absolute proof, I wrote a dedicated cross-process test script locally that spawns a child process to exclusively lock the file, while the parent process attempts to open it simultaneously

When the parent process used the old File::create logic, it crashed with ERROR_SHARING_VIOLATION, perfectly reproducing the bug.

When I switched the parent process to use our extracted OpenOptions fix, it cleanly opened the handle and waited in line for the lock without any errors, proving the fix works flawlessly

Because standard cargo test runs in a single process, the unit test I added serves to verify the syntactic correctness of the extracted open_lock_file function and ensures it successfully acquires and releases the lock

@fengmk2

rorychatt:
@fengmk2 Sorry, I removed the unused import. Linting should pass now

fengmk2 (approved):

fengmk2:
@codex review

fengmk2 (approved):
@rorychatt Thanks!

PR #1468, opened 2026-04-25
ci(upgrade-deps): teach autofix agent to handle sync-remote failures

fengmk2:
## Summary

The daily `Upgrade Upstream Dependencies` workflow has a `Sync remote and build` step running `pnpm tool sync-remote` under `continue-on-error: true`, with the downstream `Check upgrade dependencies` Claude Code step intended to repair anything that breaks. The prompt for that autofix step never mentioned sync-remote, however, so when the catalog merger errored on a cross-major version conflict (e.g. rolldown bumped `diff` to `^9` while vp's catalog still pinned `^8`), the failure was silently swallowed — yesterday's PR #1460 was opened with `sync-remote-and-build: failure` recorded in the body, but no actual fix attempted.

## What changed

`.github/workflows/upgrade-deps.yml` only — the prompt of the `Check upgrade dependencies` step now:

- Surfaces `${{ steps.build.outcome }}` and `${{ steps.build-upstream.outcome }}` so the agent can see that sync-remote already failed.
- Adds an explicit Step 1 fixup describing the typical `Incompatible semver ranges for <pkg>: ^X vs ^Y` failure mode and how to resolve it — bump the passthrough catalog entry in `pnpm-workspace.yaml` to match the rolldown/vite version, then re-run `pnpm tool sync-remote` until it exits 0.

The existing fixup steps (build-upstream rerun, cargo workspace merge, tsdown CLI sync, lint, smoke tests) are renumbered 2–8 but otherwise unchanged.

## Test plan

- [x] Manually trigger the workflow on this branch via `workflow_dispatch`.
- [x] Confirm `Sync remote and build` still fails with the existing `Incompatible semver ranges for diff` error (this branch intentionally does NOT pre-bump `diff: ^8.0.0 -> ^9.0.0` in `pnpm-workspace.yaml`, so the failure is reproduced).
- [x] Confirm the `Check upgrade dependencies` step bumps the catalog entry, re-runs sync-remote until it exits 0, and the rest of the validation pipeline passes.

### DIFF
--- a/.github/workflows/upgrade-deps.yml
+++ b/.github/workflows/upgrade-deps.yml
@@ -74,30 +74,51 @@ jobs:
           show_full_output: 'true'
           prompt: |
             Your goal: after the daily upstream-dependency upgrade, bring the project back
-            to a fully green state. The upgrade script has already bumped every dep to the
-            latest version and the `build-upstream` action has attempted a build — your job
-            is to diagnose and fix every error that surfaced, then prove the fix is complete
-            by running a final validation pass.
+            to a fully green state. The upgrade script has bumped every dep to the latest
+            version, the `Sync remote and build` step has attempted to merge the rolldown
+            and vite catalogs into the workspace, and the `build-upstream` action has
+            attempted a build — your job is to diagnose and fix every error that surfaced,
+            then prove the fix is complete by running a final validation pass.
 
             ### Background
             - Upgrade script: `./.github/scripts/upgrade-deps.mjs`
+            - Sync-remote tool: `pnpm tool sync-remote` (source in
+              `packages/tools/src/sync-remote-deps.ts`) — clones rolldown/vite into the
+              working tree and merges their pnpm-workspace catalogs into the root
+              `pnpm-workspace.yaml`.
             - Build-upstream action: `./.github/actions/build-upstream/action.yml`
             - Package manager: `pnpm`. Do NOT downgrade any dep — we want the latest.
 
+            ### Outcomes from earlier steps (both ran with continue-on-error)
+            - Sync remote and build: ${{ steps.build.outcome }}
+            - build-upstream action: ${{ steps.build-upstream.outcome }}
+
             ### Fixups to perform (in order)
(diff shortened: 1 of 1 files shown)


fengmk2:
test https://github.com/voidzero-dev/vite-plus/actions/runs/24920197224

.github/workflows/upgrade-deps.yml
```diff
@@ -74,30 +74,51 @@ jobs:
...
+            - Sync-remote tool: `pnpm tool sync-remote` (source in
+              `packages/tools/src/sync-remote-deps.ts`) — clones rolldown/vite into the
+              working tree and merges their pnpm-workspace catalogs into the root
+              `pnpm-workspace.yaml`.
             - Build-upstream action: `./.github/actions/build-upstream/action.yml`
             - Package manager: `pnpm`. Do NOT downgrade any dep — we want the latest.
 
+            ### Outcomes from earlier steps (both ran with continue-on-error)
+            - Sync remote and build: ${{ steps.build.outcome }}
+            - build-upstream action: ${{ steps.build-upstream.outcome }}
```
fengmk2, line 94: Will the `steps.build-upstream.outcome` output be too large?

fengmk2: Can Outcomes be omitted as a prompt input?

fengmk2: `steps.<id>.outcome` is just a single-word status string (`success` / `failure` / `cancelled` / `skipped`), so size isn't an issue — but the block was redundant either way and is dropped in ed8ab9c1.

fengmk2: Yes — dropped in ed8ab9c1. The agent verifies sync-remote's state directly by re-running it (Step 1 always runs it now), so passing the workflow outcome was a redundant second source of truth.

fengmk2:
works https://github.com/voidzero-dev/vite-plus/pull/1469

cpojer (approved):

PR #1471, opened 2026-04-25
feat(cli): error on `vp install -g` with no package names

liangmiQwQ:
## Summary

- `vp install -g` with no package names silently fell through to a regular workspace install, ignoring `--global` entirely
- Added a guard that emits an actionable error and exits with code 1 when `--global` is set but no packages are provided
- Added a global snap test to cover this case

## Root cause

In the `Install` command dispatcher (`cli.rs`), the `global` flag was only inspected inside the `!pkgs.is_empty()` branch. When no packages were given, the code skipped straight to `InstallCommand::new(cwd).execute(...)`, making `--global` a no-op.

## Fix

Insert a guard immediately before the workspace install fallback:

```rust
if global {
    vite_shared::output::error(
        "`vp install -g` requires at least one package name.\n\
         To reinstall all globally installed packages, use `vp update -g`.",
    );
    return Ok(exit_status(1));
}
```

## Test plan

- [x] `cargo build -p vite_global_cli` compiles cleanly
- [x] `vp check --fix` passes (format + lint)
- [x] New snap test `command-install-g-no-args` verifies exit code 1 and error message
- [ ] Snap test suite passes in CI

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -19,7 +19,7 @@ catalogs:
       specifier: ^3.6.1
       version: 3.6.1
     '@napi-rs/wasm-runtime':
-      specifier: ^1.1.3
+      specifier: ^1.1.4
       version: 1.1.4
     '@nkzw/safe-word-list':
       specifier: ^3.1.0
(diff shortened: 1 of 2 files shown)


liangmiQwQ:
Close as committed to a wrong branch by Claude

PR #1472, opened 2026-04-25
feat(cli): error on `vp install -g` without package names

liangmiQwQ:
Currently, running `vp install -g` with no package names will silently fall through to a regular workspace install, ignoring `--global` entirely, as the `global` check logic will only happen after checking package name is provided.

This PR adjusted clap config to make sure package is provided when using `--global` flag, also adjusted help output for more clear guidance.

Global snap test is also added to cover this case.

Generated with [Claude Code](https://claude.com/claude-code) and Codex


liangmiQwQ:
@codex review

fengmk2 (approved):
@liangmiQwQ Thanks!

Issue #1477, opened 2026-04-25
Skip packages which are up to date in `vp update -g`


liangmiQwQ:
### Description

If we run `vp update -g`, the current Vite+ will reinstall all the installed packages, no matter whether it is up to date, which makes update process slow.

### Suggested solution

We can follow other pm's behavior like pnpm / npm, which needs Vite+ compares user's installed version with the latest version on registry, and only reinstall ones which is needed to update. We can't delegate it to `npm install -g` like #1476 as we use different prefixes for different packages.

Since vp can record the Node version used to install the package, whether to reinstall when node version changes may become a topic need to discuss.

### Additional context

Maybe it is partially related #1476, implementing this after finishing #1476 can make work much simpler

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

PR #524, opened 2026-01-31
feat(env): implement vp env command for Node.js version management

fengmk2:
Add shim-based Node.js version management with the following features:

- `vp env setup`: Create shims (node, npm, npx) in ~/.vite-plus/bin
- `vp env doctor`: Diagnostics for environment configuration
- `vp env default [VERSION]`: Set/show global default Node.js version
- `vp env which <TOOL>`: Show path to tool binary
- `vp env --current [--json]`: Show current environment info
- `vp env --print`: Print shell snippet for session

The shims intercept node/npm/npx commands and automatically resolve
the correct Node.js version based on:
1. .node-version file
2. package.json#engines.node
3. package.json#devEngines.runtime
4. User default from config
5. Latest LTS fallback

Implementation includes:
- Shim detection via argv[0] or VITE_PLUS_SHIM_TOOL env var
- Resolution cache with mtime validation for fast repeated invocations
- Platform-specific execution (execve on Unix, spawn on Windows)
- Conflict detection for other version managers (nvm, fnm, volta, etc.)
- Updated install.sh with shim setup and PATH configuration prompt

PR #1478, opened 2026-04-25
chore: remove dead global logic for update command

liangmiQwQ:
There is a part of unreachable code in `update.rs`. It is used to run `npm update -g` when running `vp update -g`.

It was added in #238, the first implementation of `vp update`. But it became dead code after #524 got merged (the PR which introduces `vp env`).

The following code in `cli.rs` makes it unreachable.

https://github.com/voidzero-dev/vite-plus/blob/a2876b6a3ef096c9b5b1ab79aced73dc72d0c8ce/crates/vite_global_cli/src/cli.rs#L1816-L1817

This PR clears the logic about it, including related tests

It was found by Claude Code while I was writing #1477.

### DIFF
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -1840,7 +1840,6 @@ pub async fn run_command_with_options(
                 .execute(
                     &packages,
                     latest,
-                    global,
                     recursive,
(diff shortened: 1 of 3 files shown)


fengmk2 (approved):

fengmk2:
@codex review

PR #1481, opened 2026-04-26
fix(env): support PowerShell on non-Windows platforms

nekomoyi:
PowerShell is not Windows-only, but `vp env setup` was generating env.ps1 with vp.exe. That breaks pwsh users on non-Windows platforms.

Changes
- Replace vp.exe with vp in the generated env.ps1
- Add PowerShell coverage in CLI E2E
- Explicitly identify PowerShell wrapper execution via `VP_SHELL_PWSH`


fengmk2:
@codex review

fengmk2 (approved):

fengmk2:
@nekomoyi Could you add a corresponding test job in our CI? I'm concerned that future code changes might still modify this code without discovering this special logic.

nekomoyi:
@fengmk2 Thanks. I’ve updated the PowerShell coverage in CI.

The PowerShell path in our CLI E2E test can now run cross-platform. PowerShell steps go through the shell env wrapper instead of calling the `vp` binary directly. At the moment, the PowerShell steps are still skipped on `namespace-profile-linux-x64-default` because that runner does not have `pwsh` preinstalled.

I also adjusted `detect_shell()` so that PowerShell wrapper execution is identified by `VP_SHELL_PWSH` instead of `PsModulePath` and `cfg!(windows)`.

This fixes the immediate non-Windows pwsh issue, but `detect_shell()` can still be confused by inherited variables like `FISH_VERSION` in nested shell scenarios. I’m leaving that for a follow-up with a cleaner approach.

fengmk2 (approved):

PR #1491, opened 2026-04-27
feat(vscode): add language-specific formatter overrides to `.vscode/settings.json`

### LINKED ISSUE #1485: [Enhancement] Generate language-specific overrides in .vscode/settings.json
### **The Problem**
When generating configurations for **VS Code**, the current template only outputs a global `editor.defaultFormatter`. 

Because of VS Code's **Settings Precedence**, any language-specific configuration in a user's global `settings.json` will take priority over a workspace-level global setting. This means if a user has Prettier set specifically for TypeScript in their global settings, the generated workspace setting for `oxc-vscode` is ignored.

### **Precedence Conflict**
1. **User Settings (Global - High Priority):**
   ```json
   "[typescript]": { "editor.defaultFormatter": "esbenp.prettier-vscode" }
   ```
2. **Generated Workspace Settings (Low Priority):**
   ```json
   "editor.defaultFormatter": "oxc.oxc-vscode"
   ```
*Result: VS Code uses Prettier, ignoring the project's intent to use OXC.*

### Suggested solution

When the VS Code configuration is generated, the tool should include language-specific blocks based on the project template (e.g., `[typescript]`, `[typescriptreact]`, `[javascript]`). This ensures the workspace setting correctly overrides any existing user-level language settings.

**Target Output for `.vscode/settings.json` for typescript react project:**
```json
{
  "editor.defaultFormatter": "oxc.oxc-vscode",
  "[typescript]": {
    "editor.defaultFormatter": "oxc.oxc-vscode"
  },
  "[typescriptreact]": {
    "editor.defaultFormatter": "oxc.oxc-vscode"
  }
}
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

oilater:
Closes #1485

## Summary

- Re-assert `editor.defaultFormatter: oxc.oxc-vscode` inside `[javascript]`/`[javascriptreact]`/`[typescript]`/`[typescriptreact]` blocks in the generated `.vscode/settings.json`. Without these, a user-level `[language]` setting (e.g. a global Prettier override) silently wins over the workspace-level default.
- Extract the override blocks into a dedicated `VSCODE_LANGUAGE_OVERRIDES` constant so `VSCODE_SETTINGS` keeps a single concern.
- Update `docs/guide/ide-integration.md` to reflect the new shape and explain why per-language blocks are required.
- Existing `deepMerge` behavior is unchanged, so a user's pre-existing `[typescript]: prettier` block is still preserved during `vp create` against an existing `.vscode/settings.json`.

## Approach

The main question was whether to map `[lang]` blocks per project template or to always include the JS/TS family. Picked the latter because

- `vp create` accepts remote templates (`github:user/repo`, arbitrary `create-*` packages) where the language set can't be known up front — a per-template branch ends in a JS/TS fallback for these anyway, so the per-template path doesn't actually buy correctness for the cases it's meant to help.
- Keeps `vp create` and `vp migrate` consistent with a single source of truth in `VSCODE_SETTINGS`.
- Including unused `[lang]` blocks has no runtime effect (VS Code only applies them when matching files are opened) and is evolution-safe if the user later adds a `.ts` file to an initially JS-only project.

## Tests

- `pnpm -F vite-plus test src/utils/__tests__/editor.spec.ts` — extended assertions verify the four `[lang]` blocks are written, and that an existing user's `[typescript]: prettier` is preserved during merge.

### DIFF
--- a/packages/cli/src/utils/__tests__/editor.spec.ts
+++ b/packages/cli/src/utils/__tests__/editor.spec.ts
@@ -111,6 +111,9 @@ describe('writeEditorConfigs', () => {
     expect(settings['oxc.fmt.configPath']).toBe('./vite.config.ts');
     expect(settings['editor.formatOnSave']).toBe(true);
     expect(settings['npm.scriptRunner']).toBeUndefined();
+    for (const lang of ['[javascript]', '[javascriptreact]', '[typescript]', '[typescriptreact]']) {
+      expect(settings[lang]).toEqual({ 'editor.defaultFormatter': 'oxc.oxc-vscode' });
+    }
   });
 
   it('includes additionalSettings in vscode settings.json when provided', async () => {
@@ -142,6 +145,9 @@ describe('writeEditorConfigs', () => {
       `{
   // JSONC comment
   "editor.formatOnSave": false,
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+  },
   "editor.codeActionsOnSave": {
     // preserve existing key
     "source.organizeImports": "explicit",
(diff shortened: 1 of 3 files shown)


jong-kyung:
@codex review

packages/cli/src/utils/__tests__/editor.spec.ts
```diff
@@ -142,6 +145,9 @@ describe('writeEditorConfigs', () => {
       `{
   // JSONC comment
   "editor.formatOnSave": false,
+  "[typescript]": {
```
jong-kyung, line 148: Could we extend this merge test to also assert insertion of the three non-[typescript] blocks ([javascript], [javascriptreact], [typescriptreact])?

oilater: Done in [b16041f](https://github.com/voidzero-dev/vite-plus/pull/1491/commits/b16041f03bbfcabb1795d4c3edd10dffdc62b972)! extended the merge test to cover all three non-[typescript] blocks.

jong-kyung (approved):

jong-kyung:
@fengmk2 Could you re-run the CI? It seems to be a network issue on setup-node.

fengmk2 (approved):

fengmk2:
@oilater @jong-kyung Thanks!!

PR #1508, opened 2026-05-02
fix(cli): show create and migrate in local cli top-level help

nurazon59:
## Why
`vp -h` was missing `create` and `migrate`.

## What
Added both commands to the top-level help output and updated the affected CLI snapshots.

## Checks
Local snap-test harness could not run because `tool` is not installed in this environment. `cargo test -p vite_global_cli --bin vp` also stopped early because the workspace references a missing `rolldown_binding` manifest in this checkout.


fengmk2 (approved):
@nurazon59 Thanks!

nurazon59:
@fengmk2 Thank you for the review. I found a bug in the tool and am fixing it in https://github.com/voidzero-dev/vite-plus/pull/1524. Could you review that PR? Once it’s merged, please review this again.

Issue #1511, opened 2026-05-04
docs: update stats in home page


mdong1909:
### Documentation is

- [ ] Missing
- [x] Outdated
- [ ] Confusing
- [ ] Not sure

### Explain in Detail

Following stats are outdated:
1. eslint rules 600+ > 750+
2. Npm downloads
3. Github stars

<img width="435" height="271" alt="Image" src="https://github.com/user-attachments/assets/0ddbbc39-52c4-4261-996d-d7f0dbd51f32" />

<img width="857" height="449" alt="Image" src="https://github.com/user-attachments/assets/13adc51b-3e49-4eb0-a5ad-2bca6325b204" />

### Your Suggestion for Changes

1. manual updates - https://github.com/voidzero-dev/vite-plus/blob/main/docs/.vitepress/theme/components/home/FeatureCheck.vue
2 + 3. create an automation as a weekly update

nozomee:
Hi! I’d like to work on this issue.

### Planned fixes:
- Update the outdated ESLint rules count from `600+` → `750+` in `docs/.vitepress/theme/components/home/FeatureCheck.vue`
- Refresh npm downloads count
- Refresh GitHub stars count

I’ll start by manually updating the current outdated stats to ensure the documentation reflects the latest values.

Please assign this issue to me — happy to take care of it.

nozomee:
@jong-kyung I've submitted a Pull Request to address these updates:  #1512

PR #1512, opened 2026-05-04
docs: refresh home stats for oxlint, vite, and vitest  Issue: #1511

nozomee:
### 1. `docs/.vitepress/theme/components/home/FeatureCheck.vue`

- **ESLint rule count (Vite+ check section):** `600+` → `750+` (text before the `ESLint` code span).

### 2. `docs/.vitepress/theme/components/home/ProductivityGrid.vue`

**“A trusted stack to standardize on” — Vite**

- Weekly npm downloads: `69m+` → `115m+`
- GitHub stars: `78.7k` → `80.4k`

**Vitest**

- Weekly npm downloads: `35m+` → `57m+`
- GitHub stars: `16.1k` → `16.5k`

**OXC** (npm row uses `oxlint` package downloads)

- Weekly npm downloads: `5m+` → `5.3m+`
- GitHub stars: `19.8k` → `21.0k`

---

**Files touched:** 2  
**Net:** 7 lines updated (string substitutions only; no logic or layout changes).


jong-kyung (approved):
LGTM

mdong1909:
We want a way to automate the npm downloads & github stars. Otherwise the hard coded numbers will be outdated quickly. For example, vitest had 7% increase in downloads in the last week.

A potential way to do this is to have a fetch data script that is executed on a schedule via github actions. The data is stored as a json in this repo, which the home pages references

nozomee:
I’d like to take this on — I’m already actively working on it and will open a PR soon.

Right now the trusted-stack numbers on the homepage are hard-coded, which means they’ll drift quickly and require repeated manual edits over time (especially with npm download velocity changing week-to-week).

My planned implementation:

Add a lightweight fetch script that pulls:
Weekly npm downloads for Vite / Vitest / OXC
GitHub stars for their corresponding repos
Store that data in a committed JSON file inside the docs theme as the single source of truth
Replace hard-coded homepage values with JSON-driven values at build time
Add a scheduled GitHub Action (weekly + manual dispatch) that refreshes the JSON automatically and only commits when values actually change

Why this should help:

Keeps homepage metrics current automatically
Preserves the static-site model (no client-side API calls)
Avoids recurring hand-edits
Keeps changes transparent in git history
Contributors won’t need to do anything extra unless they specifically want an immediate local refresh
In the normal case, maintainers shouldn’t need to manually update stats either—the scheduled workflow handles it

Expected flow after merge:

Merge once
Workflow refreshes stats automatically on schedule
Homepage numbers stay fresh with near-zero maintenance
Manual command is only for immediate refreshes or branch-protection fallback

A few maintainer decisions may still be useful:

Whether github-actions[bot] should push JSON-only updates directly to main or use PRs
Preferred refresh cadence
Whether oxlint remains the best npm proxy for OXC

In short: automate the trusted-stack stats pipeline once, remove repetitive manual maintenance, and keep homepage credibility high with minimal long-term overhead. PR coming soon.

nozomee:
@jong-kyung @mdong1909  I’ve committed the changes to address these updates, Kindly review it once!

jong-kyung:
@codex review

docs/.vitepress/env.d.ts
```diff
@@ -28,3 +28,8 @@ declare module '*.jpg' {
   const src: string;
   export default src;
 }
+
+declare module '*.json' {
```
jong-kyung: Is this needed? **tsconfig.json** already has `resolveJsonModule: true`.

.github/workflows/update-trusted-stack-stats.yml
```diff
@@ -0,0 +1,41 @@
...
+    if: github.repository == 'voidzero-dev/vite-plus' && github.event.repository.fork == false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: '22'
```
jong-kyung: Could this use `node-version-file: .node-version` to match `release.yml` and avoid drift when the repo bumps Node?

docs/.vitepress/env.d.ts
nozomee: resolveJsonModule: true is enough for TypeScript to allow import raw from './trusted-stack-stats.json'. The extra declare module '*.json' was only widening the default export to Record<string, unknown>; we already narrow with as TrustedStackStatsFile in trusted-stack-stats.ts, so the ambient declaration was redundant. Removed it.

.github/workflows/update-trusted-stack-stats.yml
nozomee: Done — switched to node-version-file: .node-version so it tracks the repo’s Node version like release.yml and we don’t have to update the workflow when .node-version changes.

nozomee:
@jong-kyung @mdong1909 I’ve committed the changes , Kindly review it once!

docs/scripts/fetch-trusted-stack-stats.mjs
```diff
@@ -0,0 +1,117 @@
...
+      npmPackage: p.npmPackage,
+      githubRepo: p.githubRepo,
+      npmWeeklyDownloads,
+      githubStargazers: stars,
+      npmWeeklyDownloadsDisplay: formatWeeklyDownloads(npmWeeklyDownloads),
+      githubStarsDisplay: formatStars(stars),
+    });
+  }
+  const payload = {
+    generatedAt: new Date().toISOString(),
```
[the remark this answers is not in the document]
jong-kyung: @nozomee Could you consider whether `generatedAt` is actually needed here? Since git log on this file already shows when it was last refreshed, dropping the field might be simpler.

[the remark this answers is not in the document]
nozomee: Dropped generatedAt from trusted-stack-stats.json so the workflow only commits when stable fields (downloads/stars/display strings) actually change; last refresh stays visible via git history on the file. The fetch script still logs the write time to stdout for CI logs.

kindly review the latest commit @jong-kyung , thanks!

jong-kyung (approved):

docs/.vitepress/theme/data/fetch-trusted-stack-stats.ts
camc314, line 1: the min node version for vite-plus is `22.18.0`, which can run typescript natively. So I think it makes sense to convert this to a typescript script and get the typechecking advantages that come with that change

.github/workflows/update-trusted-stack-stats.yml
camc314, line 1: I _think_ writing directly to `main` is a potential security vulnerability.

To address this, I think it should work in the same way that [upgrade-deps.yml](https://github.com/voidzero-dev/vite-plus/blob/31cdef67fa086e08d28e3eb5c732ff4e26db30ea/.github/workflows/upgrade-deps.yml#L1-L0), and make a PR instead.

nozomee: Thanks for raising this — completely agree.

Writing directly to `main` from a scheduled workflow introduces unnecessary risk, even for a JSON-only docs update. To align this with existing repo security practices (like `upgrade-deps.yml`), I’ll update the workflow so it no longer pushes directly to the default branch.

### Planned change
**Before**
Scheduled workflow:
Fetch stats → Commit JSON → Push directly to `main`

**After**
Scheduled workflow:
Fetch stats → Create/Update PR → Maintainer reviews → Merge manually

### Implementation approach
- Switch to `peter-evans/create-pull-request`
- Use a dedicated automation branch (e.g. `chore/docs-trusted-stack-stats`)
- Target `main` via PR
- Add `pull-requests: write`
- Add `persist-credentials: false` for checkout hardening

### Benefits
- No direct writes to `main`
- Preserves branch protection
- Human review before merge
- Matches `upgrade-deps.yml`
- Safer and more maintainable

I’ll update this to follow the repo’s existing PR-first automation model.

docs/.vitepress/theme/data/fetch-trusted-stack-stats.ts
nozomee: Good call — since vite-plus already requires Node `22.18.0`, it makes sense to take advantage of native TypeScript execution here instead of keeping this as `.mjs`.

### Planned change
Convert:
`fetch-trusted-stack-stats.mjs` → `fetch-trusted-stack-stats.ts`

### Why this is better
- Native TS support in the project’s minimum Node version
- Stronger type safety for API responses
- Shared types between:
  - Fetch script
  - Generated JSON
  - Theme consumer
- Easier refactors
- Better long-term maintainability
- Catch schema drift at build/typecheck time instead of runtime

### Additional improvements
- Add shared types file for `trusted-stack-stats`
- Add dedicated `tsconfig` for script validation
- Add a typecheck command (e.g. `pnpm docs:update-trusted-stack-stats:check`)

### Result
This turns the stats pipeline into a more robust typed system while reducing the chance of JSON shape mismatches or silent runtime regressions.

I’ll update the implementation to use TypeScript so it better matches the project’s Node baseline and tooling standards.

nozomee:
@jong-kyung @camc314  I’ve committed the changes , Kindly review it once!

jong-kyung:
@codex review

jong-kyung:
@nozomee Could you take a look at the failing CI first?

docs/.vitepress/theme/data/fetch-trusted-stack-stats.ts
jong-kyung, line 1: Could this script live under `.github/scripts/`?
It's CI-only and uses the same peter-evans/create-pull-request pattern as `upgrade-deps.yml`.

jong-kyung:
@nozomee Could you take a look at the failing CI?

nozomee:
@jong-kyung I’ve committed the changes , Kindly review it once!

.github/workflows/update-trusted-stack-stats.yml
```diff
@@ -0,0 +1,48 @@
...
+          persist-credentials: false
+
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version-file: .node-version
+
+      - name: Fetch npm and GitHub stats
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node docs/scripts/fetch-trusted-stack-stats.ts
```
jong-kyung: The script lives at `.github/scripts/fetch-trusted-stack-stats.ts`, not `docs/scripts/`.

.github/scripts/tsconfig.json
jong-kyung, line 1: Is this file actually needed? The root `tsconfig.json` already enables the same options

docs/package.json
jong-kyung, line 1: Are these two new `devDependencies` actually needed here? `@types/node` and `typescript` are already provided by the workspace root.

.github/scripts/tsconfig.json
nozomee: Root tsconfig.json isn’t a good fit because it lacks DOM lib (needed for fetch) and would typecheck too much of the repo. I moved the minimal config to .github/scripts/tsconfig.json (since it’s for CI scripts) and updated the docs check script to use it; removed docs/scripts/tsconfig.json, will push the commit changes shortly.

.github/workflows/update-trusted-stack-stats.yml
nozomee: Good catch — you’re right.

I moved the fetch script to `/.github/scripts/fetch-trusted-stack-stats.ts` to align with the existing CI scripts (similar to `upgrade-deps.yml`). I’ve updated the workflow step to run:

`node .github/scripts/fetch-trusted-stack-stats.ts`

so it matches the actual location. 

will push the changes shortly!

.github/scripts/tsconfig.json
camc314: > Root tsconfig.json isn’t a good fit because it lacks DOM lib (needed for fetch) 

This script shouldn't rely on `DOM` lib since it'll be executing in a node env.

nozomee: > > Root tsconfig.json isn’t a good fit because it lacks DOM lib (needed for fetch)
> 
> This script shouldn't rely on `DOM` lib since it'll be executing in a node env.

Good point — agreed. Even though the script uses fetch, it runs in Node, so it shouldn’t depend on the DOM lib.

docs/package.json
nozomee: I looked into this, and with our current setup it’s not really safe to remove those dependencies.

Even though typescript and @types/node exist at the workspace root, the docs/ directory is installed and locked independently (pnpm -C docs install --frozen-lockfile with its own docs/pnpm-lock.yaml). That means docs/ won’t reliably inherit root devDependencies unless we switch to a full workspace-level install.

Since docs/ also runs its own TypeScript check (update-trusted-stack-stats:check), it needs local access to tsc plus Node typings for imports and globals.

So for now, keeping typescript and @types/node in docs/package.json makes the most sense — it keeps docs/ self-contained and avoids CI or isolated install issues.

If we want to remove that duplication later, we’d first need to change the install/CI flow to rely on a root workspace install rather than pnpm -C docs.

jong-kyung: I think we can simplify further: the `:check` script isn't wired into any CI job, so it's not actually catching anything.

Could we drop:                                      
  - docs/scripts/tsconfig.json
  - The update-trusted-stack-stats(:check) scripts in docs + root
  - @types/node and typescript from docs/devDependencies
  
  cc. @camc314 WDYT?

nozomee: Good point — let’s see what @camc314  thinks first, then we can figure out the best path forward from there. 
Happy to help once we’ve got their input.

nozomee:
@jong-kyung Looks like this is only a packageManager version conflict (10.33.0 vs 10.33.2). Could you check which version we want to standardize on across the repo/CI and align this accordingly? Once we confirm the intended pnpm version, we can resolve it cleanly.

jong-kyung:
@nozomee On pnpm: let's stick with 10.33.0 as pinned in `package.json`.

That said — the version mismatch you're seeing is really a lockfile drift issue, not a pnpm version issue. It only exists because `docs/package.json` gained `typescript` and `@types/node`, which forced `docs/pnpm-lock.yaml` to be regenerated.

If we go with the simplification (drop `:check` script, tsconfig, and the two devDependencies), the docs lockfile change goes away entirely and this pnpm question becomes moot. 
I'd still lean toward that direction since the `:check` script isn't wired into CI.

cc. @camc314

nozomee:
@jong-kyung I’ve committed the changes , Kindly review it once!

jong-kyung:
@codex review

docs/package.json
```diff
@@ -23,5 +24,5 @@
     "tailwindcss": "^4.1.18",
     "vitepress": "2.0.0-alpha.15"
   },
-  "packageManager": "pnpm@10.33.2"
-}
\ No newline at end of file
+  "packageManager": "pnpm@10.33.0"
```
jong-kyung: Could you realign this back to `pnpm@10.33.2`?

.github/scripts/fetch-trusted-stack-stats.ts
```diff
@@ -0,0 +1,125 @@
...
+const PROJECTS: readonly ProjectSource[] = [
+  { id: 'vite', npmPackage: 'vite', githubRepo: 'vitejs/vite' },
+  { id: 'vitest', npmPackage: 'vitest', githubRepo: 'vitest-dev/vitest' },
+  /** OXC row uses `oxlint` npm weekly downloads as a concrete proxy for the Oxc toolchain. */
+  { id: 'oxc', npmPackage: 'oxlint', githubRepo: 'oxc-project/oxc' },
+];
+
+function formatWeeklyDownloads(n: number): string {
+  if (n >= 10_000_000) {
+    return `${Math.round(n / 1e6)}m+`;
```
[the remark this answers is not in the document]
jong-kyung: what about using `Math.floor` here so the `+` suffix stays an honest lower bound?

docs/pnpm-lock.yaml
jong-kyung, line 1: Could you check whether these lockfile changes are actually needed in this PR, and revert them if not?

.github/scripts/fetch-trusted-stack-stats.ts
```diff
@@ -0,0 +1,125 @@
...
+ */
+import { writeFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import type {
+  TrustedStackProjectId,
+  TrustedStackStatProject,
+  TrustedStackStatsFile,
+} from '../../docs/.vitepress/theme/data/trusted-stack-stats.types.ts';
```
jong-kyung: I think `fetch-trusted-stack-stats.ts` belongs under `.github/`, but I'm a bit uneasy that types defined in `docs/` end up imported into `.github/`. 

@fengmk2 @camc314  WDYT?

jong-kyung (changes requested):

.github/scripts/fetch-trusted-stack-stats.ts
fengmk2: Should we place `fetch-trusted-stack-stats.ts` in `docs/.vitepress/theme/data/`? This way it's also possible to run this script during local development.

jong-kyung: @fengmk2 That makes sense — I had been thinking along the lines of co-locating it with the workflow since it's CI-driven, but keeping it next to the JSON it produces is a cleaner home. Thanks☺️

nozomee:
@jong-kyung hey there, check the latest commits , added the changes mentioned above!

jong-kyung:
@codex review

jong-kyung (commented):
@nozomee Thanks for working through all the feedback — looks great!
Could you resolve the merge conflict with main?

nozomee:
> @nozomee Thanks for working through all the feedback — looks great! Could you resolve the merge conflict with main?

done!

jong-kyung (approved):

.github/workflows/update-trusted-stack-stats.yml
```diff
@@ -0,0 +1,48 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node docs/.vitepress/theme/data/fetch-trusted-stack-stats.ts
+
+      - name: Create or update PR
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7.0.11
+        with:
+          base: main
+          branch: chore/docs-trusted-stack-stats
+          title: 'chore(docs): refresh trusted stack stats'
+          token: ${{ secrets.GITHUB_TOKEN }}
```
fengmk2: The token for creating PRs has been changed to an app token. You can refer to https://github.com/voidzero-dev/vite-plus/blob/main/.github/workflows/upgrade-deps.yml#L276

nozomee:
@jong-kyung @fengmk2 kindly check the changes! Pushed all the updated commits

fengmk2 (approved):
@nozomee I will take over fixing the vp check error issue, and will merge it today.

fengmk2:
https://github.com/voidzero-dev/vite-plus/pull/1573 action works, thanks again @nozomee @jong-kyung

nozomee:
Loved working with you both — learned a lot and really appreciated it! 🙌 @fengmk2 @jong-kyung

PR #1515, opened 2026-05-05
refactor(team): convert team data file to TypeScript

jong-kyung:
Convert `docs/_data/team.js` to TypeScript for consistency with the rest of the `docs/`, where every other source file is already `.ts/.mts/.vue`.


tsconfig.json
```diff
@@ -17,6 +17,7 @@
   },
   "exclude": [
     "ecosystem-ci",
+    "packages/cli/docs",
```
fengmk2, line 20: Why exclude docs? Are there type errors?

jong-kyung: Yes — the type error was caught in https://github.com/voidzero-dev/vite-plus/actions/runs/25370462495/job/74392468326.

fengmk2 (approved):

PR #1517, opened 2026-05-05
docs: add Homebrew install instructions

charpeni:
There's now a formula for [`vite-plus`](https://formulae.brew.sh/formula/vite-plus) on Homebrew.

CleanShot 2026-05-05 at 12 28 08@2x

### DIFF
--- a/docs/.vitepress/theme/components/home/InstallCommand.vue
+++ b/docs/.vitepress/theme/components/home/InstallCommand.vue
@@ -10,6 +10,11 @@ type CommandCard = {
 
 const commandCards: CommandCard[] = [
   {
+    id: 'macos-homebrew',
+    label: 'macOS (Homebrew)',
+    command: 'brew install vite-plus',
+  },
+  {
     id: 'unix',
     label: 'macOS / Linux',
     command: 'curl -fsSL https://vite.plus | bash',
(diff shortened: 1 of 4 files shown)


docs/.vitepress/theme/components/home/InstallCommand.vue
```diff
@@ -9,6 +9,11 @@ type CommandCard = {
 };
 
 const commandCards: CommandCard[] = [
+  {
+    id: 'macos-homebrew',
+    label: 'macOS (Homebrew)',
+    command: 'brew install vite-plus',
```
fengmk2, line 15: The priority of `Homebrew` installation is to be placed last. Once the winget and apt methods are supported, we need to unify these distribution channels together.

chungweileong94:
The current vite-plus on homebrew-core doesn't really work properly on `vp env`, which I don't think it's time to promote that just yet.

A bit more context is that part of the brew installation will also install `node` as a dependency, which breaks the vite-plus shim to node.

fengmk2:
@chungweileong94 Thanks for the feedback, I'll check out the Homebrew vite-plus installation script and improve the issues you mentioned. Could you submit an issue to vite-plus to report this problem too?

chungweileong94:
> @chungweileong94 Thanks for the feedback, I'll check out the Homebrew vite-plus installation script and improve the issues you mentioned. Could you submit an issue to vite-plus to report this problem too?

Done https://github.com/voidzero-dev/vite-plus/issues/1572

PR #1524, opened 2026-05-06
chore(tools): prevent false-positive `<repeat>` replacement on trailing plus

nurazon59:
Discovered while reviewing CI results after approving #1508.

The regex `/\++\n/g` was overly aggressive and matched any line ending with a single `+`, including legitimate text like "Vite+" in the CLI help output.

Changed to `/\+{2,}\n/g` to only match pnpm's actual repeated `+` progress indicators (e.g., `+++`).


fengmk2 (approved):

fengmk2 (commented):
@nurazon59 Thanks, I will regen the snap test diff before merge.

PR #1549, opened 2026-05-11
feat: Forward `--provenance` flag when publishing a package.

cpojer:
This required a `--` in between before, which is confusing. This PR makes it so `--provenance` is an actual flag for `vp pm publish`.


fengmk2 (approved):

fengmk2:
I will fix the snap test diff and merge.

PR #1576, opened 2026-05-14
test(cli): verify .gitignore and .yarnrc.yml in new-vite-monorepo snap

jong-kyung:
## Summary

`ls | sort` skips dotfiles, so the `new-vite-monorepo` snap test never noticed if:
- `_gitignore → .gitignore` rename failed
- pnpm-branch `.yarnrc.yml` cleanup broke — the bun snap covers the npm/bun branch but not pnpm

Adds two `test -f` / `test ! -f` assertions


fengmk2 (approved):

jong-kyung:
@fengmk2 Super Fast !! 😳 Thanks!

PR #1582, opened 2026-05-15
fix(cli): preserve single-segment project path in updateWorkspaceConfig

jong-kyung:
## Summary

`vp create --directory website` was writing `packages: [ ./* ]` to `pnpm-workspace.yaml` instead of `packages: [ website ]`. Fix the pattern-derivation so single-segment paths stay literal.

## Cause

```ts
  let pattern = path.dirname(projectPath);
  if (!pattern) {
    // intent : "website" -> "website"
    pattern = projectPath;
  } else {
    // "foo/bar/app" -> "foo/bar/*"
    pattern = `${pattern}/*`;
  }
```

`path.dirname('website')` returns `'.'`, not `''`. `'.'` is truthy, so the literal branch never fired — every single-segment path became `./*`. 

The fix:
```ts
const parentDir = path.dirname(projectPath);
const pattern = parentDir === '.' ? projectPath : `${parentDir}/*`;
```

`./*` over-matches in pnpm (picks up every root sibling with a `package.json`) and accumulates on repeat calls because `minimatch('name', './*')` returns false, so the skip-when-matched guard misses it.


jong-kyung:
@codex review

jong-kyung:
@fengmk2 Could you please re-run the CI for this PR?

fengmk2 (approved):

Issue #1476, opened 2026-04-25
Enhance `vp install -g` and `vp update -g` with multiple packages


liangmiQwQ:
## Description

The current `vp install -g`, and `vp update -g` will just install / reinstall packages one by one when multiple packages provided. It takes a lot of time, as well as multiple node-resolving and processes creating are also not good for performance.

We should follow the similar behavior like npm and pnpm, process the download in parallel instead.

## Suggested solution

We have two ways to improve it, `vp update` should follow similar pipeline as well.

### Delegate installing to `npm`

The current vp essentially spawns processes running `npm install <pkg>` for every single package provided. But actually `npm` itself can process parallel installing, we can just spawn one single `npm` process and create shim unitedly after npm's work gets done.

For example:
Before: `vp install -g oxlint oxfmt` -> `npm install oxlint && npm install oxfmt`
After: `vp install -g oxlint oxfmt` -> `npm install oxlint oxfmt`

<details>

<summary>🤖 A flow chart generated by Claude</summary>

```bash
# How current vp works
$ vp install -g pkg1 pkg2 pkg3
│
└─ cli.rs:1622  Commands::Install { global: true, packages: Some([pkg1, pkg2, pkg3]), … }
   │
   └─ cli.rs:1655-1666   if global { for package in &pkgs { … } }     ← SERIAL LOOP
      │                                                                  (no join_all,
      │                                                                   no concurrency,
      │                                                                   first failure aborts the rest)
      │
      ├─ iter 1 ─ global_install::install("pkg1", node, force).await
      ├─ iter 2 ─ global_install::install("pkg2", node, force).await   ← waits for iter 1
      └─ iter 3 ─ global_install::install("pkg3", node, force).await   ← waits for iter 2

  Per-iteration (global_install.rs:28-209)
  ─────────────────────────────────────────
  install(package_spec, node_version, force)
  │
  ├─ 1. Resolve Node version
  │     ├─ if --node X            → resolve_version_alias(X)
  │     └─ else                   → resolve_version(cwd)        (reads engines / .nvmrc)
  │
  ├─ 2. Ensure Node runtime installed
  │     └─ vite_js_runtime::download_runtime(Node, version)
  │
  ├─ 3. Stage on disk
  │     ├─ tmp_dir = ~/.vite-plus/tmp
  │     ├─ staging = tmp/packages/<pkg>          ← per-package dir, no cross-pkg collision
  │     └─ rm staging if leftover, mkdir -p staging
[...]

PR #1596, opened 2026-05-16
fix(global): skip current packages on update

### LINKED ISSUE #1477: Skip packages which are up to date in `vp update -g`

leno23:
## Summary

- Skip `vp update -g` installs when the recorded global package version already matches the version resolved from npm.
- Resolve the target registry version with the package's recorded Node runtime so checks match the package's managed environment.
- Support npm `view ... version --json` outputs as strings, plain strings, and version arrays.

Fixes #1477
Related #1476

## Test Plan

- `cargo fmt`
- `cargo test -p vite_global_cli global_update --no-default-features`
- `cargo test -p vite_global_cli updates_global_package_when_registry_version_differs_from_installed_version --no-default-features`
- `cargo test -p vite_global_cli parse_npm_view_version --no-default-features`
- `cargo test -p vite_global_cli --no-default-features --no-run`
- `cargo clippy -p vite_global_cli --no-default-features -- -D warnings`

Note: `cargo test -p vite_global_cli --no-default-features` aborts locally in the existing `tests::unknown_argument_detected_with_pass_as_value_hint` test with a stack overflow after the targeted tests pass.

### DIFF
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -8,7 +8,7 @@ use std::{ffi::OsStr, process::ExitStatus};
 use clap::{CommandFactory, FromArgMatches, Parser, Subcommand};
 use clap_complete::ArgValueCompleter;
 use tokio::runtime::Runtime;
-use vite_path::AbsolutePathBuf;
+use vite_path::{AbsolutePath, AbsolutePathBuf};
 use vite_pm_cli::PackageManagerCommand;
 
 use crate::{commands, error::Error, help};
@@ -528,7 +528,7 @@ async fn run_package_manager_command(
         }
 
         PackageManagerCommand::Update { global: true, ref packages, .. } => {
-            managed_update(packages).await
+            managed_update(&cwd, packages).await
         }
 
         // `pm list -g` lists vite-plus-managed globals, not the underlying PM's.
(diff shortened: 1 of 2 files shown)


liangmiQwQ:
Thank you!

crates/vite_global_cli/src/commands/env/global_install.rs
```diff
@@ -267,8 +267,71 @@ pub async fn uninstall(package_name: &str, dry_run: bool) -> Result<(), Error> {
     Ok(())
 }
 
+/// Resolve the version currently published for a package spec.
+///
+/// `package_spec` may be a bare package name (`typescript`) or include a
+/// version/tag (`typescript@beta`, `@scope/pkg@1.0.0`). The command returns the
```
liangmiQwQ, line 273: As `cli.rs` directly pass `metadata.name` here, `package_spec` should always be pure package name without version/tag?

crates/vite_global_cli/src/commands/env/global_install.rs
```diff
@@ -267,8 +267,71 @@ pub async fn uninstall(package_name: &str, dry_run: bool) -> Result<(), Error> {
...
+/// version that npm resolves for that spec.
+pub(crate) async fn latest_package_version(
+    package_spec: &str,
+    node_version: Option<&str>,
```
liangmiQwQ, line 277: There it uses the same node version which installed this package to view version, what if we just resolve from current directory like `install` function did?

liangmiQwQ (commented):
When I made up #1477, I actually thought about a question: should we reinstall the packages installed with a different node version.

For example, you installed `oxlint` globally with `node@21.0.0`, but now you are running `vp update -g` with `node@25.0.0`. For current vp, it will be re-installed.

The reason why I have this question is different nodes can have different installing behaviors, like `engine.node` compatibility checks or some post-install scripts.

By the way, I'm not the team, and there are several comments I left, please use this as a reference only.

/cc @fengmk2 What do you think of it?

fengmk2:
@liangmiQwQ What you mentioned should be a new feature requirement, and you can open a new issue for discussion.

fengmk2:
@codex review

wyf027:
Addressed the review feedback in ac4ed26:

- `vp update -g` now falls back to updating a package if the registry version lookup fails, instead of aborting the entire update.
- Local/path package specs (`.`, `./pkg`, `../pkg`, `file:...`) bypass the registry lookup so local global packages keep the previous reinstall behavior.
- The skip check now compares both package version and the current target Node version. Packages installed with a different Node version are reinstalled under the current resolved Node version.
- The version lookup/update path now uses the same resolved target Node version as the subsequent install.

Validation run locally:

```bash
cargo fmt
cargo test -p vite_global_cli global_update --no-default-features
cargo test -p vite_global_cli updates_global_package_when_registry_version_differs_from_installed_version --no-default-features
cargo test -p vite_global_cli updates_global_package_when_node_version_differs_from_target_version --no-default-features
cargo test -p vite_global_cli parse_npm_view_version --no-default-features
cargo test -p vite_global_cli test_is_local_package_spec --no-default-features
cargo test -p vite_global_cli --no-default-features --no-run
cargo clippy -p vite_global_cli --no-default-features -- -D warnings
git diff --check
```

liangmiQwQ:
> * The skip check now compares both package version and the current target Node version. Packages installed with a different Node version are reinstalled under the current resolved Node version.

~~Maybe we need to discuss about it in #1598 first? I'm sorry I didn't explain about it in the original issue.~~

I agree on how you process this problem in this PR, really thanks for solving my issue!

crates/vite_global_cli/src/cli.rs
```diff
@@ -573,21 +573,68 @@ async fn managed_uninstall(packages: &[String], dry_run: bool) -> Result<ExitSta
...
             return Ok(ExitStatus::default());
         }
-        all.iter().map(|p| p.name.clone()).collect()
+
+        for metadata in all {
+            let latest_version = global_install::latest_package_version(
+                &metadata.name,
+                Some(&metadata.platform.node),
+            )
+            .await?;
```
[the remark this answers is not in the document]
wyf027: Thanks, addressed in ac4ed26. `vp update -g` now catches per-package lookup failures and falls back to updating that package, so one private/yanked package or transient registry error no longer aborts the full update loop.

crates/vite_global_cli/src/commands/env/global_install.rs
wyf027: Good point. To keep this PR scoped to #1477 and avoid deciding the new #1598 behavior, the current patch preserves the previous behavior when the target Node version differs: it only skips when both the package version and the resolved target Node version match the recorded metadata. If the package was installed under a different Node version, `vp update -g` still reinstalls it under the current resolved Node version, as before.

So #1598 can still discuss whether that behavior should change later; this PR only avoids redundant reinstalls for the same package version under the same Node version.

crates/vite_global_cli/src/cli.rs
```diff
@@ -573,21 +573,68 @@ async fn managed_uninstall(packages: &[String], dry_run: bool) -> Result<ExitSta
...
     } else {
-        packages.to_vec()
-    };
+        for package in packages {
+            let (package_name, _) = global_install::parse_package_spec(package);
+            if let Some(metadata) = PackageMetadata::load(&package_name).await? {
+                let latest_version =
+                    global_install::latest_package_version(package, Some(&metadata.platform.node))
+                        .await?;
+                if is_global_package_up_to_date(&metadata.version, &latest_version) {
```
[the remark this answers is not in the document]
wyf027: Thanks, addressed in ac4ed26. Local/path specs (`.`, `..`, `./pkg`, `../pkg`, `file:...`) now bypass the registry-version skip and keep the previous reinstall behavior, so local package contents are refreshed even when `package.json#version` did not change.

wyf027:
Agreed that changing the cross-Node reinstall policy should be discussed in #1598. I adjusted the wording/implementation intent here: this PR keeps the existing behavior for packages installed under a different Node version (they still reinstall under the current resolved Node version) and only skips redundant reinstalls when the recorded package version and Node version both match. So #1598 remains the place to decide whether that behavior should change in the future.

fengmk2 (approved):
@leno23 Thanks!

PR #1601, opened 2026-05-16
fix(cli): skip merging standalone oxfmt/oxlint config when key already in vite.config.ts

fengmk2:
## Summary

- `vp create fate` produced a `vite.config.ts` with two `fmt:` blocks because `mergeAndRemoveJsonConfig` (`packages/cli/src/migration/migrator.ts`) was the only `mergeJsonConfig` call site missing the "skip if key already present" guard the other call sites use. The Rust merge step always prepends, so when fate's template ships both an inline `fmt:` block and a standalone `.oxfmtrc.jsonc`, both ended up in the output.
- Now reads `vite.config.ts` first; if the key is already declared, `fs.unlinkSync`s the redundant standalone file and emits an info log instead of merging. Mirrors the existing skip-existing behavior in `applyToolInitConfigToViteConfig`. Also covers the analogous `lint:` + `.oxlintrc.json` case.

## Test plan

- [x] New unit tests in `packages/cli/src/migration/__tests__/migrator.spec.ts` (fmt + lint), confirmed red before the fix and green after — `vp test --run packages/cli/src/migration/__tests__/migrator.spec.ts` (75/75 passing).
- [x] New global snap test `packages/cli/snap-tests-global/migration-preserves-existing-fmt-and-lint/` modeled on fate's real-world fmt/lint config (experimental options, ignore patterns, overrides); standalone files carry conflicting values so any regression would be obvious in the snapshot.
- [x] `pnpm -F vite-plus snap-test-global migration` — all existing migration snaps still pass with no diffs.
- [x] `pnpm -F vite-plus snap-test-global new-create-vite` — all create snaps still pass with no diffs.

Closes the duplicate-fmt regression reported when running `vp create fate`.

---

> [!NOTE]
> **Medium Risk**
> Migration logic now conditionally deletes standalone `.oxfmtrc`/`.oxlintrc` files based on a new AST-based vite config scan exposed via a native binding; mistakes in the detector could cause config merges to be skipped or applied unexpectedly.
> 
> **Overview**
> Fixes a migration regression where projects with an existing inline `fmt`/`lint` block in `vite.config.*` could end up with duplicated blocks after merging standalone `.oxfmtrc`/`.oxlintrc` files.
> 
> This adds a Rust AST-based `has_config_key` check (exported through the NAPI binding as `hasConfigKey`) and uses it in the CLI migration paths to **skip merging** when the key already exists, instead deleting the redundant standalone config and optionally logging an info message. New unit and snapshot tests cover the create-fate-style template shape and ensure only a single `fmt`/`lint` block remains.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0c99be9fc01c6af4d493731e4408de1b1d883ca. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/vite_migration/src/lib.rs
+++ b/crates/vite_migration/src/lib.rs
@@ -18,4 +18,4 @@ mod vite_config;
 pub use file_walker::{WalkResult, find_ts_files};
 pub use import_rewriter::{BatchRewriteResult, rewrite_imports_in_directory};
 pub use package::{rewrite_eslint, rewrite_prettier, rewrite_scripts};
-pub use vite_config::{MergeResult, merge_json_config, merge_tsdown_config};
+pub use vite_config::{MergeResult, has_config_key, merge_json_config, merge_tsdown_config};
--- a/crates/vite_migration/src/vite_config.rs
+++ b/crates/vite_migration/src/vite_config.rs
@@ -1,6 +1,7 @@
 use std::{borrow::Cow, path::Path, sync::LazyLock};
 
 use ast_grep_config::{GlobalRules, RuleConfig, from_yaml_string};
+use ast_grep_core::{Doc, Node};
 use ast_grep_language::{LanguageExt, SupportLang};
 use regex::Regex;
 use vite_error::Error;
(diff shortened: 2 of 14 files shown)


fengmk2:
@cursor review

cpojer (approved):
Haha, I didn't realize the fate templates ship with duplicate configs. Woops. Thank you for the fix in Vite+!

fengmk2:
@cpojer fate templates help me realize how fragile the previous temporary solutions were 😢

PR #1606, opened 2026-05-17
fix(cli): detect create output in current directory

### LINKED ISSUE #1605: Svelte project defaults to Vite when defaulting a prompt
### Describe the bug

I am doing
Creating a new project with `vp create svelte`

What I expect is
A new project to be created with tooling and configuration made by the `vp` CLI.

What is actually happening
Creates a standard Vite project with zero presence of VP or any of it's tooling

Root Cause of Issue
I did a bunch of experimenting around, recreating a svelte project time and time again. What I found is that this issue only occurs if we don't specify anything in the prompt "Where would you like your project to be created?"  when `bun x sv create --no-install` is running. If I don't specify anything and leave it to it's default, the entire project ends up being a standard Vite project as if Vite+ was never ran in the first place.

### Reproduction

This is a CLI issue, cannot really reproduce it in a sandbox or through a minimal reproduction repository. The bug must be reproduced via the CLI itself

### Steps to reproduce

Already described above, leaving an unspecified default option in the prompt `Where would you like your project to be created?`

### System Info

```shell
`vp env current` gives the output
Environment:
  Version  24.15.0
  Source   lts

`vp --version` gives the output
vp v0.1.21

Local vite-plus:
  vite-plus  v0.1.21

Tools:
  vite             v8.0.11
  rolldown         v1.0.0
  vitest           v4.1.5
  oxfmt            v0.48.0
  oxlint           v1.63.0
  oxlint-tsgolint  v0.22.1
  tsdown           v0.22.0

Environment:
  Package manager  bun v1.3.14
  Node.js          v24.15.0
```

### Used Package Manager

bun

### Logs

```shell

```

### Validations

jong-kyung:
## Summary

`vp create svelte` runs the Svelte `sv` generator through the remote template/package path. That generator can create the project directly in the command's current directory, but the remote create-output detection previously ignored root-level `package.json` writes when resolving the created project directory.

This now treats a current-directory `package.json` write as `projectDir: "."` when no generated child directory is found. Child directory creation still takes precedence, so existing remote template flows that create `my-app/package.json` keep the same behavior.

Focused tests cover current-directory detection, child-directory precedence, and parent-directory creation.

Closes #1605

### DIFF
--- a/packages/cli/src/create/__tests__/command.spec.ts
+++ b/packages/cli/src/create/__tests__/command.spec.ts
@@ -1,0 +1,100 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockRunCommand } = vi.hoisted(() => ({
+  mockRunCommand: vi.fn(),
+}));
+
+vi.mock('../../../binding/index.js', () => ({
+  runCommand: mockRunCommand,
+}));
+
+const { runCommandAndDetectProjectDir } = await import('../command.js');
+
+const tempDirs: string[] = [];
+
+function makeTempDir() {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'vp-create-command-'));
+  tempDirs.push(dir);
+  return dir;
+}
+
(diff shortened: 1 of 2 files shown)


jong-kyung:
@codex review

fengmk2 (approved):

Issue #1557, opened 2026-05-12
lint,fmt: Deprecate the wrapper bin for stable IDE support


leaysgur:
Currently, `lint|fmt` IDE support for Vite+ is implemented as follows:

- When the `vite-plus` package is installed locally, the package manager places `bin/oxlint|fmt`
  - These are special binaries with `vite.config.ts` support enabled, distinct from the standard `oxlint|fmt` binaries
- Individual editor extensions then reference these binaries and launch `--lsp`
  - The discovery process for these binaries should remain the same as in non-Vite+ projects

However, some issues have also come to light.

- When using `npm` or `yarn` as the package manager, the custom binary is not installed at all!
  - #1482 
  - `pnpm` and `bun` works
- Installing `oxlint|fmt` separately overwrites the custom binary
  - IMO: This should be triaged as a separate issue, since this overrides `vp lint|fmt` too
- When you try to setup editor by yourself, it's unclear to know what should be used
  - #979
- The priority of the method VSCode extension uses to search the `bin/` was lower than that of `require.resolve()`, causing it to fail to locate the custom binary when other workspaces were open
  - Fixed by: https://github.com/oxc-project/oxc-vscode/pull/223

### Proposed solution

Following our team meeting, we would like to proceed as follows.

- Deprecate the wrapper binary
- Editors that support LSP should use `vp lint|fmt --lsp` instead
- Editors that do not use LSP should use commands like `vp fmt --stdin-filepath`

### Tasks

- [x] Stop writing `VITE+...` header message to stdout for these usecases @fengmk2 
  - `vp lint --lsp`
  - `vp fmt --lsp`, `vp fmt --stdin-filepath`
- [ ] Define a standardized way to determine "whether a project is a Vite+ project or not" @fengmk2 
  - https://github.com/voidzero-dev/vite-plus/pull/1614
- [ ] Update editor extensions to detect Vite+ project and use `vp lint|fmt --lsp`
  - [ ] https://github.com/oxc-project/oxc-vscode/pull/384
  - [ ] https://github.com/oxc-project/oxc-zed/pull/259
  - [ ] https://github.com/oxc-project/oxc-intellij-plugin/pull/583
  - [ ] https://github.com/oxc-project/coc-oxc/pull/247
- [ ] Remove wrapper bin
  - https://github.com/voidzero-dev/vite-plus/pull/2672
- [ ] Update docs?

leaysgur:
@fengmk2 Are we getting this right?

[...]

Issue #1456, opened 2026-04-23
Use oxlint plugin to validate and auto-fix IDE configuration issues


naokihaba:
### Description

Following the discussions in issue #826, it was noted that misconfigurations in the Zed IDE (e.g., an incorrect configPath in oxfmt settings) do not provide any feedback, making it difficult to diagnose why the tool is not working as expected.

Once the implementation of #1408 (oxlint plugin) is complete, we will use this feature to detect and automatically fix issues within .zed/settings.json and .vscode/settings.json.

- Add an oxlint plugin rule to validate IDE configuration files against schemas generated by vp create or vp migrate.
- Support the --fix option to enable automatic corrections for safe, known configuration errors, such as aligning configPath with ./vite.config.ts.
- Integrate this into vp check --fix so that these corrections are performed as part of the regular development workflow.

### Suggested solution

Implement an oxlint plugin rule (e.g. `vite-plus/ide-config`) that reads the Vite+ config and checks IDE config files for known misconfigurations. The rule runs as part of `vp check`, and with `--fix` it rewrites the config to match what `vp create`/`vp migrate` would generate.

### Alternative

A separate `vp config check` subcommand could validate and repair IDE config files without depending on the oxlint plugin infrastructure. However, integrating with `vp check --fix` keeps the workflow unified and avoids adding a new command.

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

PR #1614, opened 2026-05-18
docs(rfc): Vite+ Project Detection for Editor Extensions

fengmk2:
## Summary

Draft RFC defining how the four oxc editor extensions — `oxc-vscode`, `oxc-zed`, `oxc-intellij-plugin`, `coc-oxc` — determine whether a workspace is a Vite+ project and which `vp` executable to spawn for `--lsp`. This is the prerequisite identified in #1557 for removing the per-package `bin/oxlint` / `bin/oxfmt` wrappers.

## Core rule

A workspace is **Vite+** iff some walked-up `package.json` directly declares `vite-plus` in `dependencies` or `devDependencies` (bounded by the workspace root). The runnable `vp` binary is resolved separately, also bounded by the workspace root, and validated against a real `vite-plus` package (`name === "vite-plus"`).

Detector returns a tri-state:

- `null` — not Vite+; editor uses plain oxlint/oxfmt.
- `{ root, vpPath }` — Vite+ and runnable; editor launches `vp lint --lsp` / `vp fmt --lsp`.
- `{ root }` — declared but not yet installed (fresh clone, Berry PnP, broken install); editor falls back to plain oxlint/oxfmt and never spawns a bare `vp`.

A transitively hoisted `node_modules/vite-plus/`, a global `vp`, a `vp` on `\$PATH`, and a configured `oxc.<tool>.binPath` all return `null` by construction.

## Distribution

- `oxc-vscode` and `coc-oxc` consume a shared bundled-devDependency npm package (proposed `@voidzero-dev/detect-vite-plus`) at `packages/detect-vite-plus/`.
- `oxc-zed` (Rust/WASM) and `oxc-intellij-plugin` (Kotlin) port the same algorithm against the shared conformance fixtures inside this RFC.

## Status

**Draft for discussion.** Open points called out in the RFC:

- Final package name.
- Caching policy in editor consumers.
- "Declared but not installed" UX (silent fallback vs. install hint).
- Workspace-root marker parity gap with `vite-task` (recognized as a known follow-up; deliberately not blocked on).

Refs #1557


rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+`VitePlusPackage.kt`; oxc-vscode and coc-oxc have no explicit
+detection. The goal of this RFC is one rule, four implementations.
+
+## The rule
+
+A workspace is **Vite+** iff some `package.json` between the start
+path and the workspace root declares `vite-plus` directly in
+`dependencies` or `devDependencies`. A `node_modules/vite-plus/`
+directory on its own does not qualify — that could be a transitive
+install hoisted from an unrelated dependency tree.
```
Sysix, line 37:
```
A `node_modules/vite-plus/` directory on its own does not qualify — that could be a transitive install hoisted from an unrelated dependency tree.
```

I am not sure if `vite-plus` will ever be a dependency of another package. My naive brain does not like this.
VS Code Extension searched for `package.json` and tries to find `node_modules/.bin/xyz`. 
Opening the package-file, transforming it to JSON, and checking its properties, will be done by every `package.json` it will find. 
On a monorepo this could be expensive. 🤔

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+    vp_path = walk_up_to_workspace_root(root, |dir, _|:
+        return valid_vite_plus_install_at(dir)  # bin/vp + sibling package.json with correct name
+    )
+
+    return Some({ root, vp_path })  # vp_path may be None
+```
+
+Both phases stop AT the workspace root and never cross into its
+parent. The walk-up bound is what prevents a nested checkout from
+inheriting an unrelated parent's Vite+ install.
```
Sysix: Some users will open only a sub-package of a monorepo. Vite-plus from the root will not be detected.
Note: VS Code Extension does search outside the root (beside `require.resolve`), because mostly they are using [Nested Multi Root Workspaces](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces). But we should make sure setups like this will detect vite-plus

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+    P2Valid -- no, orphan --> P2Bound
+    P2Exists -- no --> P2Bound{"probe is workspace root<br/>or filesystem root?"}
+    P2Bound -- no --> P2Up["probe = parent(probe)"]
+    P2Up --> P2Check
+    P2Bound -- yes --> ResultDeclared(["return { root }"])
+```
+
+### Workspace root markers
+
+A directory is a workspace root if any of the following is true:
```
Sysix: Workspace roots are managed by the IDE itself. I do not think we should implement something new. (At least for VS Code).
Specially because users can add/remove workspaces roots from a single IDE session:
https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces

For what is the workspace root marker useful?

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+  `["fmt", "--lsp"]` when launching `vp`. Zed's WASM API only reads
+  the worktree root, so deeper walk-up is a known limitation worth
+  noting in the Zed PR.
+- **`oxc-intellij-plugin`** — `VitePlusPackage.kt` already locates
+  `vite-plus` via IntelliJ's `NodePackageDescriptor`. Tighten it to
+  require a direct dep, change the returned path from
+  `<vite-plus>/bin/oxlint` to `<vite-plus>/bin/vp`, update launch
+  args.
+
+## Conformance fixtures
```
Sysix, line 352: What happens for workspaces where the root package.json does include oxlint, and only a sub-package includes vite-plus?
Should we prefer the vite-plus from the sub-package, even if other packages will use the roots oxlint?
Maybe related: https://github.com/oxc-project/oxc-vscode/issues/145

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+| `yarn4-pnp`                             | Berry/PnP, no `node_modules`, root `package.json` declares `vite-plus`                                                                         | `{ root: "<repo>" }` — install hint                                                                                  |
+
+## Open questions
+
+1. **Publish the detector as a shared npm package?** The current
+   proposal is `@voidzero-dev/detect-vite-plus` at
+   `packages/detect-vite-plus/`, consumed as a bundled devDependency
+   by `oxc-vscode` and `coc-oxc`. The alternative is to let each
+   Node-capable extension copy the ~50-line snippet directly.
+   Decision deferred to the maintainers.
```
Sysix: The package makes sense for me. But I am not sure if this will be useful for other editors as they don't have JS runtime and needs to recreate it anyway by hand.

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+  return Boolean(pkg?.workspaces);
+}
+
+function declaresVitePlus(pkg: any | null): boolean {
+  return Boolean(pkg?.dependencies?.['vite-plus'] || pkg?.devDependencies?.['vite-plus']);
+}
+
+/** `bin/vp` exists AND the sibling package.json identifies as vite-plus. */
+function resolveVpAt(dir: string): string | null {
+  const vpPath = join(dir, 'node_modules', 'vite-plus', 'bin', 'vp');
```
Sysix: Normally we search for `node_modules/.bin/oxlint` and not `node_modules/oxlint/bin/oxlint`. 
Is there a good reason to not target the `.bin` wrapper?

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+(`crates/vite_workspace/src/package_manager.rs:135`) only recognizes
+the first two and carries a `TODO(@fengmk2)` for Lerna. The RFC
+deliberately keeps the broader set; aligning vite-task is a known
+follow-up that does not block this RFC.
+
+### What we deliberately do not check
+
+- `$PATH`, user's global `node_modules`, or
+  `oxc.<tool>.binPath` settings (the last is for oxlint/oxfmt, not
+  `vp`).
```
Sysix: I would like to search for global node_modules and `$PATH` variable too.  
Users want a global editor formatter, which they can use without forcing the repo-dependency.

VP-Users needs to know that `oxfmt` is used and manual install it globally.

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,302 @@
...
+| `parent-vite-plus-nested-repo`          | Outer dir declares + installs `vite-plus`; inner subdir is its own workspace root and does not                                                 | From inside the nested workspace: `null`                                                                             |
+| `plain-non-vite-plus`                   | A normal Node project, no `vite-plus` anywhere                                                                                                 | `null`                                                                                                               |
+| `yarn4-pnp`                             | Berry/PnP, no `node_modules`, root `package.json` declares `vite-plus`                                                                         | `{ root: "<repo>" }` — install hint                                                                                  |
```
Sysix: No Yarn PnP support for vp? For VSCode `oxlint` and `oxfmt` it is supported:
https://github.com/oxc-project/oxc-vscode/blob/aeeb69c6a7990738d31e53dd21b73dce9776f63c/client/findBinary.ts#L171-L196

rfcs/editor-extension-vite-plus-detection.md
fengmk2: In this case, if VS Code opens the root workspace, it won’t be treated as a Vite+ project. We also plan to provide a capability similar to https://github.com/voidzero-dev/vite-plus/issues/1456: ship a rule via Vite+’s built-in oxlint plugin to prevent mixing oxlint and vite-plus within the same monorepo.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: I have considered this issue. If we do not read package.json to determine but instead judge by checking `node_modules/vite-plus`, there will be an unavoidable problem: if `vite-plus` is indirectly introduced by a dependency of the project through `dependencies` or `peerDependencies`, and it is installed using `npm install`, vite-plus will be hoisted to `node_modules/vite-plus` due to the hoisting mechanism, which will lead us to incorrectly determine the current project as Vite+.

If this check is only executed once during the vscode startup phase, the performance should be acceptable?

rfcs/editor-extension-vite-plus-detection.md
fengmk2: You're right, let me correct the description here.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: > Some users will open only a sub-package of a monorepo

In this case, will there be similar issues with the search for the oxlint bin? When the sub-package does not depend on oxlint, but only the root workspace depends on oxlint.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: The workspace root here is a concept in monorepo, which is different from the vscode workspace. Let me improve the description and change it to `root workspace` would be more appropriate.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: For Vite+ projects, this might not be possible. If the project doesn't have a dependency on `vite-plus`, the oxc IDE extension probably won't recognize it as a Vite+ project, and therefore won't look for the `vp` binary in the global path.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: If the current project is detected to be a Vite+ project, the `vp` binary can be located following the same rules as the `oxlint` binary, which supports Yarn PnP.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: I think `coc-oxc` can also use this package too, it seem to support Node.js runtime https://github.com/oxc-project/coc-oxc/blob/main/src/common.ts

rfcs/editor-extension-vite-plus-detection.md
Sysix: > If this check is only executed once during the vscode startup phase, the performance should be acceptable?

The startup is already effected by the `package.json` search for sub-packages. On a big project this could take over 1 second. 
I can accept this performance cost. I have plans to "delay" to search to not block the extension activation.

rfcs/editor-extension-vite-plus-detection.md
Sysix: Does `require.resolve` not resolves this correctly 🤔 The problem is for `oxlint` too :)

rfcs/editor-extension-vite-plus-detection.md
Sysix: We need to be careful, formatting in editor is very common. It should work every time, if possible.

> IDE extension probably won't recognize it as a Vite+ project

Then it should not find `oxlint` & `oxfmt` too, and fallback to global search. Here we can search for vite-plus first, and then fallback to `oxlint`?

Scenario A:
User opens a project where no formatter or linter is installed. It is an old project where he edits one or two files every year. He still wants the formatting capability of the editor without creating a new setup. 

Scenario B:
User opens a single file, there is no workspace/root concept anymore. The extension should fall back to globally installed packages, to allow users to format the file (with the default config).

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,319 @@
...
+`dependencies` or `devDependencies`. A `node_modules/vite-plus/`
+directory on its own does not qualify — that could be a transitive
+install hoisted from an unrelated dependency tree.
+
+> _"Root workspace"_ in this RFC means the **monorepo root** — the
+> directory containing `pnpm-workspace.yaml`,
+> `package.json#workspaces`, or `lerna.json`. It is **not** the
+> editor's "workspace folder" (the folder a user opens in VS Code,
+> Zed, etc.). The two concepts are distinct: a single editor
+> workspace folder may sit at, inside, or alongside a root workspace.
```
nrayburn-tech, line 48: If a project does not use workspaces, then how do you know when to stop recursing upwards?

nrayburn-tech: I got my answer from the mermaid graph. It seems like the intent is to recurse upwards until the filesystem root.

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,319 @@
...
+
+- **`oxc-vscode`, `coc-oxc`** — target `node_modules/.bin/vp`
+  (`vp.cmd` on Windows), the same shim path they already use for
+  `oxlint`. Call the detector before the existing
+  `findBinary("oxlint" | "oxfmt", ...)` chain. Do **not** parameterize
+  the existing chain with `"vp"` as a target — its
+  `searchSettingsBin`, `searchGlobalNodeModulesBin`, `searchEnvPath`,
+  and `require.resolve` paths can escape the root workspace or
+  consult settings meant for oxlint/oxfmt.
+- **`oxc-zed`** — keep targeting `node_modules/vite-plus/bin/vp` (the
```
nrayburn-tech, line 331: Zed has a very limited extension API currently. I wouldn't expect the Zed extension to be able to implement this RFC as documented. (I'm not suggesting the RFC change because of Zed, but just something to be aware of. That extension API is very limiting today, but I believe it's on their roadmap to eventually improve.)

For example, you can't traverse the file system within a Zed extension directly. I think this alone prevents implementing this RFC.

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,319 @@
...
+Today each extension's idea of "is Vite+" differs — Zed checks
+`package.json` deps and points at the wrapper bin
+(`oxc-zed/src/lsp.rs:28`); IntelliJ has a dedicated
+`VitePlusPackage.kt`; oxc-vscode and coc-oxc have no explicit
+detection. The goal of this RFC is one rule, four implementations.
+
+## The rule
+
+A project is **Vite+** iff some `package.json` between the start
+path and the root workspace declares `vite-plus` directly in
```
nrayburn-tech: > start path

What exactly would be the definition of this? I think it's just the current file open in the editor UI, but I'm not entirely sure. Clarifying this may be helpful.

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,319 @@
...
+`dependencies` or `devDependencies`. A `node_modules/vite-plus/`
+directory on its own does not qualify — that could be a transitive
+install hoisted from an unrelated dependency tree.
```
nrayburn-tech, line 37: (Different thread for a different concern.)

> A `node_modules/vite-plus/`
directory on its own does not qualify — that could be a transitive
install hoisted from an unrelated dependency tree.

This explicitly prevents somebody from wrapping vite-plus in another package. Is that going to be a problem for some subset of users? I don't have a specific use case in mind, but it feels limiting so it may be worth thinking about in case that request does come up from users.

rfcs/editor-extension-vite-plus-detection.md
nrayburn-tech: I don't think this is a one time problem for startup only. I think it has to be performed every time a file is opened (at least, that's a naive but simple approach).

If I open `<repo>/packages/app1/index.ts` then I have to check everything from the `app1` folder all the way up to the system root until I find something matching the required criteria. If I then open `<repo>/packages/app1/something/index.ts` I have to do that same check again from `something` all the way up to the system root until I find something matching the required criteria.
The number of lookups may be mitigated with some more complex strategies (caching, file watchers, etc.).

---

I'm pretty sure the IntelliJ plugin uses the naive approach today. Anytime a file is opened, search upwards until we find what that plugin considers a root.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: Indeed, if it has been clearly determined that the current project is a Vite+ project, then it is necessary to search globally to see if the `vp` bin exists. Let me make a modification.

rfcs/editor-extension-vite-plus-detection.md
```diff
@@ -0,0 +1,350 @@
...
+    )
+
+    return Some({ root, vp_path })  # vp_path may be None
+```
+
+Both phases stop AT the root workspace and never cross into its
+parent. The walk-up bound is what prevents a nested checkout from
+inheriting an unrelated parent's Vite+ install.
+
+### Start path per extension
```
fengmk2, line 93: @nrayburn-tech I will try to write POC code for each IDE extension to verify the correctness of the logic described in the documentation.

rfcs/editor-extension-vite-plus-detection.md
fengmk2: https://github.com/voidzero-dev/vite-plus/pull/1614/changes#r3285677648

fengmk2:
@Sysix @nrayburn-tech I will try to write POC code for each IDE extension to verify the correctness of the logic described in the documentation.

fengmk2:
POC'd this in **oxc-zed** (oxc-project/oxc-zed#199): ported Phases 1–3 to Rust, all 10 fixtures pass, verified live in Zed against a real vite-plus project (`urllib` → launches `vp lint/fmt --lsp`). Three Zed notes for the RFC:

1. **Phase 3 is implementable but low-value in Zed.** The §oxc-zed note says Zed exposes no `$PATH`, but `Worktree::which` has existed since the earliest WIT, so `worktree.which("vp")` resolves the `$PATH` lookup. *However*, after testing it I don't think it's worth implementing for Zed: a real vite-plus project's `vite.config.ts` imports `vite-plus`, so with no local install `vp`, global or otherwise, dies with `UNRESOLVED_IMPORT`. A missing/deleted `node_modules` breaks `vp` regardless of which binary you launch, so the global fallback can't actually rescue a missing install; the local install is what matters. (Global `node_modules` lookup also isn't possible, no process spawn.)

2. **The real blocker: `node_modules` reads are unreliable** ([zed#10760](https://github.com/zed-industries/zed/issues/10760), closed *not planned*), `read_text_file` there "mostly doesn't work" and `Path::exists` fails in WASM. So Phase 2's *"validate against a real `vite-plus` package (`name === "vite-plus"`)"* isn't reliable in Zed; it'd return false "declared but not installed" for real installs. The shipping extension already sidesteps this by reading only the root `package.json` and *trusting* the constructed `node_modules/vite-plus/bin/vp` path; a missing install then fails at spawn time (the upgrade-hint trigger the RFC already describes).

3. **Phase 1's walk is bounded at the worktree root**, `read_text_file` can't read above `worktree.root_path()`, so a monorepo root where only a subpackage declares `vite-plus` returns `null` (consistent with the RFC's "no per-file resolution" note).

Net for oxc-zed: we landed on a **local-only construct-and-trust** design, Phase 1 detects via the root `package.json`, Phase 2 trusts `node_modules/vite-plus/bin/vp`, and there's no Phase 3 global fallback because it can't help a missing install anyway. A missing install surfaces as a clear `Cannot find module …/vite-plus/bin/vp` ("run `pnpm install`").

liangmiQwQ (commented):
> The real blocker: node_modules reads are unreliable (https://github.com/zed-industries/zed/issues/10760, closed not planned)

If we end up finding problem to read these files to determine if it is a Vite+ project, could we give up `detection`, and have an explicit setting for that?

Users can choose whether to use a different version of Oxlint in a project, the encapsulated Vite+ will not be affected.

That said, I think this is a temporary solution, there is room to improve this in the future. But considering the difficulty to detect, and users now already have some explicit settings in the editor to configure Vite+, I believe this is a feasible solution to fix this major bug before RC.

PR #1609, opened 2026-05-17
chore: bump pnpm to 11.1.2

Boshen:
Bumps the repo-root `packageManager` pin to `pnpm@11.1.2`.

Only the root `package.json` `packageManager` field is changed.

PR #1615, opened 2026-05-18
ci(e2e): remove vinext vp check workaround

fengmk2:
## Summary

- vinext upstream main now reflects oxlint 1.63's `eslint-` prefix removal ([oxc-project/oxc#21806](https://github.com/oxc-project/oxc/pull/21806)), so `vp check --fix` and `vp run check` no longer need the `|| true` report-only escape hatch in the ecosystem-ci matrix.
- Bumps the pinned vinext hash (`a1d2b610` → `d558923f`) to pick up the source-side fix.
- After this lands, a regression in `vp check` against vinext will fail the e2e matrix instead of being silently swallowed.

## Test plan

- [ ] E2E Test workflow passes the `vinext` job without `|| true`.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -265,14 +265,8 @@ jobs:
             node-version: 24
             command: |
               vp run build
-              # Report-only: oxlint 1.63 drops the `eslint-` prefix from plugin names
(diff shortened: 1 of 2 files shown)


fengmk2:
required https://github.com/voidzero-dev/vite-plus/pull/1609

fengmk2:
fix in https://github.com/voidzero-dev/vite-plus/pull/1767

Issue #1621, opened 2026-05-18
Do not resolve project-level Node.js version for global pm commands


liangmiQwQ:
### Description

For now, `vp add -g`, `vp install -g`, `vp update -g` use the node version returned by `resolve_version`. which will care about project level's node version like `.node-version` file and even `package.json#engines.node`.

Semantically speaking, would it make more sense to always use global node version (set with `vp env default`)?

### Suggested solution

Skip project-level's node version, the resolve order is like:

1. *`VP_NODE_VERSION` env var (session override from `vp env use`) and `.session-node-version` file (session override written by `vp env use` for shell-wrapper-less environments)* **⬅️ Should we keep this layer? I prefer ignore them actually.**
2. User default from config.json
3. Latest LTS version

### Alternative

N / A

### Additional context

I'm willing to help with it, it may be partially related to #1598

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

Adrian-Baran-GY:
> Semantically speaking, would it make more sense to always use global node version (set with vp `env default`)?

To me current behaviour makes more sense. I belive the resolved node version should be used. I would be quite surprised if `vp install -g` would use different node version than what I would get from checking `node --version` in a given directory.

This is aligned with the behaviour of the current solutions, like `nvm` for example. `npm install -g` won't use the version pointed by default alias but instead the currently selected node version.

liangmiQwQ:
> I would be quite surprised if `vp install -g` would use different node version than what I would get from checking `node --version` in a given directory.

Umm... The main reason why I brought it up is that I don't think Vite+ is only a Node.js version manager in this case, it acts more like a global package manager here. The node.js version used to install a global package will be also used to run the installed global package binary. What's more, commands like `(p)npm install -g`  can still be used which strictly follow the resolved node-version, the same as `nvm`'s behavior.

Adrian-Baran-GY:
In case a user is in a directory where Node.js can be resolved (`.node-version`, `pkg#engines.node`, or `pkg#devEngines.runtime`) and that resolved Node.js version differs from the default one I would at least ask the user whether he wants to use the resolved version for the global installation or the default one.

ywenhao:
[...]

PR #1622, opened 2026-05-18
fix(global): prevent finding package.json for global outdated and why commands

liangmiQwQ:
The current `vp outdated -g` and `vp why -g` use `build_package_manager`, which needs cwd to have a `package.json` file. Semantically speaking, global commands can run without local package.json

Similar to #1621, but for `vp outdated -g` and `vp why -g` commands

🤖 Generated with Codex


liangmiQwQ:
@codex

fengmk2 (approved):

PR #1626, opened 2026-05-19
chore: consolidate build chain into pnpm build script

fengmk2:
Move the upstream prerequisites into the pnpm build script and have the justfile recipe just call pnpm build


justfile
```diff
@@ -42,14 +42,7 @@ _fix_symlinks:
 
 build:
   pnpm install
-  pnpm --filter @rolldown/pluginutils build
```
fengmk2, line 45: something wrong here

Brooooooklyn (approved):

PR #1631, opened 2026-05-19
fix(cli): make env shims use the active vp executable

### LINKED ISSUE #1572: Homebrew installation breaks vp env due to bundled Node.js dependency
### Describe the bug

The current Homebrew installation for vite-plus does not work properly with `vp env`.

## Problem

Installing vite-plus via Homebrew automatically installs node as a dependency. This appears to interfere with the vite-plus node shim resolution and causes vp env to behave incorrectly.

## Expected Behavior

`vp env` should correctly resolve and use the vite-plus managed Node.js environment after installation via Homebrew.

## Actual Behavior

The Homebrew-installed node dependency conflicts with the vite-plus shim setup, resulting in incorrect behavior when running: `node`

### Reproduction

https://formulae.brew.sh/formula/vite-plus

### Steps to reproduce

- Install vite-plus via homebrew, `brew install vite-plus`
- Make sure `vp env` is setting up
- Run `node --version` or `which node` to verify if it's coming from vite-plus.

### System Info

```shell
`vp env current`

Environment:
  Version  24.15.0
  Source   lts

Tool Paths:
  node  /Users/chungwei/.vite-plus/js_runtime/node/24.15.0/bin/node
  npm   /Users/chungwei/.vite-plus/js_runtime/node/24.15.0/bin/npm
  npx   /Users/chungwei/.vite-plus/js_runtime/node/24.15.0/bin/npx


`vp --version`

vp v0.1.21

Local vite-plus:
  vite-plus  Not found

Tools:
  vite             Not found
  rolldown         Not found
  vitest           Not found
  oxfmt            Not found
  oxlint           Not found
  oxlint-tsgolint  Not found
  tsdown           Not found

Environment:
  Package manager  Not found
  Node.js          v24.15.0


`node --version`

v26.0.0
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

leohara:
## Summary

Fix Unix env shims created by `vp env setup` so they point to the active `vp` executable instead of always assuming the standalone `VP_HOME/current/bin/vp` layout.

This keeps the existing standalone behavior when the active executable is `VP_HOME/current/bin/vp`, while making Homebrew-style installs work when `vp` lives outside `VP_HOME`.

Fixes #1572 .

## Details

- Resolve the Unix shim target from the currently running `vp`
- Preserve `../current/bin/vp` for standalone installs
- Recreate stale or broken Unix core shim symlinks without requiring --refresh.
- Add a global snap test that simulates an external/Homebrew-style `vp` and verifies:
  - shims point to the external `vp`
  - `command -v node` resolves to the Vite+ shim
  - `node -v` uses the project Node version

> [!NOTE] 
> This PR intentionally keeps the scope focused on `vp env setup` core shims. I noticed that package binary shims created by `vp install -g` also use the `../current/bin/vp` Unix target today, but I left that out to keep this PR focused on the reported Homebrew `node` shim issue. If you’d prefer fixing those package shims in the same PR, I’m happy to add that follow-up change here.

## Verification

- `cargo test -p vite_global_cli shim -- --nocapture`
- `pnpm bootstrap-cli`
- `pnpm -C packages/cli snap-test-global command-env-setup-external-vp`
- `vp check --fix`
- `cargo clippy -p vite_global_cli --all-targets`
- `cargo fmt --check`
- `git diff --check`


fengmk2:
@cursor review

fengmk2:
@codex review

nekomoyi (approved):

fengmk2 (approved):
@leohara Thanks!

fengmk2:
@leohara It looks like this change would cause the E2E Test to hang?

fengmk2:
> @leohara It looks like this change would cause the E2E Test to hang?

Ignore my reply, it seems that the main branch also has this issue. I'll check why https://github.com/voidzero-dev/vite-plus/actions/runs/26228753482

fengmk2:
e2e fails will be fixed in https://github.com/voidzero-dev/vite-plus/pull/1668

PR #1635, opened 2026-05-19
ci: replace zizmor and cargo-deny with oxc-project/security-action

Boshen:
## Summary

- Remove `.github/workflows/zizmor.yml` and `.github/workflows/deny.yml` workflows
- Remove `deny.toml` configuration
- Add `.github/workflows/security.yml` using `oxc-project/security-action@v1.0.5`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1645, opened 2026-05-20
ci: fix release-day flakes in upgrade test and snap test

fengmk2:
## Summary

Right after a release commit lands on main, the dev build's `package.json` version equals npm latest. Two CI checks were hard-coded to assert outputs that only happen when those versions differ, so they fail on every release-day PR run.

**CLI E2E `Test upgrade (bash/pwsh/cmd)`** (ci.yml) — asserted `UPDATED_VERSION != INITIAL_VERSION` after `vp upgrade --force`. When dev == latest, the upgrade runs end-to-end but the version is unchanged.
- Fix: assert on the `current` install dir basename (`local-dev-<ts>/` → `<version>/`) instead. The dir flip proves the download/extract/swap and rollback flows ran. Node's `fs.realpathSync` resolves both Unix symlinks and Windows junctions, so all three shells share the same helper.

**Snap test `command-upgrade-check`** (`packages/cli/snap-tests-global/command-upgrade-check/`) — recorded the "Update available" output of `vp upgrade --check`. When dev == latest, the actual output is "Already up to date".
- Fix: query the `alpha` dist-tag instead. Alpha is always a different version from latest (currently `0.1.21-alpha.7`), so the "Update available" branch is always taken. Semver normalization masks the actual version, so the snap body is unchanged.

## Failure examples

- Upgrade test: https://github.com/voidzero-dev/vite-plus/actions/runs/26112690623/job/76915605617
- Snap test: https://github.com/voidzero-dev/vite-plus/actions/runs/26148842387/job/76910949450 (also blocks #1635)

## Test plan

- [ ] CI passes on this PR.
- [ ] After next release, both checks still pass on the post-release-day merge-to-main run.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only affect CI assertions and snapshot test inputs, not production code. Main risk is masking a real regression if install directory swapping semantics change.
> 
> **Overview**
> Makes the CI `vp upgrade` E2E checks resilient to release-day runs by **asserting on the `.vite-plus/current` install directory target** (via `fs.realpathSync`) instead of comparing CLI versions, across bash/pwsh/cmd.
> 
> Stabilizes the `command-upgrade-check` snapshot by running `vp upgrade --check` against the `alpha` dist-tag so the *"Update available"* branch is consistently exercised.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4556aecaadf39c6a65beaabf1d0680232bf8b1a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,11 +207,6 @@ jobs:
           - os: windows-latest
             target: x86_64-pc-windows-msvc
     runs-on: ${{ matrix.os }}
-    env:
-      # Any released version older than the current package.json version.
-      # Used by upgrade tests to create a guaranteed version delta against npm
-      # latest (otherwise `vp upgrade --force` is a no-op on release-day mains).
-      UPGRADE_TEST_PIN_VERSION: '0.1.21'
     steps:
       - uses: taiki-e/checkout-action@7d1e50e93dc4fb3bba58f85018fadf77898aee8b # v1.4.2
       - uses: ./.github/actions/clone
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -424,9 +429,18 @@ jobs:
             node -p "require(require('path').resolve(process.env.USERPROFILE || process.env.HOME, '.vite-plus', 'current', 'node_modules', 'vite-plus', 'package.json')).version"
           }
 
-          # Save initial (dev build) version
+          # Pin to a known older version first so the upgrade-to-latest below
+          # has a version delta to assert against. Right after a release commit
+          # on main, the dev build's package.json version equals npm latest, so
+          # `vp upgrade --force` would be a no-op and the assertion would fail.
+          vp upgrade "$UPGRADE_TEST_PIN_VERSION" --force
```
fengmk2: That won't work this way. We need to test the dev version of the `vp upgrade` logic.

fengmk2:
@cursor review

Boshen (approved):

PR #1647, opened 2026-05-20
fix(cli): add managed bin to hook path

TheAlexLichter:
When committing via VS Code on MacOS I hit the error

```bash
./node_modules/.bin/vp: line 20: exec: node: not found
VITE+ - pre-commit script failed (code 127)
VITE+ - command not found in PATH=./node_modules/.bin:/Library/Developer/CommandLineTools/usr/libexec/git-core:/usr/bin:/bin:/usr/sbin:/sbin
```

But node was installed and `vp env doctor` was happy.
The issues seemed that the managed Node is not available in the Git hook.

I learned that Git hooks often run with a much smaller, non-interactive `PATH`, possibly not including the shims.
In my case, path was limited to what is shared in the error above. That included the local `vp` bin but not `node`, leading to the error.

The PR fixes that by making generated hooks add Vite+’s managed bin directory before running the user hook. Then `vp staged` can start even when Git launches the hook with a restricted PATH.

### DIFF
--- a/packages/cli/src/config/hooks.ts
+++ b/packages/cli/src/config/hooks.ts
@@ -49,6 +49,10 @@ i="\${XDG_CONFIG_HOME:-$HOME/.config}/vite-plus/hooks-init.sh"
 
 { [ "\${HUSKY-}" = "0" ] || [ "\${VITE_GIT_HOOKS-}" = "0" ]; } && exit 0
 
+d=${rootExpr}
+__vp_shell=/bin/sh
+[ -x "$__vp_shell" ] || __vp_shell=$(command -v sh)
+
 if [ -n "\${VP_HOME-}" ]; then
   __vp_bin="$VP_HOME/bin"
 elif [ -n "\${HOME-}" ]; then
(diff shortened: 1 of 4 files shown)


fengmk2:
@codex review

fengmk2 (approved):
@TheAlexLichter Thanks! 💯

packages/cli/src/config/hooks.ts
```diff
@@ -49,6 +49,15 @@ i="\${XDG_CONFIG_HOME:-$HOME/.config}/vite-plus/hooks-init.sh"
...
 { [ "\${HUSKY-}" = "0" ] || [ "\${VITE_GIT_HOOKS-}" = "0" ]; } && exit 0
 
+if [ -n "\${VP_HOME-}" ]; then
+  __vp_bin="$VP_HOME/bin"
+elif [ -n "\${HOME-}" ]; then
+  __vp_bin="$HOME/.vite-plus/bin"
+else
+  __vp_bin=""
+fi
+[ -n "$__vp_bin" ] && [ -d "$__vp_bin" ] && export PATH="$__vp_bin:$PATH"
```
fengmk2: `export PATH="$PATH:$__vp_bin"` Put at the end as a fallback, to avoid overriding the system's PATH priority.

fengmk2 (approved):

PR #1648, opened 2026-05-21
ci: add codiff to ecosystem-ci

fengmk2:
## Summary

- Add [nkzw-tech/codiff](https://github.com/nkzw-tech/codiff) (an Electron-based fast local diff viewer) to ecosystem-ci.
- Runs `vp run test:all` (which expands to `vp check && vp test` via the task defined in codiff's `vite.config.ts`) on ubuntu-latest only — matching upstream's `test.yml` workflow.
- Codiff already depends on `vite-plus` (via catalog), so `forceFreshMigration: true` is set so `vp migrate` performs a full dependency rewrite rather than skipping.

## Test plan

- [ ] e2e-test workflow passes for the new `codiff` matrix entry on ubuntu-latest

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -337,6 +337,9 @@ jobs:
           - name: codiff
             node-version: 24
             command: |
+              # vp migrate rewrites package.json without reformatting,
+              # so format first before running the project's `vp check && vp test` task.
+              vp fmt
               vp run test:all
         exclude:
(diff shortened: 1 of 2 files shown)


cpojer (approved):
I don't mind but does Codiff do anything worth testing that's not already covered elsewhere?

fengmk2:
I don't think so, but I want to add an Electron app as a test case, and I just saw codiff.🤣

fengmk2:
move to https://github.com/voidzero-dev/vite-plus/pull/2551

PR #1655, opened 2026-05-21
chore(build): drop `@rolldown/pluginutils` build step

shulaoda:
## Summary
- `@rolldown/pluginutils` has been extracted from the rolldown repo into a standalone package (rolldown/rolldown#9317), so it is no longer a workspace member.
- Remove the `vp run @rolldown/pluginutils#build` step from the `build:src` task in `vite.config.ts` to keep it aligned with the root `pnpm build` script.


fengmk2 (approved):
@shulaoda Thanks!

PR #1687, opened 2026-05-26
fix: use fmt.configPath for Zed oxfmt settings

chungweileong94:
According to the [Zed OXC extension example](https://github.com/oxc-project/oxc-zed/blob/main/examples/oxfmt/.zed/settings.json), the `configPath` should be `fmt.configPath` instead, and I verified it myself.


fengmk2 (approved):

fengmk2 (approved):
@chungweileong94 Thanks!

PR #739, opened 2026-03-09
refactor: move typeAware/typeCheck from CLI flags to vite.config.ts

fengmk2:
oxlint natively reads `options.typeAware` and `options.typeCheck` from
its config file, so there's no need for `vp check` to inject these as
CLI flags when calling `vp lint`.

- Remove `--no-type-aware` / `--no-type-check` flags from `vp check`
- Add `options: { typeAware: true, typeCheck: true }` defaults in
  `vp lint --init`, `vp migrate`, and project root vite.config.ts
- Strip `--type-aware` / `--type-check` from migrated scripts
- Remove `oxc.typeAware` from VS Code settings (extension reads from
  config automatically)
- Add `OxlintOptions` type with `typeAware` and `typeCheck` fields

PR #1688, opened 2026-05-26
fix(cli): parse tsconfig as jsonc

TheAlexLichter:
If tsconfig has comments, JSON.parse fails and we can't detect nor remove baseUrl!

This PR resolves this.

Related: https://github.com/voidzero-dev/vite-plus/pull/739


fengmk2 (approved):
@TheAlexLichter 👍

PR #1697, opened 2026-05-27
fix(create): normalize sub-package vite-plus and drop aliased vite/vitest

fengmk2:
## Summary

Two clean-up fixes for `vp create vite:monorepo`:

- **`packages/utils/package.json` keeps `\"vite-plus\": \"^x.y.z\"`** inherited from the remote library template (`github:sxzz/tsdown-templates/vite-plus`). The per-package migrator's catalog rewrite is gated on `needVitePlus`, which only flips true when `vite`/`vitest`/`oxlint`/`oxfmt`/`tsdown`/`@vitest/browser*` is present — none of which the library template has. So `apps/website` ended up with `vite-plus: catalog:` while `packages/utils` stayed pinned, leaving sub-packages drifted.

  Fix: in `rewritePackageJson`, also normalize an *existing* `vite-plus` value to the canonical spec (`catalog:` in catalog-supporting monorepos, or `VITE_PLUS_VERSION` under force-override). Scoped to catalog-supporting PMs so `vp migrate` on npm/standalone doesn't silently overwrite a user's pinned version.

- **`apps/website/package.json` retains aliased `vite` / `vitest` devDependencies** left by `create-vite vanilla-ts`. The migrator only rewrites their *value* (e.g. `vite: catalog:`), it doesn't drop the keys. But in a freshly-scaffolded project the CLI scripts have already been rewritten to `vp ...` and no user file imports `'vite'` / `'vitest'`, so `vite-plus` provides them transitively. The dangling keys are dead weight.

  Fix: in `executeMonorepoTemplate`, after the per-package migrator runs on `apps/website`, strip the `vite` and `vitest` keys from devDependencies. (Only in `vp create`, not `vp migrate`.)

## CI coverage

New monorepo-only assertion step in `.github/workflows/test-vp-create.yml` runs across all 4 package managers and verifies:

- `apps/website` devDependencies contain neither `vite` nor `vitest`
- both `apps/website` and `packages/utils` declare `vite-plus`
- for pnpm/yarn/bun, the `vite-plus` spec matches across both sub-packages (npm is skipped since its non-catalog `vp migrate` path keeps user pins untouched)

## Test plan

- [x] `vp test run packages/cli` — 407 passed (added 2 unit cases covering the pnpm-normalize and npm-leave-alone paths)
- [ ] CI: `test-vp-create` matrix (pnpm/npm/yarn/bun × monorepo template) passes the new assertion step

---

> [!NOTE]
> **Medium Risk**
> Touches package.json rewrite logic used by migration and create; behavior is scoped (catalog PMs, create-only strip) and covered by new tests and e2e assertions.
> 
> **Overview**
> Fixes **`vp create vite:monorepo`** so sub-packages don’t end up with inconsistent or redundant dependency entries.
> 
> **Migrator (`rewritePackageJson`):** When a package already declares `vite-plus` but nothing else triggers migration, catalog-supporting monorepos (pnpm/yarn/bun) now rewrite plain version ranges to the canonical `catalog:` (or force-override path). Protocol-pinned specs (`workspace:*`, `catalog:named`, etc.) are left alone; **npm** monorepos still keep user pins. Auto-adding **vitest** for vitest-adjacent peers is limited to real migration passes, not this normalize-only path.
> 
> **Monorepo create:** After migrating `apps/website`, **`vite` and `vitest`** are removed from that app’s devDependencies (scripts already use `vp`; runtime comes via `vite-plus`).
> 
> **CI & tests:** `test-vp-create` gains a monorepo-only check for aligned `vite-plus`, no stray `vite`/`vitest` on the website app, and updated bun snap output; unit tests cover normalize vs npm-leave-alone and protocol preservation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0132c60e08e0eb44ef9bde44e5cfeb18f6493ef6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fengmk2:
@cursor review

cpojer (approved):

PR #1698, opened 2026-05-27
fix(global): install packages into final prefix

### LINKED ISSUE #664: `vp install -g ./some-local-package` not working
Needs support for installing local npm packages to the global bin directory, similar to pnpm.

liangmiQwQ:
As https://github.com/voidzero-dev/vite-plus/issues/664#issuecomment-4544649734 said, #1685 did not completely fix #664.

The current global package install flow runs `npm install -g` in a temporary prefix and then moves the installed package prefix into `VP_HOME/packages`. 

However, `npm` and some packages's `postinstall` script may create files that links an absolute path to this temporary dir, or a relative path use the temp dir as the base dir, they will be invalid after moving, causing bugs like that.

```
$ vp install -g .

info: Installing 1 global package with Node.js 25.0.0
✓ Installed just-a-normal-package 0.0.0
  Bins: just-a-normal-package

tip: Available short aliases: i = install, rm = remove, un = uninstall, up = update, ls = list, ln = link

$ just-a-normal-package
vp: Binary 'just-a-normal-package' not found: Package just-a-normal-package not found
```

~~(It is quite strange that it can't be reproduced on local snap tests. I am still figuring out the reason now.)~~
**UPDATE: please refer to the comment below**

So to fix this problem, this PR installs global packages directly into their final package prefix to avoid moving, to fix this problem.

There is another runtime behavior change worth mentioning: If one package in a multi-package install fails, this PR still finalizes shims for packages that already installed successfully instead of deleting them or leaving them without runnable bins. This is because if users are trying to reinstall a package or update, uninstall the installed packages can make problems. This problem only happens after the changes mentioned above, and it also follows `npm`'s behavior. 

🤖 Generated with Codex.


liangmiQwQ:
> (It is quite strange that it can't be reproduced on local snap tests. I am still figuring out the reason now.)

I've found the causes with the help of Codex. It is a little bit complex and related to the current snaptest architecture.

Vite+ calls `npm` to install a package into a temp prefix, and move to the final directory.

```bash
# The example cwd: ~

# Vite+'s temp directory
~/.vite-plus/tmp/packages/just-a-normal-package

tree ~/.vite-plus/tmp/packages/just-a-normal-package/lib/node_modules/
#~/.vite-plus/tmp/packages/just-a-normal-package/lib/node_modules/
#└── just-a-normal-package -> ../../../../../../
# This is linked to ~, 6 layers of parent directory (`../`)

# and if it is moved to ~/.vite-plus/packages/just-a-normal-package (not ~/.vite-plus/tmp)
tree ~/.vite-plus/packages/just-a-normal-package/lib/node_modules/
#~/.vite-plus/packages/just-a-normal-package/lib/node_modules/
#└── just-a-normal-package -> ../../../../../../
# It is still  ../../../../../../ after moving (6 layers of `../`)
# So it will be linked to `~/../`, like `/Users`, instead of `~`. (We only need 5 layers of `../` there), make packages broken.
```

However, the snap test runner doesn't use `~` as the cwd. Instead, it copies the fixture to a random temp directory under `/private/var`, I will use `/private` as example.

```bash
# The example cwd: /private ⚠️

# Vite+'s temp directory
~/.vite-plus/tmp/packages/just-a-normal-package

tree ~/.vite-plus/tmp/packages/just-a-normal-package/lib/node_modules/
#~/.vite-plus/tmp/packages/just-a-normal-package/lib/node_modules/
#└── just-a-normal-package -> ../../../../../../../../private
# This is linked to ~, 8 layers of `../`

# and if it is moved to ~/.vite-plus/packages/just-a-normal-package (not ~/.vite-plus/tmp)
tree ~/.vite-plus/packages/just-a-normal-package/lib/node_modules/
#~/.vite-plus/packages/just-a-normal-package/lib/node_modules/
#└── just-a-normal-package -> ../../../../../../../../private
# It is still  ../../../../../../ after moving (still 8 layers of `../`)
# 👉 Different from the `~` case, one more `../` won't break the path because it hits the root `/` already.
# So it will be still linked to `/private`.
```

This logic makes snap tests invalid for this fixes.

@fengmk2, do you have any good idea to add validation tests for this fix?

liangmiQwQ:
@codex

fengmk2:
@liangmiQwQ If it's too much trouble, let's skip adding a snap test for this special case for now.

liangmiQwQ:
@codex

liangmiQwQ:
@codex

packages/cli/snap-tests-global/command-env-install-fail/steps.json
```diff
@@ -1,5 +1,6 @@
 {
   "env": {},
   "ignoredPlatforms": ["win32"],
-  "commands": ["vp install -g voidzero-nonexistent-pkg-xyz-12345 # Install non-existent package"]
+  "commands": ["vp install -g . voidzero-nonexistent-pkg-xyz-12345", "install-fail-local-package"],
```
fengmk2: Can we create a new snap test to verify this? Reusing this failed snap test seems a bit odd.

fengmk2 (approved):
LGTM

packages/cli/snap-tests-global/command-env-install-fail/steps.json
liangmiQwQ: Adjusted in 2ed25fdaf8d9bcba715db4e4f46b221d957503a3

PR #1708, opened 2026-05-29
docs: update troubleshooting.md

### LINKED ISSUE #930: lint,fmt: Known vite.config.ts releated issues
Currently, `vp lint`, `vp fmt`, and the VSCode extension all transparently utilize Oxlint and Oxfmt respectively.

In Oxapps side, the ability to extract `.lint` and `.fmt` fields from `vite.config.ts` has been added, but they are not fully complete at present.

### 1: All Vite config patterns are not supported

> [!NOTE]
> This is fixed in oxlint@1.67.0 and oxfmt@0.52.0 = vite-plus@0.1.23

<details>

> https://vite.dev/config/

Currently supported static config only:

```ts
export default { ... };
export default defineConfig({ ... });
```

Conditional and async config are NOT supported.

```ts
defineConfig(({ ... }) => ({ ... }));
defineConfig(async ({ ... }) => ({ ... }));
```

Since they do not know how to handle arguments.

For the first place, Oxlint and Oxfmt do not support functional config even with their `oxlint.config.ts`. Also, be aware that the file name, they only support the `.ts` extension.

</details>

### 2: Some valid Vite config fails to load even with static config

> [!NOTE]
> This is fixed in oxlint@1.67.0 and oxfmt@0.52.0 = vite-plus@0.1.23

<details>

Most reported cases fall into this category.

Oxlint and Oxfmt uses normal ESM dynamic `import()` to load `vite.config.ts`.

This means it is roughly equivalent to `--configLoader native` for Vite, so if configuration files rely on the default `--configLoader bundle` (also `runner`), which requires some *transform*, it will fail to load.

e.g.

- Use `__dirname` in `type: module` project
- Try to `import X from "./foo.js"`, but actually there is only `foo.ts` on the disk
- Try to `import "foo.json` without `with { type: "json" }`
- etc...

Refs: https://github.com/vitejs/vite/issues/21546

</details>

### 3: VSCode extension does not pick vite.config.ts

<details open>

If you have multiple workspaces open, please make sure that the correct version of Oxlint/Oxfmt is being used to your dedicated workspace.

Currently, single extension and LSP process is shared across multiple workspace folders.

If an unintended project's LSP is selected, it may not have the capability to read `vite.config.ts`.

Refs: https://github.com/oxc-project/oxc-vscode/issues/145

</details>

### 4: vite.config.ts is not used if oxapps config file also exists

> [!NOTE]
> This is fixed in oxlint@1.60.0 and oxfmt@0.45.0

<details>

With current implementation, vite.config.ts is used as fallback.

The priority order is: the respective config files (such as `.oxfmtrc.json`) and then `vite.config.ts`.

I don't think it's recommended to combine these together in a Vite+ environment, but that’s how things stand at the moment.

</details>

### 5: Oxlint|Oxfmt fails even in non-Vite+ project

> [!NOTE]
> This is fixed in oxlint@1.60.0 and oxfmt@0.45.0

<details>

It's not directly related to Vite+, but it's a noteworthy point.

Oxlint and Oxfmt might run into problems if they find `vite.config.ts` (which triggers 2nd issue) while searching for config files by finding-up the parent directories. Just by attempting to load the relevant `.lint|fmt` field causes error.

This may happen to Oxfmt CLI+LSP, and Oxlint LSP usecases.

</details>

### 6: Nested config is not used by the `vp lint|fmt`, but it is used by the LSP

<details open>

Currently, `vp lint|fmt` always executes with `--config` flag with the `vite.config.ts` located at the project root.

This is because, as a recommendation for Vite+, we want to keep all `lint|fmt` settings contained within the root. (And use `overrides` for each package style)

However, it is not possible to enforce the same specification for an LSP (since it is unknown which directory the workspace will be opened from).

Therefore, if nested `vite.config.ts` files are placed and they contain `.lint|fmt` fields, different options will be used.

We are considering adding a built-in Oxlint rule to detect and prevent this.

Refs: https://github.com/voidzero-dev/vite-plus/issues/1456

(Ideally, `--config` is not needed for `vp lint|fmt`. It will auto-discover root config at any cwd. But perf-wise, since `-c` effectively disables nested config discovery. Or just adding `--disable-nested-config` is fine too. But still, LSP does not have equivalent option yet.)

</details>

leaysgur:
Fixes #930

`vite.config.ts` related issues should be fixed by updating oxlint+oxfmt.

Rest issues are already covered by indivisual issue.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,7 +50,7 @@ vite-plus/
 - **Static `vite.config.ts` extraction**: start at `crates/vite_static_config/README.md` and `packages/cli/src/resolve-vite-config.ts`.
 - **Bundled toolchain surfaces**: start with `packages/core/BUNDLING.md`, `packages/cli/BUNDLING.md`, and `packages/test/BUNDLING.md`.
 - **Generated project agent guidance**: `packages/cli/AGENTS.md` and `packages/cli/src/utils/agent.ts`; do not edit these when the task is only to improve root repo guidance.
-- **Product/repo docs**: root contributor docs live at the repo root and `docs/`; CLI package docs live under `packages/cli/docs/`; generated agent guidance is separate.
+- **Product/repo docs**: root contributor docs live at the repo root and the VitePress site under `docs/` (`docs/guide/`, `docs/config/`); generated agent guidance is separate.
 - **CLI output behavior**: inspect the relevant code plus `packages/cli/snap-tests/` or `packages/cli/snap-tests-global/`.
 
 ## Command and Config Model
@@ -79,7 +79,7 @@ Important distinctions:
 - Do not introduce `vite-task.json`; current Vite+ task configuration lives under `run` in `vite.config.ts`.
 - Do not run `cargo test -p vite_task` in this repo; Vite Task crates are git dependencies, not local workspace members.
 
-Reference: `packages/cli/docs/guide/run.md` and `packages/cli/docs/config/run.md`.
+Reference: `docs/guide/run.md` and `docs/config/run.md`.
 
 ## Development Workflow
 
@@ -184,8 +184,8 @@ Use the validation matrix above as the source of truth. For behavior-bearing cha
 - Root scripts: `package.json`
 - Repo config: `vite.config.ts`
 - Vite+ guide: `docs/guide/index.md`
-- Run guide: `packages/cli/docs/guide/run.md`
-- Run config: `packages/cli/docs/config/run.md`
+- Run guide: `docs/guide/run.md`
+- Run config: `docs/config/run.md`
 - CLI package architecture: `packages/cli/BUNDLING.md`
 - Core package architecture: `packages/core/BUNDLING.md`
 - Test package architecture: `packages/test/BUNDLING.md`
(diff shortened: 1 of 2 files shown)


leaysgur:
* **#1708** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1707** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/voidzero-dev?ref=merge-queue-instructions-in-stack-comment&prId=1708).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

fengmk2 (approved):

PR #1714, opened 2026-05-30
fix: avoid locked exe on forced upgrade

fengmk2:
## Summary

- Install forced reinstalls of the active version into a unique semver build-metadata directory
- Repoint `current` after the install completes so Windows does not overwrite the running `vp.exe`
- Add focused unit coverage for forced reinstall directory selection

## Directory structure

Before this change, `vp upgrade --force` reused the active version directory when the target version matched the current version:

```text
~/.vite-plus/
├── current -> 0.1.23
└── 0.1.23/
    └── bin/
        └── vp.exe   # currently running on Windows
```

That meant the forced reinstall extracted the downloaded platform package back into `0.1.23/bin/vp.exe`, which fails on Windows because the running executable is locked.

After this change, a forced reinstall of the active version writes into a new semver-compatible build-metadata directory, then repoints `current` only after installation succeeds:

```text
~/.vite-plus/
├── current -> 0.1.23+force.<pid>.<timestamp>
├── 0.1.23/
│   └── bin/
│       └── vp.exe   # previous active install remains untouched
└── 0.1.23+force.<pid>.<timestamp>/
    └── bin/
        └── vp.exe   # freshly installed binary
```

Normal upgrades to a different version still use the plain target version directory, for example `0.1.24/`.

## Verification

- Reproduced the original Windows `vp upgrade --force` failure with `os error 32`
- `cargo build -p vite_global_cli`
- Temp `VP_HOME` end-to-end locked-executable test: `upgrade --force` succeeds
- `cargo test -p vite_global_cli commands::upgrade::tests`
- `cargo fmt --check`

---

> [!NOTE]
> **Medium Risk**
> Touches the global CLI upgrade/install swap path and directory cleanup; wrong naming or cleanup could break upgrades or rollbacks, but behavior is scoped to forced same-version reinstalls with new unit and Windows E2E coverage.
> 
> **Overview**
> Fixes **Windows `vp upgrade --force`** when the target version is already active: the upgrade no longer extracts into the running install folder (which locks `vp.exe` and caused os error 32).
> 
> **Upgrade flow** now picks an install directory name via `target_install_dir_name`. For a **forced reinstall of the active version**, it uses a unique semver build-metadata path like `{version}+force.{pid}.{nanos}`; otherwise it still uses the plain version folder. **`current`**, rollback metadata, and old-version cleanup follow that **directory name** (not only the semver string).
> 
> **CI** adds a Windows-only E2E step that simulates a locked same-version install under a temp `VP_HOME` and asserts `current` moves to a `+force.` dir, the original version tree stays, and `.previous-version` is correct.
> 
> **Tests** cover forced-dir selection in `upgrade` and that `cleanup_old_versions` treats `+force` dirs like other semver installs (removable when old, kept when protected).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc5393e8ee75d546c61a48ae2cba121025db0a8a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fengmk2:
@cursor review

cpojer (approved):

PR #1720, opened 2026-05-30
fix(cli): remove global why flag

### LINKED ISSUE #1656: `vp why -g` and `vp outdated -g` should check global packages managed by Vite+
### Describe the bug

Ideally, all global pm commands in Vite+ like `vp outdated -g`, `vp update -g`, `vp install -g`, `vp uninstall -g` should be used to control packages inside `~/.vite-plus/packages`.

However, `vp outdated -g` and `vp why -g` here just delegate the command to `npm` with a `--global` field, makes it show packages managed by the corresponding pm, not Vite+'s global packages.

It needs Vite+ to implement `vp outdated -g` and `vp why -g` here. 

### Reproduction

https://github.com/liang-demos/vp-global-outdated.git

### Steps to reproduce

```bash
vp outdated -g
```

### Validations

liangmiQwQ:
Close #1656

The current `vp why -g` behavior delegates to the underlying package manager global mode, which does not inspect Vite+-managed global packages.

This PR removes the `-g`/`--global` option from `vp why`, keeps `vp why` project-scoped, and updates the global CLI help snapshot. `vp outdated -g` remains handled by the Vite+-managed global package flow.

🤖 Generated with Codex


liangmiQwQ:
Should it be a breaking change?

liangmiQwQ:
@codex

fengmk2:
> Should it be a breaking change?

This command has never worked, so it probably doesn't count. 😄

fengmk2 (approved):

PR #1726, opened 2026-05-31
fix(cli): remove windows current link without rmdir

TheAlexLichter:
I've hit a powershell issue while installing Vite+ after it hang in the git bash:

```sh
cmd : The directory is not empty.
  At line:589 char:9
  +         cmd /c rmdir "$CurrentLink" 2>$null
  +         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      + CategoryInfo          : NotSpecified: (The directory is not empty.:String) [], RemoteException
      + FullyQualifiedErrorId : NativeCommandError
```

To fix this, the installer now removes the existing `current` via Powershell logic and detects junctions or symlinks, as well as handling stale directories.

### DIFF
--- a/packages/cli/install.ps1
+++ b/packages/cli/install.ps1
@@ -262,6 +262,33 @@ function Cleanup-OldVersions {
     }
 }
 
+function Remove-CurrentLink {
+    param([string]$Path)
+
+    try {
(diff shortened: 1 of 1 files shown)


fengmk2 (approved):
@TheAlexLichter Nice!

PR #1728, opened 2026-06-01
fix(create): keep aliased vite in pnpm monorepo website so override stays effective

fengmk2:
## Summary

In a monorepo created via `vp create vite:monorepo` with **pnpm**, running `vp why vite` reported the override as ineffective: it resolved `vite` to upstream `vite` (a transitive dependency of `@voidzero-dev/vite-plus-core`) rather than to `@voidzero-dev/vite-plus-core` itself.

## Root cause

PR #1697 added `dropAliasedRuntimeDevDeps`, which strips `vite`/`vitest` from `apps/website/package.json` for every package manager. For pnpm this removed the only direct `vite` dependency in a fresh workspace, so the `pnpm-workspace.yaml` `overrides.vite: catalog:` entry had no consumer to redirect. The only `vite` left in the tree was vite-plus-core's own internal `vite`, so `vp why vite` showed upstream vite and the override looked ineffective. The single-package app keeps `vite: catalog:` and resolves correctly.

## npm / yarn / bun: confirmed not affected (no fix needed)

I verified their override is effective in both single-app and monorepo: `node_modules/vite` resolves to `@voidzero-dev/vite-plus-core`, and `<pm> why @voidzero-dev/vite-plus-core` shows it redirecting even vite-plus-test's peer `vite`. Their root `overrides`/`resolutions` redirect the transitive/peer `vite` regardless of a direct dependency. Their `vp why vite` display gap is a pre-existing limitation of `npm explain` / `yarn why` / `bun why` (they query by resolved package name, not the `vite` alias key), identical with or without the monorepo strip, so it is out of scope here.

## Fix

`dropAliasedRuntimeDevDeps` now returns early for pnpm, keeping the aliased `vite`/`vitest` so the workspace override has a direct consumer. npm/yarn/bun keep stripping the dead-weight keys.

## Before / after (`vp why vite` in a fresh pnpm monorepo)

Before:

```
vite@8.0.14
└─┬ @voidzero-dev/vite-plus-test@0.1.23
  └─┬ vite-plus@0.1.23
    ...
Found 1 version of vite
```

After:

```
@voidzero-dev/vite-plus-core@0.1.23
├─┬ @voidzero-dev/vite-plus-test@0.1.23
│ └─┬ vite-plus@0.1.23
│   └── website@0.0.0 (devDependencies)
...
```

## Verification

[...]

PR #1736, opened 2026-06-01
feat(lint): add pnpm vite alias oxlint rule

fengmk2:
## Summary

- Add `vite-plus/require-pnpm-vite-alias` to the Vite+ Oxlint plugin.
- Enable the new rule in generated/default Vite+ lint configs.
- Update generated-config snapshots to include the new default rule.

## Why

PR #1728 fixed a pnpm monorepo issue where removing the direct aliased `vite` dependency from an application package made the workspace override appear ineffective. This lint rule catches that class of configuration drift in Vite+ application packages.

## Validation

- `pnpm exec tsc --noEmit --pretty false`
- `vp fmt --check packages/cli/src/oxlint-plugin.ts packages/cli/src/oxlint-plugin-config.ts packages/cli/src/__tests__/oxlint-plugin.spec.ts`
- `vp lint packages/cli/src/oxlint-plugin.ts packages/cli/src/oxlint-plugin-config.ts packages/cli/src/__tests__/oxlint-plugin.spec.ts`
- `vp test packages/cli/src/__tests__/oxlint-plugin.spec.ts`
- `git diff --check`

Note: `pnpm -F vite-plus snap-test-global new-vite-monorepo` was skipped on Windows by the snap-test runner.

### DIFF
--- a/packages/cli/snap-tests/lint-pnpm-vite-alias/snap.txt
+++ b/packages/cli/snap-tests/lint-pnpm-vite-alias/snap.txt
@@ -1,0 +1,8 @@
+[1]> vp lint --fix apps/website/vite.config.ts # should fail when a pnpm Vite+ app is missing the direct vite alias
+apps/website/vite.config.ts:1:1: error vite-plus(require-pnpm-vite-alias): pnpm Vite+ application packages must keep a direct 'vite' dependency so the workspace override resolves to @voidzero-dev/vite-plus-core.
+
+> vp lint --fix apps/website-with-vite/vite.config.ts && echo 'website-with-vite passed' # should pass when the pnpm Vite+ app keeps the direct vite alias
+website-with-vite passed
+
+> vp lint --fix packages/utils/vite.config.ts && echo 'utils passed' # should ignore non-application packages
+utils passed
--- a/packages/cli/src/oxlint-plugin.ts
+++ b/packages/cli/src/oxlint-plugin.ts
@@ -1,139 +1,19 @@
-import { definePlugin, defineRule } from '@oxlint/plugins';
-import type { Context, ESTree } from '@oxlint/plugins';
+import { definePlugin } from '@oxlint/plugins';
 
 import {
   PREFER_VITE_PLUS_IMPORTS_RULE_NAME,
+  REQUIRE_PNPM_VITE_ALIAS_RULE_NAME,
   VITE_PLUS_OXLINT_PLUGIN_NAME,
 } from './oxlint-plugin-config.ts';
-
-function rewriteVitePlusImportSpecifier(specifier: string): string | null {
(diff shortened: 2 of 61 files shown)


packages/cli/src/oxlint-plugin.ts
```diff
@@ -135,14 +263,49 @@ export const preferVitePlusImportsRule = defineRule({
...
 });
 
+export const requirePnpmViteAliasRule = defineRule({
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Require pnpm Vite+ application packages to keep a direct vite alias dependency.',
+      recommended: true,
+      url: 'https://github.com/voidzero-dev/vite-plus/pull/1728',
```
fengmk2: We need to create a document that specifically describes the role of the oxlint rule, so that links can point to this document.

packages/cli/snap-tests/lint-pnpm-vite-alias/snap.txt
```diff
@@ -0,0 +1,21 @@
+[1]> vp lint --format default apps/website/vite.config.ts # should fail when a pnpm Vite+ app is missing the direct vite alias
```
fengmk2: `vp lint --fix` should be better?

packages/cli/src/oxlint-plugin/rules/require-pnpm-vite-alias.ts
```diff
@@ -0,0 +1,166 @@
...
+    return /\boverrides\s*:/.test(previousText) && catalogViteAliasesCore;
+  });
+}
+
+export function shouldRequirePnpmViteAlias(filename: string): boolean {
+  if (!isViteConfigFile(filename)) {
+    return false;
+  }
+
+  const packageJsonPath = path.join(path.dirname(filename), 'package.json');
```
fengmk2, line 108: @camc314 Does oxlint plan to support json files? We currently have two requirements for json configuration checks, both related to rule judgments for json files. This PR is one of them, and the other one is https://github.com/voidzero-dev/vite-plus/issues/1456.

camc314: > Does oxlint plan to support json files? 

yes - it should work via a plugin once https://github.com/oxc-project/oxc/discussions/21936 has shipped

camc314: in the interim, maybe only check the json files when linting `vite.config.ts`?

fengmk2: Yes, this is how it is currently implemented, but the error message points to the vite.config.ts file, which is not user-friendly.

PR #1743, opened 2026-06-02
fix(deps): resolve docs dependabot advisories

fengmk2:
Resolves the open Dependabot advisories in the documentation site (`docs/` is a separate pnpm workspace, `docs/pnpm-lock.yaml`).

Most fixes land through a normal lockfile refresh, since the patched versions already fall within the ranges their parents declare. Only `lodash-es` needs an override: `chevrotain` (via `mermaid`) pins it to exactly `4.17.23`, so the patched `4.18.x` can't be reached any other way.

| Package | Resolved | How | Advisories |
| --- | --- | --- | --- |
| `vite` | 7.3.5 | lockfile | GHSA-v2wj-q39q-566r, GHSA-p9ff-h696-f583 |
| `defu` | 6.1.7 | lockfile | GHSA-737v-mqg7-c878 |
| `picomatch` | 4.0.4 | lockfile | GHSA-c2c7-rcm5-vvqj |
| `lodash-es` | 4.18.1 | override | GHSA-r5fr-rjxr-66jc |

## Validation

- `pnpm install` in `docs/` succeeds and the lockfile is consistent.
- No vulnerable versions remain (`lodash-es` 4.17.23 is fully replaced by the override).

### DIFF
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -5,7 +5,7 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
-  lodash-es: ^4.18.0
+  lodash-es: ^4.18.1
 
 importers:
 
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -5,10 +5,7 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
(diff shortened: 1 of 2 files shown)


fengmk2:
renovate re-enable https://github.com/voidzero-dev/vite-plus/pull/1744

PR #1753, opened 2026-06-03
fix(cli): handle stale global install backups

fengmk2:
## What changed

- Use a unique backup directory when reinstalling managed global packages.
- Treat old backup cleanup as best-effort so locked Windows binaries do not fail an otherwise successful reinstall.
- Add a Windows-focused global snap test that reproduces a stale locked backup path before reinstalling a scoped package.

## Why

On Windows, a running globally installed tool can keep files locked under the previous backup location. The old reinstall flow reused a fixed backup path under `VP_HOME/tmp/packages/...`, so cleaning that stale path could fail with `Access is denied. (os error 5)` before npm ran.

## Validation

- `pnpm -F vite-plus snap-test-global env-install-stale-backup`
- `cargo test -p vite_global_cli test_package_backup_uses_unique_tmp_dir_for_scoped_package`
- `git diff --check -- crates/vite_global_cli/src/commands/global/install.rs packages/cli/snap-tests-global/env-install-stale-backup .github/workflows/ci.yml`

---

> [!NOTE]
> **Medium Risk**
> Changes filesystem behavior on the global install commit path; low blast radius but reinstall/rollback timing is user-facing on all platforms.
> 
> **Overview**
> Global reinstalls no longer rely on a **fixed** `VP_HOME/tmp/packages/...` backup path. Before npm runs, the existing package dir is moved into a **unique** sibling name (`{pkg}.{pid}.{nanos}.old`) via new `unique_backup_dir`, so a leftover or **file-locked** stale backup (common on Windows when a tool is still running) is not deleted or overwritten.
> 
> After a successful install, **discarding** the previous backup is **best-effort**: cleanup failures are `tracing::warn` only and no longer fail the install or block success output.
> 
> Adds a unit test for scoped packages with a stale fixed backup path and a **Windows-only** global snap test that seeds a locked stale backup, reinstalls, and asserts the stale tree and bin config stay valid.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a4b2163f148ea26ceaa3aa4209129e8ddcf9174. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fengmk2:
@cursor review

fengmk2:
cc @liangmiQwQ

cpojer (approved):

liangmiQwQ (approved):
Thank you for finding and fixing it ❤️

PR #1758, opened 2026-06-04
fix(cli): keep absolute tsgolint path for workspace lint runs

### LINKED ISSUE #1747: `vp run -r lint` fails on Windows in monorepos: OXLINT_TSGOLINT_PATH is set to a root-relative path resolved from the wrong cwd
### Describe the bug

On Windows, running `vp run -r lint` (or `vp lint` for a workspace package) in a pnpm monorepo fails with:

```
Failed to find tsgolint executable: OXLINT_TSGOLINT_PATH points to
'.\node_modules\.pnpm\vite-plus@0.1.22_...\node_modules\vite-plus\node_modules\.bin\tsgolint.cmd'
which does not exist
```

The `tsgolint` executable **does** exist, the path is just wrong. `lint()` computes `OXLINT_TSGOLINT_PATH` as a path relative to `process.cwd()` of the launcher (the repo root under `vp run -r`), but oxlint is then spawned with its cwd set to the **workspace package** directory. The relative `.\node_modules\.pnpm\...` therefore resolves against `packages/<pkg>/`, where no `.pnpm` directory exists (in a pnpm workspace, `.pnpm` lives only at the repo root).

## Root cause

In `dist/bin.js`, `lint()` resolves the correct **absolute** path to `tsgolint`, then downgrades it to a relative path before exporting it:

```js
if (!oxlintTsgolintPath) throw new Error("Unable to resolve oxlint-tsgolint executable, ...");
const relativePath = relative(process.cwd(), oxlintTsgolintPath);
oxlintTsgolintPath = /^[a-zA-Z]:/.test(relativePath) ? relativePath : `.\\${relativePath}`;
```

`oxlintTsgolintPath` is already a valid absolute path at this point. The relative conversion assumes oxlint will run with the same cwd as `lint()`, which is not true for workspace runs: oxlint receives a path that is relative to the wrong base directory.

(For confirmation: the error string itself is emitted by the `oxlint` binary, which is correctly reporting that the path it was handed does not exist. The bad value originates here in vite-plus.)

## Suggested fix

Keep the resolved absolute path instead of converting it to a launcher-relative path:

```js
if (!oxlintTsgolintPath) throw new Error("Unable to resolve oxlint-tsgolint executable, ...");
// keep the absolute path: oxlint may be spawned with a different cwd (e.g. the workspace
// package dir under `vp run -r`), where a root-relative path would not resolve.
```

i.e. drop the two `relativePath` lines. If a relative path is required for some other reason, it should be computed against the directory oxlint is actually spawned in, not the launcher's `process.cwd()`.

## Workaround

A local `pnpm patch` of `vite-plus@0.1.22` that removes the relative conversion makes `vp lint` work correctly across all workspace packages (or use WSL).

## Steps to reproduce

1. pnpm workspace with a package under `packages/lib` (its own `lint` script calling `vp lint`).
2. `.pnpm` exists only at the repo root (standard pnpm layout).
3. On Windows, run `vp run -r lint` from the repo root.
4. Lint fails with the `OXLINT_TSGOLINT_PATH points to '...' which does not exist` error.

Only triggers on `process.platform === "win32"` (the relative conversion is Windows-only).

## System Info

- vite-plus: **0.1.22** (also present unchanged in latest **0.1.24**
- OS: Windows 11
- Node: v24.15.0
- pnpm: 10.25.0
- oxlint 1.63.0 / oxlint-tsgolint 0.22.1

### Validations

semimikoh:
Closes #1747

### Problem

On Windows, `vp run -r lint` (or `vp lint` for a workspace package) in a pnpm monorepo fails with:

```
Failed to find tsgolint executable: OXLINT_TSGOLINT_PATH points to
'.\node_modules\.pnpm\...\tsgolint.cmd' which does not exist
```

The executable exists; the path is just wrong.

### Root cause

In `packages/cli/src/resolve-lint.ts`, the Windows-only branch resolves a valid **absolute** path to `tsgolint`, then downgrades it to a path relative to the launcher's `process.cwd()`:

```js
const relativePath = relative(process.cwd(), oxlintTsgolintPath);
oxlintTsgolintPath = /^[a-zA-Z]:/.test(relativePath) ? relativePath : `.\\${relativePath}`;
```

This assumes oxlint runs with the same cwd as the launcher. Under `vp run -r`, oxlint is spawned with the **workspace package** directory as its cwd, so a root-relative `.\node_modules\.pnpm\...` path resolves against `packages/<pkg>/`, where no `.pnpm` exists (in a pnpm workspace `.pnpm` lives only at the repo root).

### Fix

Keep the resolved absolute path; drop the launcher-relative conversion. The candidate resolution is extracted into a pure `resolveWindowsTsgolintExecutable` helper so the contract can be unit-tested without mocking `process.platform`/`fs`.

### Testing

- Added a regression test that locks the resolved Windows path as absolute (never `.\`-relativized). Runs deterministically on all platforms.
- `resolve-lint` unit tests pass; the changed file lints clean.

> Note: I don't have a Windows machine, so I could not reproduce the end-to-end `vp run -r lint` scenario locally. The fix matches the analysis and the reproduction in #1747, where the reporter confirmed that removing the relative conversion (via `pnpm patch`) makes `vp lint` work across all workspace packages. CI's `windows-latest` jobs should exercise the non-regression path.


jong-kyung:
Could you check the `vp check` CI failure?

jong-kyung (approved):
Looks good to me. Thanks!

semimikoh:
Updated the branch with `main` and fixed a formatting issue in `resolve-lint.ts` that was failing `vp check` in the E2E jobs.

fengmk2:
@codex review

fengmk2 (approved):
@semimikoh Thanks!

PR #1759, opened 2026-06-04
chore(upgrade-deps): keep generated bindings

fengmk2:
## Summary

- Fixes the upgrade-deps prompt so generated build diffs are inspected before the automated PR is created.
- Keeps generated NAPI binding updates, instead of discarding `packages/cli/binding/index.cjs` and `packages/cli/binding/index.d.cts`.

## Root cause

PR #1757 failed CI because `build-upstream` regenerated `packages/cli/binding/index.d.cts`, but the upgrade workflow checked out the binding files before opening the PR, so the required generated change was missing.

## Validation

- `git diff --check -- .github/workflows/upgrade-deps.yml`
- `git diff --check HEAD~1..HEAD`
- Commit hook ran `vp check --fix`.


fengmk2:
verify at https://github.com/voidzero-dev/vite-plus/actions/runs/26937935248

fengmk2:
it works https://github.com/voidzero-dev/vite-plus/pull/1761

cpojer (approved):

PR #1765, opened 2026-06-04
fix(cli): remove stale global package shims

liangmiQwQ:
The current `vp install -g` flow backs up and replaces the package installation directory, then writes shims for binaries declared by the newly installed package. If a package update removes a binary from its `package.json#bin`, the old shim and per-binary config can remain in `VP_HOME/bin` and `VP_HOME/bins`. It seems introduced in #524.

This PR removes stale managed binary shims during successful global package reinstall. It compares the package's previous metadata and bin configs against the newly installed bin list, then deletes shims/configs for binaries that the package no longer declares.

This keeps existing behavior where `vp install -g` does not remove shims before invoking npm; cleanup only happens after the new package has installed and its current binaries have been finalized.

This PR also adds a global snapshot case with two local package fixtures that share the same package name: v1 declares two bins, v2 declares one bin, and the test verifies the removed bin's shim and config are gone after reinstall.

🤖 Generated with Codex


liangmiQwQ:
@codex

fengmk2 (approved):
@liangmiQwQ Thanks! 💯

PR #1770, opened 2026-06-06
fix(global): copy package backups before install

liangmiQwQ:
The current `vp install -g` backup flow moves the installed package directory into the temporary backup path before npm starts. If the process is interrupted by an unrecoverable failure like SIGINT or process termination, the original package can be left unavailable from its install location.

This PR changes global package backups logic, it now copies the package to `tmp` instead of moving it, so that the original package can still exist. 

It doesn't mean that the backup logic is now meaningless, it is still used for bin conflict case (when `npm install` is finished).

🤖 Generated with Codex


liangmiQwQ:
@codex

crates/vite_global_cli/src/commands/global/install.rs
```diff
@@ -561,6 +561,39 @@ async fn remove_dir_all_if_exists(path: &AbsolutePathBuf) -> Result<(), Error> {
...
+fn copy_dir_all_sync(from: &AbsolutePathBuf, to: &AbsolutePathBuf) -> Result<(), Error> {
+    std::fs::create_dir_all(to.as_path())?;
+    for entry in std::fs::read_dir(from.as_path())? {
+        let entry = entry?;
+        let source = entry.path();
+        let target = to.as_path().join(entry.file_name());
+        let file_type = entry.file_type()?;
+
+        if file_type.is_symlink() {
+            continue;
```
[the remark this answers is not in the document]
liangmiQwQ: Resolved in 31be5d8

crates/vite_global_cli/src/commands/global/install.rs
```diff
@@ -497,7 +497,7 @@ impl PackageBackup {
             tokio::fs::create_dir_all(parent).await?;
         }
 
-        tokio::fs::rename(package_dir, &backup_dir).await?;
+        copy_dir_all(package_dir, &backup_dir).await?;
```
fengmk2, line 500: Will this copy operation cause a large amount of IO and make the speed much slower? And under the same interrupt situation, the copy would be incomplete as well, right?

liangmiQwQ: Yes, what you said is true. I also thought about it.

Actually I do not really like the backup logic added in #1698, `npm` itself can handle most installation failure's recoveries so Vite+'s side backup is mainly used for resolving bin conflict. When I was thinking how to fix this bug, I thought we may remove the backup logic by using like `npm view` and process bin conflict before running `npm install`. But I was not sure which cost more, coping or network fetching. so I chose the copy approach to fix it.

fengmk2: Let's remove the backup logic. Change it to the plan you think is most appropriate. 👍

PR #1773, opened 2026-06-06
fix(global): preflight managed install bin conflicts

liangmiQwQ:
Close #1770

The current managed global install path uses a package-directory backup to recover from binary ownership conflicts after npm has already installed the package. According to #1770. That move-based backup logic can sometimes delete the package, like process kill or SIGINT. 

This PR resolves managed bin conflicts before running `npm install`, then lets `npm install` directly into the final package prefix. It completely removes the backup logic and avoids potential problems. However, it may slightly reduce the installation speed (won't influence global update).

There is some behavior changes

1. If users run `vp install -g pkga pkgb --force`, while `pkga` and `pkgb` have the same bin name, even if `--force` flag, it will still report error. The original behavior (install both of them, and uninstall the earlier one) can be a little bit odd and inefficient after this change, since we can get the package's bins before any installation. I believe this helps simplify the implementation and ensures the transparency and semantics of `--force`.
2. Vite+ won't run the rest of packages' installation if there is a package not found. This is because the package's status can be known while checking bins, if we simply ignore `npm view`'s error, network errors may cause bin overlapping issues.

🤖 Generated with Codex


liangmiQwQ:
@codex

liangmiQwQ:
@codex

liangmiQwQ:
@codex

fengmk2:
@codex review

liangmiQwQ:
For now, based on Codex's review suggestions, I don't think this is and there is a completely effective and feasible solution that can ensure all packages can be runnable for reinstall due to npm limitation or system IO performance (#1770). Changes like this PR just make the problem happen in different conditions that is not much meaningful. I also made some local attempts, but I really didn't come up with any good ideas. So based on that, maybe keeping the backup logic will be better?

fengmk2 (approved):

liangmiQwQ:
@fengmk2 thanks for reviewing, but considering what I laid out in https://github.com/voidzero-dev/vite-plus/pull/1773#issuecomment-4640835953, this PR just moves the problem rather than solving it, I suggest closing it and keeping the current behavior.

I will be kind of busy in the next two weeks, but I am still thinking about a way which can completely resolve this problem. I believe it would be better to revisit this later rather than merge a solution that doesn't really address the root cause ❤️

fengmk2:
@liangmiQwQ No problem, good luck on your exam! 💯

PR #1774, opened 2026-06-06
feat: integrate vite-task runner-aware tools (auto output + tracked envs)

### LINKED ISSUE #1894: Task cache: a build with outDir under a tracked dir (e.g. Laravel public/build) never caches — emptyOutDir delete-probe counted as a read
### Describe the bug

When a cached `vp build` task empties its output directory before writing (Vite's default `emptyOutDir: true`), Node's recursive remove probes each existing output file with `openat(path, O_RDONLY|O_NOFOLLOW|O_CLOEXEC|O_DIRECTORY)`, which fails with `ENOTDIR` on a regular file and is followed by `unlink`. The task-cache fs tracker counts that **failed** `O_RDONLY` probe as a content read. Because the build then writes a fresh same-named file (a manifest, `index.html`, a copied static asset) to that path, the task is flagged as having both read and written it → `modified its input` → never cached.

This is the same class as the already-fixed #1187 / #1095 ("output files in input scope → task modifies its own input"). Those were fixed by excluding the **default** output dir from input tracking, which is why a default `dist/` build caches fine today. The bug here is the **residual case that exclusion does not reach**: when `outDir` is a custom path **nested under a directory that is itself tracked input** — e.g. `laravel-vite-plugin`'s `build.outDir = 'public/build'`, nested under `public/` whose static assets the build reads and copies — the delete-probe under that path is still counted, and the build is uncacheable.

### Minimal standalone reproduction (no Laravel, no framework)

Public repo: **https://github.com/gsmeira/vite-plus-cache-repro** (`pnpm install`, then the steps below). Two sibling apps with identical source; **only `build.outDir` differs**:

```ts
// apps/control/vite.config.ts — default outDir → caches
export default defineConfig({
  build: { manifest: "manifest.json", emptyOutDir: true },          // outDir defaults to dist/
  run:   { tasks: { build: { command: "vp build", cache: true } } },
})

// apps/bug/vite.config.ts — outDir nested under the tracked public/
export default defineConfig({
  build: { outDir: "public/build", manifest: "manifest.json", emptyOutDir: true },
  run:   { tasks: { build: { command: "vp build", cache: true } } },
})
```

`public/` is Vite's default `publicDir`; its static asset (`public/robots.txt`) is read and copied by the build, so the tracker treats `public/` as input scope and `public/build/` (the outDir) is nested inside it.

Run for each app (build #2 executes over build #1's output, which is what arms `emptyOutDir`'s delete):

```bash
vp cache clean && vp run --filter <app> build   # build #1 — writes output
vp cache clean && vp run --filter <app> build   # build #2 — executes over prior output, repopulates cache
vp run --filter <app> build                     # build #3 — should be a cache hit
vp run --last-details
```

Observed on **`vite-plus 0.2.1`**:

| App | outDir | build #3 |
|-----|--------|----------|
| `control` | `dist/` (default) | **100% cache hit** |
| `bug` | `public/build/` | **0% — `Not cached: read and wrote '…/public/build/<file>'`** |

The named file varies run to run (whichever stable-named output under `public/build/` the tracker counts first — `manifest.json`, `index.html`, or a copied `public/*` asset); `strace -f -e trace=%file` confirms the manifest receives the identical probe:

```
lstat(".../public/build/manifest.json") = 0
openat(".../public/build/manifest.json", O_RDONLY|O_NOFOLLOW|O_CLOEXEC|O_DIRECTORY) = -1 ENOTDIR
unlink(".../public/build/manifest.json") = 0
# ...later, in a worker thread, the fresh write:
openat(".../public/build/manifest.json", O_WRONLY|O_CREAT|O_TRUNC) = 29
```

So the trigger is purely structural — a stable-named output written into an `outDir` nested under a tracked input dir — and is independent of Laravel; `public/build` is just the most common instance of that shape.

### What actually fixes it (fix localized to input-side exclusion)

Tested on `0.2.1`, `website-laravel`, building over prior output:

| Task config | Result |
|-------------|--------|
| baseline (the bug) | **0% — `read and wrote public/build/…`** |
| **exclude the outDir from input:** `input: [{ auto: true }, "!public/build/**"]` | ✅ **100% cache hit** |
| declare output: `output: ["public/build/**"]` | ❌ 0% — `read and wrote public/build/…` |

So the cure is **input-side exclusion of the output dir**, not output declaration. Two further non-invasive confirmations the delete-probe is the sole cause: with **no** prior output (`rm -rf public/build` first) the build caches; and `build.emptyOutDir: false` (no recursive delete → no probe) makes it cache.

### Relationship to existing work

- **#1187, #1095** (closed/fixed): the same "task modifies its own input" class, fixed for the default output dir. This report is the uncovered case for a **custom `outDir` nested under a tracked directory**.
- **#1774** (open, draft — "integrate vite-task runner-aware tools: auto output + tracked envs"): routes Vite's real inputs/outputs to the runner via `ignoreInput`/`ignoreOutput`. That `ignoreInput` channel is exactly the input-side exclusion shown above to fix this, so **#1774 very likely resolves this** — could a maintainer confirm whether it covers a custom `outDir` nested under a tracked dir (not just the default `dist/`)? If so this can ride on #1774; if not, the targeted fix below stands on its own.

### Impact

Any cacheable build that empties an output directory located under a tracked path before writing stable-named files (Laravel `public/build/manifest.json`, SSR `bootstrap/ssr/ssr-manifest.json`, copied public assets) is spuriously uncacheable on **every** incremental rebuild where prior output exists — i.e. exactly the warm-cache case the task cache exists to accelerate. Fresh checkouts / CI (no prior output) are unaffected.

**Environment:** `vp 0.1.12`; `vite-plus 0.2.1` (`vite-plus-core 0.2.1`, `vite 8.0.16`, `rolldown 1.1.1`); Linux. fs tracker = fspy.

### Reproduction

https://github.com/gsmeira/vite-plus-cache-repro

### Steps to reproduce

Public minimal repo (verified from a clean `git clone` + `pnpm install --frozen-lockfile`): two identical apps differing only in `build.outDir` — `apps/control` (`dist/`, caches 100%) vs `apps/bug` (`public/build/`, 0% — "read and wrote …/public/build/…").

```bash
pnpm install
# bug — never caches:
vp cache clean && vp run --filter bug build      # build #1 (writes output)
vp cache clean && vp run --filter bug build      # build #2 (executes over prior output, repopulates cache)
vp run --filter bug build && vp run --last-details   # -> 0%, "read and wrote …/public/build/…"
# control — caches (same steps, --filter control) -> 100% cache hit
```

### System Info

```shell
$ vp --version
vp v0.1.12
Local vite-plus:
  vite-plus  v0.2.1
Tools:
  vite             v8.0.16
  rolldown         v1.1.1
  vitest           v4.1.8
  oxfmt            v0.55.0
  oxlint           v1.70.0
  tsdown           v0.22.3
Environment:
  Package manager  pnpm v11.8.0
  Node.js          v>=22.12.0 (engines.node)

$ vp env current
Environment:
  Version       24.17.0
  Source        engines.node
  Project Root  <repro root>
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

wan9chi:
Integrates the runner-aware-tools work now available on `vite-task/main`: Vite reports its inputs/outputs/envs to the `vp` runner via `@voidzero-dev/vite-task-client`, so `vp build` no longer needs a hand-synthesized cache config. This enables auto output restoration and auto tracked envs through the runner protocol.

## Changes

- **Bump vite-task git deps to `82ec12b2`** (latest `main`): runner-aware IPC server/client + embedded napi addon, tracked env fingerprinting, automatic output tracking/restoration by default, per-schema-version cache directories, and prefix-env executable lookup fixes. Regenerate the napi binding (`RunnerClient`/`load` exports) and `run-config.ts` for the new task-config types (`AutoTracking`).
- **Point the vite behind `vp build`/`vp dev` to latest vitejs/vite (`c13a37b53`)**, which carries the vite-task-client integration. vite lists `@voidzero-dev/vite-task-client` in **devDependencies**, so it is bundled into vite's output, and therefore into `@voidzero-dev/vite-plus-core`, the same way as vite's other bundled deps (cac, esbuild, ...). No extra runtime dependency or build wiring is needed.
- **Remove `vp build`'s synthetic cache config** (`env: ["VITE_*"]`, manual input excludes). The all-`None` `EnabledCacheConfig` keeps caching enabled with auto input/output inference; vite's `ignoreInput`/`ignoreOutput`/`getEnv`/`getEnvs` refine the fingerprint at runtime. `vp dev` stays `disabled()`.
- **Docs**: update cache/run docs for automatic output restoration by default and `output: []` as the opt-out.
- **Snap tests**: `vp-build-auto-output-restoration` (dist restored on cache hit without a synthetic `output` config) and `vp-build-auto-tracked-env` (a `VITE_` env change invalidates the cache and names the changed env). Edge cases live in vite-task.

## Notes

- The vite-task stack has landed on `main`; this PR now pins latest `main` (`82ec12b2`). Changelog diff for the latest repin: https://github.com/voidzero-dev/vite-task/compare/68fb22e49ad55ee308de24cef387995ddcedd5bd...82ec12b2956c81f55e40749f65c062ab368a09e7#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed
- rolldown stays at main's pin (`c462c7c5`, v1.1.0), which is compatible with vite `c13a37b53`'s rolldown `1.0.3` requirement.


packages/core/build.ts
```diff
@@ -315,9 +315,20 @@ async function buildVite() {
...
+    let rewrittenFile = rewriteModuleSpecifiers(file, srcDtsFile, {
       rules: [...createViteRewriteRules(pkgJson.name), ...createRolldownRewriteRules(pkgJson.name)],
     });
+    // Upstream bug (vitejs/vite#21863, commit cc39e5540): `node/index.ts`
+    // re-exports `KnownQueryTypeMap` from `#types/importGlob`, but the type is
+    // declared there without `export`. Bundling vite's types into a downstream
+    // project then fails with `[MISSING_EXPORT]`. Add the missing `export` here
+    // until it is fixed upstream (drop this once vite exports it).
+    if (relativePath === '/importGlob.d.ts') {
+      rewrittenFile = rewrittenFile.replace(
```
fengmk2, line 334: A check needs to be added here; if the replacement is not successful, an error should be reported, indicating that the upstream has already been fixed, and we can remove this section of code.

packages/cli/binding/src/cli/resolver.rs
```diff
@@ -313,7 +319,7 @@ fn exclude_glob(pattern: &str, base: InputBase) -> UserInputEntry {
 /// by a runner-aware approach that automatically excludes task output directories.
 fn build_pack_cache_inputs() -> Vec<UserInputEntry> {
     vec![
-        UserInputEntry::Auto(AutoInput { auto: true }),
+        UserInputEntry::Auto(AutoTracking { auto: true }),
         exclude_glob("!node_modules/.vite-temp/**", InputBase::Workspace),
         exclude_glob("!node_modules/.vite-temp/**", InputBase::Package),
```
liangmiQwQ: According to https://discord.com/channels/1475973262193459293/1479410337223802951/1518498738413244446, these two lines which exclude `.vite-temp` config files could probably be safely removed, since the config files will be ignored automatically.

 https://github.com/vitejs/vite/blob/403cc608e4e8633b0335f5e8221c83dfca4994f8/packages/vite/src/node/config.ts#L2652-L2655

fengmk2 (approved):

PR #1787, opened 2026-06-08
fix(tools): keep sync-remote from churning pnpm-workspace.yaml

fengmk2:
`pnpm tool sync-remote` made two unwanted changes to `pnpm-workspace.yaml` on every run. This PR fixes both.

**Redundant `minimumReleaseAgeExclude` entries.** The merge was a naive `Set` union, so versioned upstream entries (e.g. `oxc-parser@0.134.0`) were re-added even though the main file already excludes them via broader patterns (`@oxc-parser/*`, bare `oxc-parser`), adding ~65 redundant lines. `mergeMinimumReleaseAgeExclude` now drops any versioned entry already covered by a version-less pattern.

**Dropped comments.** The file was parsed to a plain object and re-stringified, deleting all comments (e.g. the zod-v3 pin rationale added in #1777). `mergeWorkspaceYaml` now parses the main file as a yaml `Document`, merges via the unchanged `mergePnpmWorkspaces`, and reconciles the result back into the document so surviving keys keep their comments.

Tests cover both: the exclude dedupe, plus comment preservation and a deep-equal check that the comment-preserving path produces the same merged data as before. tools suite and lint pass.


fengmk2:
test on https://github.com/voidzero-dev/vite-plus/actions/runs/27136598214

cpojer (approved):

PR #1788, opened 2026-06-08
fix(cli): reject git options inside existing monorepos

jong-kyung:
## Summary

Reject `--git` and `--no-git` when using `vp create vite:generator`.

`vite:generator` adds a generator package to an existing monorepo, so git initialization options do not apply. The command now fails early with a clear message when either option is provided, while the normal `vp create vite:generator` flow continues to skip git initialization.


jong-kyung:
@codex review

jong-kyung:
@codex review

fengmk2:
@codex review

packages/cli/src/create/bin.ts
```diff
@@ -584,6 +585,12 @@ Use \`vp create --list\` to list all available templates, or run \`vp create --h
     );
     cancelAndExit('Cannot create a generator outside a monorepo', 1);
   }
+  if (selectedTemplateName === BuiltinTemplate.generator && options.git !== undefined) {
+    cancelAndExit(
+      'The --git/--no-git options are not available with vite:generator because it adds a package to an existing monorepo',
```
fengmk2: Should it not be only for `vite:generator`? The `vp create` command does not support the git parameter in a monorepo configured with git.

jong-kyung: You are right — I think I missed the broader monorepo case here. `--git/--no-git` should not apply when `vp create` is adding a package inside an existing git-configured monorepo, not only for `vite:generator`.

I’ll update this behavior and push a follow-up commit.

jong-kyung: @fengmk2 Thanks for the review 🫡 I updated it

fengmk2 (approved):

PR #1791, opened 2026-06-08
ci: enable step debug logging on all setup-node jobs (temporary)

fengmk2:
## Problem

Jobs running `oxc-project/setup-node` (wrapping `actions/setup-node`) sporadically fail right after `Attempting to download <version>...` with no error message. The built-in `downloadTool` retry never engages, so the process dies before the retryable path. Matches actions/setup-node#1555 (open, same `v6.4.0`).

## What this does

Scopes `ACTIONS_STEP_DEBUG=true` to every job that runs setup-node, via a `debug` GitHub Environment (no protection rules), so the next reproduction emits debug logs to attach to setup-node#1555. Jobs without setup-node keep normal log verbosity.

Covered: `ci.yml` (lint, cli-e2e-test, cli-snap-test, cli-e2e-test-musl, install-e2e-test, install-e2e-test-sfw), `e2e-test.yml` (build), `test-vp-create.yml` (build), `prepare_release.yml` (prepare), `reusable-release-build.yml` (build-rust), `upgrade-deps.yml` (upgrade).

Temporary: revert this PR and delete the `debug` environment once a failing run with debug logs is captured.

Note: this repo is public, so debug logs are public. The release-path jobs (prepare, build-rust, upgrade) run with secrets; GitHub masks secret values even in debug logs, but tell me if you would rather exclude those three.


fengmk2:
https://github.com/voidzero-dev/vite-plus/pull/1792

PR #1808, opened 2026-06-10
feat(env): add corepack to the default shims tool list

### LINKED ISSUE #858: Shim for `pnpm` & `yarn`
### Description

Currently, if I install vite-plus globally, it works great for the most part, but as soon as I use any AI agent, the problem starts to show up, where the agent will attempt to use `pnpm`/`yarn` instead of `vp`, which is something already solved in the project-level `AGENTS.md` vite-plus section. However, if other teams don't have Vite-plus installed globally, their agent will tend to use `vp` and result in a command not found.

The only workaround is to disable the `vp env`, and use `fnm`/`nvm` + `corepack` to manage my node/package manager.

### Suggested solution

Have a shim for both `pnpm` and `yarn`, then let Vite+ resolve the correct version of the `pnpm`/`yarn`, just like how `node`, `npm` & `npx` shims work.

### Alternative

_No response_

### Additional context

_No response_

### Validations

### LINKED ISSUE #1309: Vite+ and corepack-like functionality
### Describe the bug

**TL;DR** Using Vite+ without an existing Node installation means no corepack, so any scripts/commands that try to run pnpm/yarn fail. Corepack can be installed, but will manage its own versions of pnpm/yarn, same as Vite+, leading to differences between running `pnpm i` and `vp i`.

Also, this is a little bit of a dupe of #858 (sorry, didn't see it before opening this 😞), but I think this issue better explains the problem & proposes a better solution.

---

I tried cloning the `vite-plus` repo according to the contribution guide with `just init`, and I've encountered some problems. I decided to document this here as I think it has some value regarding `vp` as a full-featured tool for everything JS-related, which seems to be its goal, and does very well at it too.

I'll start by saying that I removed `nvm` and associated global `node` installations first, to rely only on `vp env`. The issue described here is encountered by anyone that uses `vp env` for managing Node, without an existing node installation.

`just init` tries to run `pnpm install`, but fails with:

```
Error: Command failed: pnpm install --no-frozen-lockfile
'pnpm' is not recognized as an internal or external command, operable program or batch file.
```

Makes sense, since it downloaded a new node version. Normally, I'd enable corepack when I see this, but this doesn't work:

```shell
> corepack enable
corepack: The term 'corepack' is not recognized as a name of a cmdlet, function, script file, or executable program.
> npm list -g
C:\Users\Matei\.vite-plus\js_runtime\node\24.14.1
+-- corepack@0.34.6
`-- npm@11.11.0
```

Corepack is there, but Vite+ did not create a shim for it in `~/.vite-plus/bin`. Vite+ _does_ create shims for global packages, when using `npm i -g` or `vp i -g`. But since `corepack` comes preinstalled with node, this never happens.

> **Note:** Both `npm i -g` and `vp i -g`, from my understanding, install a package globally, but for a specific node version. `yarn ...` will always run on the Node version I installed it with, _not_ the dynamic Node version (based on `.node-version` or `package.json#engines`). It will remain stuck on whatever node version I happened to be using, which _does_ depend on CWD, `.node-version` etc., forever. This seems like a potential design flaw, but I may be wrong, and regardless it's not related to the current issue. I thought it was worth pointing out, especially since it's not documented.

Continuing, the problem is that trying to call package managers directly, which relies on corepack, doesn't work out of the box. This includes `vite-plus` itself, as it runs `pnpm` directly. This can be fixed easily, however, by running:

```shell
vp i -g corepack
```

This works, but IMO is not "clean". Corepack now downloads and manages package manager versions independently of Vite+, so depending on whether you use `vp add` or `pnpm add`, a different installation (which may or may not be the same version) is used, as well as with a different node version.

My suggestions to address this, the first one being IMO the best:
- Vite+ should implement `corepack` capabilities itself. It already does this when using `vp i`, `vp add` etc., which install and run the correct package manager and version automatically. The only piece missing is shimming the actual `yarn`, `pnpm` etc. executables by default.
- If not, at least include some documentation on its [environment page](https://viteplus.dev/guide/env) that corepack is not available by default, and `vp install`, `vp add` etc. should be used instead. However, I do think some people will be turned away by this - some people will simply not want to adjust their workflows (scripts, muscle memory) to use `vp` instead of `pnpm`.

If this is out of scope, and `vp` only wants to support managing package managers for its own commands, then:
- Document on its [environment page](https://viteplus.dev/guide/env) that corepack is not available by default, and that it should be installed with `vp install -g corepack`. Even as Node will no longer come with corepack by default from now on, this information would still be useful.
- Automatically run `vp install -g corepack` when installing `vp`, or an equivalent.

### Reproduction

1. Make sure there is no global Node installation and you're only relying on `vp env`
2. Try to clone/work on any repo which uses pnpm/yarn commands

### Validations

fengmk2:
Closes #1309
Closes #858

Without a system Node.js, corepack is unreachable under Vite+ even though Node.js <= 24 bundles it, and Node.js 25+ removed it entirely. This adds `corepack` to the default shims created by `vp env setup`.

## How it works

- Resolution order: vp-managed global install (`vp install -g corepack`) > Node-bundled corepack (Node.js <= 24) > auto-install as a managed global package (Node.js 25+). The auto-install links only the `corepack` bin, so it never conflicts with an existing `vp install -g pnpm` (the corepack package also declares pnpm/yarn launcher bins).
- `corepack enable`/`disable` without an explicit `--install-directory` get `--install-directory ~/.vite-plus/bin` injected, so the created pnpm/yarn launchers land on PATH and still resolve the project Node.js version. Vite+-owned shims (npm/npx, `vp install -g` binaries) are restored with a warning if corepack removes or replaces them.
- `vp remove -g corepack` keeps the default shim; `vp env doctor` and `vp env which corepack` cover the new shim.

RFC updates in `rfcs/env-command.md` (new Corepack Shim section), `rfcs/trampoline-exe-for-shims.md`, and `rfcs/upgrade-command.md` describe the design.

## Testing

- New snap tests: `shim-corepack-bundled` (real bundled corepack via Node 20.18.0) and `shim-corepack-enable-install-directory` (hermetic test for the install-directory injection and npm shim restore).
- Verified manually: bundled dispatch, auto-install on Node 25 (corepack 0.34.7, only `corepack` bin linked), managed-first precedence, `vp remove -g corepack` fallback, and a real `corepack enable`/`disable` cycle (`yarn --version` works through the launcher, vp-managed pnpm restored).

Note: `just lint` currently fails on main due to a pre-existing `clippy::unused_async_trait_impl` warning in `js_executor.rs` from the nightly-2026-06-10 bump, unrelated to this change.

---

> [!NOTE]
> **High Risk**
> Changes core shim dispatch, global bin ownership, and auto-install side effects on first `corepack` use; mistakes could break npm/corepack on PATH or corrupt bin layout on Windows.
> 
> **Overview**
> **Adds `corepack` as a default env shim** (setup, install scripts, docs, E2E varlet matrix uses `corepack enable` instead of `vp i -g pnpm`).
> 
> **New dedicated corepack dispatch** resolves in order: `vp install -g corepack` → Node-bundled corepack → auto-install as a managed global (Node 25+), with optional **`only_bins`** so auto-install links only `corepack` (not pnpm/yarn launchers). **`bins_restricted`** metadata preserves that restriction across `vp update -g`.
> 
> **`corepack enable`/`disable`** default `--install-directory` to `VP_HOME/bin`, then **restore** Vite+-owned shims (default npm/npx, `vp install -g`, npm links) if corepack clobbers them; Windows setup also strips conflicting `.ps1`/legacy launchers.
> 
> **Global install/refactor**: `InstallOptions` replaces positional args; **`is_protected_shim` / `package_may_own_bin`** guard default shims (corepack only ownable by the `corepack` package); protected shims survive uninstall/dry-run; **`ensure_installed`** returns the node path.
> 
> **Shim UX**: **`route_user_output_to_stderr`** during shim/`vp env exec` dispatch so wrapped tools keep parseable stdout; version warnings and npm install log routing follow the same rule.
> 
> **Tests/docs**: snap tests for bundled corepack and enable install-directory; RFCs updated for corepack design.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75c5f98d4d7757831248e0bc13fba6f83f5c1dfc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -327,8 +327,10 @@ jobs:
             node-version: 22
             command: |
               # scripts/bootstrap.mjs spawns `pnpm build` via tinyexec and needs
-              # pnpm on PATH (not exposed by the vp install itself).
-              vp i -g pnpm
+              # pnpm on PATH (not exposed by the vp install itself). corepack
+              # enable creates a pnpm launcher in the vp bin dir that resolves
+              # the project's pinned packageManager version (pnpm@9.15.9).
+              corepack enable
               node scripts/bootstrap.mjs
               # Report-only: oxlint 1.68.0 surfaces ts1038 ("A 'declare' modifier
               # cannot be used in an already ambient context.") on varlet's
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -667,6 +667,7 @@ async fn managed_install(
         concurrency.unwrap_or(DEFAULT_GLOBAL_INSTALL_CONCURRENCY),
         false,
         None,
+        false,
     )
     .await
     {
@@ -799,6 +800,7 @@ async fn managed_update(
         concurrency,
         true,
         None,
+        false,
     )
     .await
     {
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -662,12 +662,14 @@ async fn managed_install(
 ) -> Result<ExitStatus, Error> {
     if let Err((package_name, error)) = global::install::install(
         packages,
-        node,
-        force,
-        concurrency.unwrap_or(DEFAULT_GLOBAL_INSTALL_CONCURRENCY),
-        false,
-        None,
-        false,
+        global::install::InstallOptions {
+            node_version: node,
+            force,
+            concurrency: concurrency.unwrap_or(DEFAULT_GLOBAL_INSTALL_CONCURRENCY),
+            update: false,
+            only_bins: None,
+            progress_to_stderr: false,
+        },
     )
     .await
     {
@@ -795,12 +797,14 @@ async fn managed_update(
     // Call reinstall logic
     if let Err((package_name, error)) = global::install::install(
         &to_update,
-        Some(&current_node_version),
-        false,
-        concurrency,
-        true,
-        None,
-        false,
+        global::install::InstallOptions {
+            node_version: Some(&current_node_version),
+            force: false,
+            concurrency,
+            update: true,
+            only_bins: None,
+            progress_to_stderr: false,
+        },
     )
     .await
     {
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -256,7 +256,7 @@ impl Commands {
 #[command(after_help = "\
 Examples:
   Setup:
-    vp env setup                  # Create shims for node, npm, npx
+    vp env setup                  # Create shims for node, npm, npx, corepack
     vp env on                     # Use vite-plus managed Node.js
     vp env print                  # Print shell snippet for this session
 
@@ -666,6 +666,7 @@ async fn managed_install(
         force,
         concurrency.unwrap_or(DEFAULT_GLOBAL_INSTALL_CONCURRENCY),
         false,
+        None,
     )
     .await
     {
@@ -791,9 +792,15 @@ async fn managed_update(
     }
 
     // Call reinstall logic
-    if let Err((package_name, error)) =
-        global::install::install(&to_update, Some(&current_node_version), false, concurrency, true)
-            .await
+    if let Err((package_name, error)) = global::install::install(
+        &to_update,
+        Some(&current_node_version),
+        false,
+        concurrency,
+        true,
+        None,
+    )
+    .await
     {
         output::error(&format!(
             "Failed to update {}: {error}",
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -668,7 +668,6 @@ async fn managed_install(
             concurrency: concurrency.unwrap_or(DEFAULT_GLOBAL_INSTALL_CONCURRENCY),
             update: false,
             only_bins: None,
-            progress_to_stderr: false,
         },
     )
     .await
@@ -803,7 +802,6 @@ async fn managed_update(
             concurrency,
             update: true,
             only_bins: None,
-            progress_to_stderr: false,
         },
     )
     .await
--- a/crates/vite_global_cli/src/commands/env/exec.rs
+++ b/crates/vite_global_cli/src/commands/env/exec.rs
@@ -64,6 +64,8 @@ pub async fn execute(
         // - Recursion prevention via VP_TOOL_RECURSION
         // - Shim mode checking (managed vs system-first)
         let args: Vec<String> = command[1..].to_vec();
+        // stdout belongs to the dispatched tool; route vp's own output to stderr.
+        vite_shared::output::route_user_output_to_stderr();
         let exit_code = shim_dispatch(tool, &args).await;
         return Ok(exit_status(exit_code));
     }
--- a/crates/vite_global_cli/src/commands/env/mod.rs
+++ b/crates/vite_global_cli/src/commands/env/mod.rs
@@ -15,7 +15,7 @@ mod off;
 mod on;
 pub mod package_metadata;
 mod pin;
-mod setup;
+pub(crate) mod setup;
 mod unpin;
 mod r#use;
 mod which;
--- a/crates/vite_global_cli/src/commands/env/package_metadata.rs
+++ b/crates/vite_global_cli/src/commands/env/package_metadata.rs
@@ -24,6 +24,11 @@ pub struct PackageMetadata {
     /// Binary names that are JavaScript files (need Node.js to run).
     #[serde(default)]
     pub js_bins: HashSet<String>,
+    /// Whether `bins` was deliberately restricted to a subset of the bins the
+    /// package declares (e.g., the corepack shim auto-install links only
+    /// `corepack`). Updates keep the restriction; explicit installs reset it.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub bins_restricted: bool,
     /// Package manager used for installation (npm, yarn, pnpm)
     pub manager: String,
     /// Installation timestamp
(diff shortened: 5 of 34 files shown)


fengmk2:
## Manual verification checklist

### Shim creation and upgrade paths
- [x] Fresh install via `install.sh`: `~/.vite-plus/bin/corepack` symlink exists and `corepack --version` works without a system Node.js
- [x] Upgrade from a pre-corepack vp (`vp upgrade`): corepack shim is created by the automatic `env setup --refresh`
- [x] Plain `vp env setup` (no `--refresh`) on an existing install: corepack shim is created while node/npm/npx are reported as skipped
- [x] `vp env doctor`: corepack listed under Shims and under PATH checks
- [ ] Windows: `vp env setup` creates `corepack.exe` trampoline; `corepack --version` works from PowerShell, cmd.exe, and Git Bash

### Resolution order
- [x] Node <= 24 project (e.g. `.node-version` 20.18.0): `corepack --version` runs the bundled corepack (0.29.3); `vp env which corepack` shows the js_runtime path
- [x] `vp install -g corepack`: managed copy wins afterwards on every Node version; `vp env which corepack` shows the package path
- [x] Node 25+ project, no managed corepack: first `corepack --version` auto-installs (notice on stderr, `Bins: corepack` only); second invocation runs without reinstalling
- [x] Shim stdout stays parseable: on Node 25+ the very first `v=$(corepack --version)` captures only the version; all vp progress/notices (including the one-time auto-install) go to stderr (fixed by the process-wide shim output routing)
- [x] Corrupt `~/.vite-plus/bins/corepack.json` (e.g. truncate it): corepack warns about unusable managed state and falls back to the bundled copy instead of exiting 1
- [x] Node 25+ offline with no managed corepack: clear error with the `vp install -g corepack` hint

### corepack enable / disable
- [x] `corepack enable`: pnpm/yarn launchers land in `~/.vite-plus/bin` (injected `--install-directory`); `yarn --version` works and resolves the project Node.js
- [x] `corepack enable --install-directory /custom/dir`: explicit directory respected, nothing written to vp bin
- [x] `corepack disable`: launchers removed again
- [x] With `vp install -g pnpm` present: `corepack enable` restores vp-managed pnpm/pnpx shims with a warning; `corepack disable` preserves them. On Unix this may restore with a warning, while on Windows the `.exe` trampolines remain intact without a warning
- [x] With a pnpm link created via `npm install -g pnpm` interception: `corepack disable` restores the link with a warning
- [x] After deliberately deleting `~/.vite-plus/bin/npm`: `corepack enable` does NOT resurrect it and prints no restore warning
- [x] `corepack enable npm`: npm shim is restored with a warning; `npm --version` still resolves per project
- [x] `corepack enable --help`: prints help, touches no files
- [ ] Windows PowerShell: after `corepack enable npm` + restore, `npm` resolves the vp trampoline (no leftover `npm.ps1`/`npm.cmd`/extensionless wrapper)

### Old/new version compatibility
- [x] Existing `vp install -g corepack` (the old workaround) survives the upgrade: still wins over bundled, `vp env setup --refresh` keeps its BinConfig
- [x] Pre-PR `npm install -g corepack` user: after upgrade + `vp env setup`, the stale Npm BinConfig is dropped and a later `npm uninstall -g corepack` does NOT delete the default shim
- [x] `vp remove -g corepack`: package removed, default shim kept, resolution falls back to bundled; `--dry-run` marks the shim as `(kept: default shim)`
- [x] Auto-installed corepack then `vp update -g` (when a newer corepack exists): still links only the `corepack` bin, no pnpm/yarn shims appear, no BinaryConflict with a vp-managed pnpm
- [x] Explicit `vp install -g corepack` after an auto-install: resets the restriction and exposes the package's launcher bins (conflict semantics with `--force` unchanged)
- [x] Old metadata files without the `binsRestricted` field load fine (`vp list -g` works after upgrade)
- [ ] Windows downgrade via old `install.ps1`: stale `corepack.exe`/`vpr.exe` trampolines are removed
- [x] `vp upgrade --rollback` to a pre-corepack version: corepack shim remains in bin but old vp reports "Binary 'corepack' not found" (known behavior, confirm acceptable)

### Hardening added after the initial list
- [x] A package other than `corepack` declaring a `corepack` bin (e.g. a test package with `"bin": {"corepack": ...}`) does NOT take ownership: warn + skipped, `corepack` still resolves bundled/managed, no BinConfig written
- [x] `corepack enable -- pnpm` style invocations: `--install-directory` is injected before the `--` separator; `corepack -- enable` is passed through untouched
- [x] With corepack package state broken (corrupt `bins/corepack.json` or deleted package dir): both the shim AND `vp env which corepack` print the same "unusable vp-managed corepack" warning and fall back to the bundled copy
- [x] `vp update -g` on an auto-installed (restricted) corepack keeps linking only the `corepack` bin; explicit `vp install -g corepack` resets to the full bin list (BinaryConflict against a vp-managed pnpm is expected then)
- [ ] Windows: with an npm-interception-linked `yarn`, `corepack enable` overwrites the `.cmd` launcher and the restore detects it via content check, rewriting vp's wrapper and removing corepack's `.ps1`/extensionless files

### Modes and overrides
- [x] `vp env off` (system-first) with a system corepack installed: shim prefers the system corepack; `vp env on` restores managed behavior
- [x] `VP_BYPASS=1 corepack --version` bypasses the shim
- [ ] `vp env use 20` vs `vp env use 25` in the same shell: corepack switches between bundled and managed/auto-install resolution
- [x] `vp env exec corepack --version` works (shim mode without `--node`)

### CI
- [x] snap tests green: `shim-corepack-bundled`, `shim-corepack-enable-install-directory`, `command-env-which`, `command-env-setup-external-vp`, `cli-helper-message`

fengmk2:
@cursor review

fengmk2:
@codex review

fengmk2:
bugbot run

cpojer (approved):

fengmk2:
bugbot run

fengmk2:
@codex review

fengmk2:
@codex review

PR #1809, opened 2026-06-10
ci: reuse just lint and just test as the single source of truth

fengmk2:
The nightly-2026-06-10 bump (#1725) added allows for the new clippy lints (`unused_async_trait_impl`, `useless_borrows_in_formatting`) to the CI clippy command only, so local `just lint` fails on the vendored rolldown crates (they have no `[lints]` table, so workspace-level allows cannot reach them).

Fix: move the `-A` flags into the `just lint` recipe and have CI run `just lint`, so there is no second copy of the clippy command to keep in sync.

Also apply the same reuse to the test jobs: the test matrix and test-musl jobs now run `just test` instead of duplicating the package-enumerating cargo test command. The windows test recipe gains `RUST_MIN_STACK` to match what CI previously set via env.

Notes:
- The old CI clippy command omitted `--workspace` while the justfile has it; the root manifest is a virtual workspace, so lint coverage is identical.
- The test matrix job installs `just` via setup-rust tools (same pattern as upgrade-deps.yml); the musl job installs it from the Alpine community repo.
- On the Windows runner, `just test` now uses the `[windows]` recipe (powershell) instead of the bash command; it additionally sets `__COMPAT_LAYER=RunAsInvoker`, which is harmless on CI.

---

> [!NOTE]
> **Low Risk**
> Changes only affect CI orchestration and local/CI command parity; no application or security logic is modified.
> 
> **Overview**
> CI **test**, **test-musl**, and **lint** now delegate to **`just test`** and **`just lint`** instead of inlining duplicate `cargo test` / `cargo clippy` commands. **`just`** is installed in those jobs (via `setup-rust` tools or Alpine `apk`).
> 
> The **justfile** becomes the single source of truth: **`lint`** picks up the same clippy `-A` allows that CI used to carry alone (so local **`just lint`** matches CI after newer toolchain lints), and **`test`** documents the all-`crates/*` plus **`vite-plus-cli`** package list. The Windows **`test`** recipe now sets **`RUST_MIN_STACK=8388608`** like CI did via env.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f1d7027c986691cefcd3259d29aa8968dc5d399. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fengmk2:
@cursor review

wan9chi (approved):

PR #1811, opened 2026-06-11
fix(cli): set tsgolint path in oxlint wrapper

### LINKED ISSUE #1800: Add tsgolint to bin
### Description

When you install vite-plus in a project with pnpm (pnpm i -D vite-plus), oxfmt, oxlint, and vp are installed as binaries.
oxlint-tsgolint is installed as a dependency of vite-plus, but not as a binary (tsgolint).

vp check (typeAware: true) can be executed successfully, but the following error occurs with the [oxc-vscode](https://github.com/oxc-project/oxc-vscode) extension.
```sh
WARN oxlint::lsp::server_linter: Failed to initialize type-aware linting: Failed to find tsgolint executable. You may need to add the `oxlint-tsgolint` package to your project?
```

### Suggested solution

If you install oxlint-tsgolint, the warning will disappear, but I would like to match the version of oxlint-tsgolint used with vite-plus.
How about adding a wrapper script like oxfmt or oxlint and adding tsgolint to the bin field?
https://github.com/voidzero-dev/vite-plus/blob/24ac7276b68dbbc45c631c262e852afb63f3488e/packages/cli/package.json#L16-L20

Related PR
https://github.com/voidzero-dev/vite-plus/issues/629

### Alternative

_No response_

### Additional context

_No response_

### Validations

jong-kyung:
## Summary

- Set `OXLINT_TSGOLINT_PATH` from the Vite+ `oxlint` wrapper so oxlint LSP can find Vite+’s pinned `oxlint-tsgolint`.
- Avoid exposing `tsgolint` as a public `vite-plus` binary.
- Align the wrapper’s tsgolint resolution with the existing `vp lint` behavior.

 Closes #1800


fengmk2:
Why do we need to export tsgolint? Are there other fix solutions besides this one? 
BTW: We are in the process of removing the oxlint and oxfmt wrappers.

jong-kyung:
> Why do we need to export tsgolint? Are there other fix solutions besides this one? BTW: We are in the process of removing the oxlint and oxfmt wrappers.

The reason for this PR is that `oxlint-tsgolint` is a dependency of `vite-plus`, but from the user project’s perspective it is still transitive. With pnpm, its `tsgolint` bin is not reliably linked into the project root `node_modules/.bin`, which matches the reproduction in #1800 after reinstalling from the lockfile.

That said, I understand the concern about exposing `tsgolint`, especially if the `oxlint` / `oxfmt` wrappers are being removed.

I think there are a couple of possible directions here:
1. Make the Vite+ `oxlint` wrapper set `OXLINT_TSGOLINT_PATH` so oxlint LSP can find Vite+’s pinned `oxlint-tsgolint` without exposing `tsgolint`.
2. Fix discovery upstream in oxlint / oxc-vscode so it can find the `oxlint-tsgolint` dependency installed under `vite-plus`.

fengmk2:
Please first try the `OXLINT_TSGOLINT_PATH` env, which is also the current implement for `vp lint`.

jong-kyung:
> Please first try the `OXLINT_TSGOLINT_PATH` env, which is also the current implement for `vp lint`.

@fengmk2 Thanks for the suggestion. I updated the PR to follow the `vp lint` approach by setting `OXLINT_TSGOLINT_PATH` from the Vite+ `oxlint` wrapper instead of exposing `tsgolint` as a public bin.

Could you please take another look when you have time? 🙇‍♂️

packages/cli/bin/oxlint
```diff
@@ -10,19 +10,37 @@ if (!process.argv.includes('--lsp')) {
...
+  const localBinDir = join(scriptDir, '..', 'node_modules', '.bin');
+  const oxlintTsgolintPackagePath = dirname(tsgolintPackageJsonPath);
+  const projectBinDir = join(oxlintTsgolintPackagePath, '..', '.bin');
+  const candidates = [
+    join(localBinDir, 'tsgolint.exe'),
+    join(localBinDir, 'tsgolint.cmd'),
+    join(projectBinDir, 'tsgolint.exe'),
+    join(projectBinDir, 'tsgolint.cmd'),
+  ];
+  tsgolintBin = candidates.find((candidate) => existsSync(candidate)) ?? tsgolintBin;
```
fengmk2: https://github.com/voidzero-dev/vite-plus/blob/main/packages/cli/src/resolve-lint.ts#L66 Can this same logic code be reused? We have handled the path issue of bun before.

jong-kyung: I’ll extract the shared tsgolint resolution helper and reuse it from both `resolve-lint.ts` and `bin/oxlint` so they stay aligned.

fengmk2:
@jong-kyung I've triggered the pkg.new deployment. Please verify it once the deployment is successful, and I'll merge it if everything looks good.

jong-kyung:
@fengmk2 Thank you for the detailed review as always. I’ll verify this and leave a follow-up comment🙇‍♂️

camc314 (approved):
looks reasonable to me - but is this a bug in how oxlint resolves tsgolint? (in which case should it be fixed upstream)

fengmk2:
@camc314 Based on several case experiences I know, for projects that do not use vite-plus, according to oxlint's requirements, you need to add the tsgolint to the project dependencies to enable typecheck, so there shouldn't be any issues. Currently, this problem only exists in the case of vite-plus.

jong-kyung:
@fengmk2  I verified the pkg.pr.new build for #1811 in a local test project, and `oxc-vscode` no longer shows the Failed to find tsgolint executable warning.

This is my first time testing with `pkg.pr.new`, so please let me know if anything looks off with my setup.

image

fengmk2 (approved):
@jong-kyung Nice job! Thanks

PR #1818, opened 2026-06-11
feat(cli): use version range instead of latest for vite-plus dependencies

fengmk2:
## Problem

`vp create` and `vp migrate` wrote the `latest` dist-tag for `vite-plus` and the `vite`/`vitest` aliases. Dist-tag specs are not re-resolved consistently by package managers, so `vp update vite-plus` was often a no-op, and npm did not record the alias overrides in the lockfile.

## Changes

- Derive a caret range (e.g. `^0.1.24`) from the CLI's own version for the `vite-plus`, `@voidzero-dev/vite-plus-core`, and `@voidzero-dev/vite-plus-test` specs; cli/core/test are published in lockstep. `VP_VERSION` and `VP_OVERRIDE_PACKAGES` overrides keep working.
- `vp migrate` normalizes a legacy `"vite-plus": "latest"` devDependency to the versioned range; deliberate user pins are untouched.
- Snap runner normalizes the new specs to `^<semver>` so snapshots stay stable across releases; regenerated 67 snapshots.
- Updated the upgrade guide: `vp update vite-plus` updates within the range, `vp update --latest vite-plus` or `vp add vite-plus@latest` moves beyond it.
- Test stability: disable commit signing in snap fixture commits and mock `getNpmRegistry` in `package.spec.ts`.

With real ranges npm now records the `vite`/`vitest` alias overrides in the lockfile: the previously failing verification step in the `migration-standalone-npm` snap test now passes.

### DIFF
--- a/docs/guide/upgrade.md
+++ b/docs/guide/upgrade.md
@@ -36,14 +36,20 @@ Update the project dependency with the package manager commands in Vite+:
 vp update vite-plus
 ```
 
-You can also use `vp add vite-plus@latest` if you want to move the dependency explicitly to the latest version.
+Projects created or migrated by Vite+ declare `vite-plus` with a version range (for example `^0.1.24`), so `vp update vite-plus` picks up new releases within that range. To move beyond the range, update the range explicitly:
+
+```bash
+vp update --latest vite-plus
+# or
+vp add vite-plus@latest
+```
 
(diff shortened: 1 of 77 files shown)


fengmk2:
merge into https://github.com/voidzero-dev/vite-plus/pull/1891

PR #1819, opened 2026-06-11
fix(cli): surface git output when initial commit fails

fengmk2:
## Problem

When the initial commit fails during `vp create --git`, the warning always says "Check your git user.name/user.email config" even when the failure has nothing to do with git identity (e.g. a failing pre-commit hook). The actual git error is discarded, making the failure hard to diagnose.

## Fix

`createInitialCommit` now returns the combined stdout/stderr of `git commit`, and both call sites print it under the `Initial commit failed` warning. When identity is the real problem, git's own message is shown verbatim.

Adds a snap test (`create-git-commit-fail`) that scaffolds the builtin `vite:monorepo` template offline and forces a deterministic pre-commit hook failure via `GIT_CONFIG_*` env vars, locking in the surfaced output.

---

> [!NOTE]
> **Low Risk**
> CLI-only UX and return-type change for initial commit; no auth or data-path changes, covered by a snap test.
> 
> **Overview**
> When `vp create --git` cannot create the initial commit, the CLI no longer assumes a **user.name/user.email** problem. **`createInitialCommit`** now returns **`success`** plus combined **`git commit`** stdout/stderr, and both create flows print that text under a short **Initial commit failed** warning.
> 
> A new snap test (**`create-git-commit-fail`**) forces a failing pre-commit hook via **`core.hooksPath`** and asserts the hook message appears while **`.git`** is still created.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b83cfd2828c039a907203d25b239478c2640fa1f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fengmk2:
@cursor review

cpojer (approved):

PR #1832, opened 2026-06-15
fix(cli): unify help alias display

jong-kyung:
## Summary

 Command aliases in CLI help now appear in the command label column instead of as trailing `[aliases: ...]` suffixes.
This makes `vp env`, `vp pm`, and nested `vp pm` help match the existing `vp --help` style, such as `fmt, format`.

This does not change alias execution behavior; it only normalizes help rendering and updates the related parser coverage and snapshots.


jong-kyung:
@codex review

fengmk2 (approved):

PR #1833, opened 2026-06-15
fix(ci): keep upgrade-deps green when rolldown bumps oxc

fengmk2:
## Problem

The nightly **Upgrade Upstream Dependencies** job has failed for several days. The upgrade bumps the rolldown hash to a commit built against a newer oxc release (0.135.0), but nothing updated vp's root `Cargo.toml` oxc pins (stuck at 0.134.0), so the vendored rolldown crates fail to compile (e.g. `oxc_ast::template_element` changed arity between 0.134 and 0.135). With the NAPI binding missing, the unguarded `Format code` step (`vp fmt`) then crashed and hard-failed the whole job, so no PR was created at all.

`main` itself is healthy (oxc 0.134.0 + rolldown `c462c7c5`, consistent), so no product code on main is touched.

## Fix

- `sync-remote` now rewrites the root `Cargo.toml` oxc pins to match `rolldown/Cargo.toml` whenever the rolldown hash moves. The oxc same-version family (incl. `oxc_ast`/`oxc_parser`/`oxc_span`, which rolldown does not declare) follows rolldown's umbrella `oxc` version; independently-pinned crates (`oxc_resolver`, `oxc_sourcemap`) follow their explicit pin; `[patch.*]` is left untouched; formatting and features are preserved. Added a unit test.
- `Format code` is now `continue-on-error` so a broken build still produces a PR, matching the existing build steps.
- Added a fallback note in the in-workflow agent prompt about the oxc/rolldown desync.

## Validation

Ran the new `syncCargoOxcVersions()` against the real `main` `Cargo.toml` and the failing `rolldown@d7f919c1` `Cargo.toml`: the 12-crate oxc family bumps 0.134.0 to 0.135.0, `oxc_index` 4 to 5, independents unchanged, no leftover 0.134.0 pins, line count preserved. Committed test covers the family/index bump, no-op, and `[patch]` scoping.

### DIFF
--- a/.github/workflows/upgrade-deps.yml
+++ b/.github/workflows/upgrade-deps.yml
@@ -82,8 +82,9 @@ jobs:
             - Upgrade script: `./.github/scripts/upgrade-deps.ts`
             - Sync-remote tool: `pnpm tool sync-remote` (source in
               `packages/tools/src/sync-remote-deps.ts`) — clones rolldown/vite into the
-              working tree and merges their pnpm-workspace catalogs into the root
-              `pnpm-workspace.yaml`.
+              working tree, merges their pnpm-workspace catalogs into the root
+              `pnpm-workspace.yaml`, and syncs the root `Cargo.toml` oxc crate
+              versions to match `rolldown/Cargo.toml`.
             - Build-upstream action: `./.github/actions/build-upstream/action.yml`
             - Package manager: `pnpm`. Do NOT downgrade any dep — we want the latest.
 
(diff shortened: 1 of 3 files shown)


fengmk2:
test on https://github.com/voidzero-dev/vite-plus/actions/runs/27520652382

fengmk2:
it works https://github.com/voidzero-dev/vite-plus/pull/1834

Brooooooklyn (approved):

PR #1838, opened 2026-06-15
fix(cli): expose vpr package bin

### LINKED ISSUE #1827: `vpr` is a global-only shim and is unusable in CI (Vercel etc.) — expose it as a package bin or document `vp run` as the CI path
### Summary

`vpr` (the task-runner shorthand for `vp run`) is only installed as a global shim by `vp env setup` / the install script. It is **not** placed in the project's `node_modules/.bin`, because the `vite-plus` npm package only ships `vp`, `oxfmt`, and `oxlint` bins. As a result, any build/CI environment that does a clean install without the global Vite+ shims — e.g. Vercel's build container — cannot run a `vpr <task>` command and fails with `command not found` (exit 127).

### Environment

- `vite-plus` / `@voidzero-dev/vite-plus-core` `0.1.24`
- Clean dependency install (no global `vp env setup`), e.g. Vercel build image, generic CI runner.

### Reproduction

1. In a project using Vite+ tasks, set the platform build command to `vpr build` (a task defined under `run.tasks` in `vite.config.ts`).
2. Run a clean install + build the way a CI container does (only `node_modules/.bin` on `PATH`, no global Vite+ shim dir).
3. The build fails: `vpr: command not found` (exit 127).

Confirmed against the package contents: `node_modules/vite-plus/package.json` `bin` lists only `oxfmt`, `oxlint`, `vp`; there is no `vpr` entry, and nothing creates `node_modules/.bin/vpr`. `vpr` is implemented in the global CLI crate (`crates/vite_global_cli/src/commands/vpr.rs`) and resolved through global shim dispatch (`crates/vite_global_cli/src/shim/mod.rs`), so it only exists where the global shims were installed.

### Current workaround

Use the `vp run` subcommand, which **is** present in `node_modules/.bin` and runs the same `run.tasks` task:

```
vp run build   # works in CI; equivalent to `vpr build`
```

The docs already confirm these are equivalent ("`vpr` is available as a standalone shorthand for `vp run`", `docs/guide/run.md`), and the CI guide's examples all use `vp run`. But the asymmetry isn't called out anywhere, so it's an easy trap when setting a platform's build command to `vpr <task>`.

### Proposed fix (happy to send a PR)

Either or both:

1. **Expose `vpr` as a package bin.** Add a `bin/vpr` entry to the `vite-plus` package alongside the existing `bin/vp`, so `vpr` lands in `node_modules/.bin` on a normal install and behaves identically in CI. (`vp` is already shipped this way as `./bin/vp`.)
2. **Document the limitation.** In `docs/guide/ci` and `docs/guide/run`, explicitly note that `vpr` is a global shim that may be absent in clean CI installs, and that `vp run <task>` is the CI-safe equivalent to use in build commands.

Option 1 removes the foot-gun entirely; option 2 is a cheap stopgap. Happy to open a PR for whichever direction you prefer.

wyf027:
## Summary
- add a package-level vpr bin so clean installs get node_modules/.bin/vpr
- make the vpr wrapper delegate to the existing CLI as vp run
- add a local snap fixture matching the existing global vpr coverage

Fixes #1827.

## Verification
- node -e "const pkg=require('./packages/cli/package.json'); if (pkg.bin.vpr !== './bin/vpr') throw new Error('missing vpr bin'); console.log(pkg.bin.vpr)"
- test -x packages/cli/bin/vpr
- node -e "const fs=require('fs'); const text=fs.readFileSync('packages/cli/bin/vpr','utf8'); if (!text.includes('process.argv.splice(2, 0, \'run\')')) throw new Error('vpr does not inject run'); if (!text.includes('import \'../dist/bin.js\'')) throw new Error('vpr does not load CLI'); console.log('vpr wrapper ok')"
- pnpm exec oxfmt --check packages/cli/bin/vpr packages/cli/package.json packages/cli/snap-tests/command-vpr/steps.json packages/cli/snap-tests/command-vpr/package.json

Note: I attempted pnpm -F vite-plus build-ts before running the snap test, but this local checkout cannot build because the vendored rolldown workspace has no generated dist/native binding after sync-remote failed with git index-pack errors. The change itself is covered by static checks and mirrors the existing global command-vpr snap fixture.

### DIFF
--- a/packages/cli/bin/vpr
+++ b/packages/cli/bin/vpr
@@ -6,4 +6,4 @@ if (module.enableCompileCache) {
 }
 
 process.argv.splice(2, 0, 'run');
-import '../dist/bin.js';
+await import('../dist/bin.js');
--- a/packages/cli/bin/vpr
+++ b/packages/cli/bin/vpr
@@ -1,0 +1,9 @@
+#!/usr/bin/env node
+
+import module from 'node:module';
+if (module.enableCompileCache) {
+  module.enableCompileCache();
+}
+
+process.argv.splice(2, 0, 'run');
+import '../dist/bin.js';
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -16,7 +16,8 @@
   "bin": {
     "oxfmt": "./bin/oxfmt",
     "oxlint": "./bin/oxlint",
-    "vp": "./bin/vp"
+    "vp": "./bin/vp",
+    "vpr": "./bin/vpr"
   },
   "files": [
     "AGENTS.md",
--- a/packages/cli/snap-tests/command-vpr/args.mjs
+++ b/packages/cli/snap-tests/command-vpr/args.mjs
@@ -1,0 +1,4 @@
+// Print arguments passed to this script, one per line
+for (const arg of process.argv.slice(2)) {
+  console.log(arg);
+}
--- a/packages/cli/snap-tests/command-vpr/package.json
+++ b/packages/cli/snap-tests/command-vpr/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "command-vpr",
+  "version": "1.0.0",
+  "scripts": {
+    "hello": "node args.mjs hello from script",
+    "greet": "node args.mjs greet"
+  },
+  "packageManager": "pnpm@10.19.0"
+}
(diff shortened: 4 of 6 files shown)


fengmk2:
@codex review

wyf027:
Follow-up pushed.

- Fixed the `vpr` wrapper so it injects `run` before loading the CLI entrypoint.
- Updated the local `command-vpr` snapshot to match the actual `vp run` help output.
- Latest CI has no remaining failures on this PR.

fengmk2:

Issue #1586, opened 2026-05-15
`vp pack --exe` fails with "Failed to import module @tsdown/exe" when `@tsdown/exe` is installed as documented


Saeris:
### Describe the bug

**Summary**

The [Pack documentation](https://viteplus.dev/guide/pack) advertises standalone executable builds as a supported feature, but using it requires non-obvious workarounds that contradict or go beyond what the docs describe. The root cause is a cascading package resolution problem: `@tsdown/exe` has a hard peer dependency on `tsdown`, but Vite+ bundles tsdown internally under `@voidzero-dev/vite-plus-core` and doesn't expose it as a resolvable top-level package.

While the bug here might better be resolved in the tsdown repository, I'm raising it as an issue here because of the confusing documentation and to flag it as an interop concern so that you can adjust your strategy during this early stage of Vite+'s development. Resolution issues like this could easily pop up in other places within the Vite+ toolchain.

> [!Note]
>
> In general, I have observed that LLMs in particular have issues with the way Vite+ bundles together all of these tools, not just tsdown. Claude for example often goes down rabbit holes attempting to resolve particular packages like vitest. Numerous times I've had to teach it workarounds because of this architectural decision. Personally, I don't think it's a bad idea, I just want to report what I have observed is a frequent pain point.
>
> A related example can be observed trying to use `@tsdown/css`. Moving forward it may be worth investigating how to better evolve these tools together. Food for thought.

---

**What I Tried**

1. **Adding `"tsdown": "npm:vite-plus@latest"` to `resolutions`** — `vite-plus` does expose `./pack` and `./internal` subpaths, but its `./internal` export is Vite's internal module, not tsdown's. The `tsdown/internal` subpath required by `@tsdown/exe` exports different symbols (`fsExists`, `fsRemove`, `Logger`) that are not present in `vite-plus/internal`.

2. **Adding `"tsdown": "0.22.0"` to `resolutions`** — Yarn 4's `resolutions` only overrides packages already in the dependency graph. Since no explicit dependency declares `tsdown` (it's an unsatisfied peer of `@tsdown/exe`), the resolution is never applied and `tsdown` is never installed at the top level.

3. **Using `packageExtensions` in `.yarnrc.yml` to mark the peer as optional** — This silenced the peer warning but didn't make `tsdown` resolvable at runtime.

[...]

PR #1839, opened 2026-06-15
docs(pack): document exe peer dependency

wyf027:
## Summary

- document that `pack.exe` users must install both `@tsdown/exe` and `tsdown`
- explain why Vite+'s bundled tsdown does not satisfy `@tsdown/exe` peer resolution

Addresses #1586 with the documented-workaround path described in the issue.

## Verification

- `pnpm exec oxfmt --check docs/guide/pack.md`
- `node -e "const fs=require('fs'); const text=fs.readFileSync('docs/guide/pack.md','utf8'); for (const s of ['vp add -D @tsdown/exe tsdown','tsdown/internal','peer dependency']) if (!text.includes(s)) throw new Error('missing '+s); console.log('pack exe docs ok')"`

### DIFF
--- a/docs/guide/pack.md
+++ b/docs/guide/pack.md
@@ -30,12 +30,12 @@ Use it for:
 - [standalone executables](https://tsdown.dev/options/exe#executable)
 
 ```ts [vite.config.ts]
-import { defineConfig } from 'vite-plus';
+import { defineConfig } from "vite-plus";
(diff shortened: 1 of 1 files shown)


fengmk2:
This problem should not be solved at the documentation level, we need to solve it at the `vp pack` command.

PR #1842, opened 2026-06-15
feat: allow vp config opt-outs

### LINKED ISSUE #1311: Ability to turn off stuff from `vite config`
### Description

I find myself a lot reverting changes `vite config` does when running `pnpm install`. The docs aren't clear on that, so I don't know what this script does, but for me, it only caused problems, which is why I removed it from my `prepare` scripts. It would've been a lot nicer if

- I was able to turn off changing my AGENT.md (I intentionally formed it the way it is and wan't to keep it)
- I was able to turn off pre commit hook creation (I intentionally removed them and don't want them coming back all the time)

### Suggested solution

two flags

- `--noHooks`
- `--noAgentMd`

### Alternative

It automatically notices that I changed it and want to have it this way, so it doesn't annoy me anymore.

### Additional context

_No response_

### Validations

wyf027:
## Summary

- add `vp config --no-hooks` to skip hook installation
- add `vp config --no-agent` to skip existing coding agent instruction updates
- document the new opt-out flags alongside `--hooks-only` and `VITE_GIT_HOOKS=0`

Closes voidzero-dev/vite-plus#1311.

## Verification

- `pnpm exec oxfmt --check packages/cli/src/config/bin.ts docs/guide/commit-hooks.md`
- static source/docs assertion for the new config flags
- `mri` parsing assertion for `--no-hooks`, `--no-agent`, and `--hooks-only`

Could not run `pnpm -F vite-plus test packages/cli/src/config/__tests__/hooks.spec.ts` in this checkout because the local install does not expose the `vitest` binary.


packages/cli/src/config/bin.ts
```diff
@@ -18,7 +18,7 @@ import { install } from './hooks.ts';
 
 async function main() {
   const args = mri(process.argv.slice(3), {
-    boolean: ['help', 'hooks-only'],
+    boolean: ['help', 'hooks-only', 'hooks', 'agent'],
```
fengmk2: I think we should remove `--hooks-only` instead of keeping it as a compatibility alias.

With the new `--no-hooks` and `--no-agent` flags, the `vp config` behavior can be modeled as two independent setup steps:

```bash
vp config                         # install hooks + update agent instructions
vp config --no-hooks              # update agent instructions only
vp config --no-agent              # install hooks only
vp config --no-hooks --no-agent   # skip both
```

Update tests to cover:
   - `--no-hooks`
   - `--no-agent`
   - `--no-hooks --no-agent`

This makes the API easier to explain: `vp config` performs two independent actions, and each action has a matching `--no-*` opt-out flag.

wyf027:
Addressed the review feedback.

- Removed `--hooks-only` instead of keeping it as an alias.
- Modeled `vp config` as two independent setup steps with `--no-hooks` and `--no-agent`.
- Updated the migration internal call to use `--no-agent`.
- Added snap coverage for `--no-hooks`, `--no-agent`, and `--no-hooks --no-agent`.

The main CI is passing; the remaining `Deploy Docs Preview` failure is unrelated to this PR content because the workflow has no `VOID_TOKEN` available on this fork PR.

fengmk2 (approved):
@leno23 Thanks!

PR #1845, opened 2026-06-15
docs(team): update Christoph X profile

jong-kyung:
Christoph Nakazawa’s team profile now links to his current X account, `cnakazawa`, instead of the old `cpojer` URL.


fengmk2 (approved):

cpojer:
Oh thank you! Twitter said they would keep the redirects, but they didn't 😭

PR #1852, opened 2026-06-16
feat: default yarn nodeLinker to pnpm

fengmk2:
## What

Change the default yarn linker written by `vp create` and `vp migrate` from `node-modules` to `pnpm`.

## Why

The `pnpm` linker keeps real on-disk `node_modules` entries (backed by a content-addressable store), so it avoids Plug'n'Play's zip entries that break `@oxlint/migrate`'s `fileURLToPath` resolution, while giving strict resolution and disk savings.

## Notes

- `nodeLinker` is only set when absent, so projects that already pin a linker keep their value.
- Migrating a yarn project without an explicit linker now switches it to pnpm's strict resolution (undeclared transitive deps may surface as `Cannot find module`).

### DIFF
--- a/packages/cli/snap-tests-global/migration-monorepo-yarn4/snap.txt
+++ b/packages/cli/snap-tests-global/migration-monorepo-yarn4/snap.txt
@@ -71,7 +71,7 @@ export default defineConfig({
 }
 
 > cat .yarnrc.yml # check .yarnrc.yml
-nodeLinker: node-modules
+nodeLinker: pnpm
 catalog:
   vite: npm:@voidzero-dev/vite-plus-core@latest
   vitest: npm:@voidzero-dev/vite-plus-test@latest
--- a/packages/cli/src/create/__tests__/utils.spec.ts
+++ b/packages/cli/src/create/__tests__/utils.spec.ts
@@ -352,12 +352,12 @@ describe('renameFiles', () => {
 
   it('renames `_npmrc` and `_yarnrc.yml`', () => {
     write('_npmrc', 'auto-install-peers=true\n');
-    write('_yarnrc.yml', 'nodeLinker: node-modules\n');
+    write('_yarnrc.yml', 'nodeLinker: pnpm\n');
     renameFiles(projectDir);
     expect(exists('_npmrc')).toBe(false);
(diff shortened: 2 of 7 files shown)


fengmk2:
## Why it fails under yarn's `nodeLinker: pnpm` but not under pnpm

Core does a runtime `import.meta.resolve('@typescript/native-preview')` (a phantom dep it never declares). Both pnpm and yarn's pnpm linker use a store + symlinks, so why does this resolve under pnpm but break under yarn? One directory: **pnpm has a global hoist dir inside the store; yarn's pnpm linker doesn't.**

**pnpm**
```
node_modules/
  .pnpm/
    @voidzero-dev+vite-plus-core@x.y.z/
      node_modules/
        @voidzero-dev/vite-plus-core/   <- real files
        <core's declared deps>          -> symlinks
    @typescript+native-preview@a.b.c/
      node_modules/@typescript/native-preview/
    node_modules/                        <- catch-all hoist (hoist-pattern=['*'])
      @typescript/native-preview         -> symlink
      ...~every package in the graph...
  <project's direct deps>                -> symlinks into .pnpm
```
Resolving native-preview from core walks up to `.pnpm/node_modules/` and finds it. pnpm is strict only at the project root; deeper in the store, any dep can reach almost any package (a deliberate compat compromise).

**yarn `nodeLinker: pnpm`**
```
node_modules/
  .store/
    @voidzero-dev-vite-plus-core-virtual-f02751107a/
      node_modules/
        @voidzero-dev/vite-plus-core/   <- real files
        <core's declared deps>          -> symlinks
    (no global .store/node_modules catch-all)
  <project's direct deps>               -> symlinks into .store
```
Same shape, but no catch-all. Each package sees only its declared (+ injected peer) deps, so the walk-up finds nothing and resolution fails.

**Takeaway:** yarn's `nodeLinker: pnpm` is actually *stricter* than pnpm, so it breaks phantom deps that real pnpm quietly hides, here `vp pack`. `node-modules` hid the problem via full hoisting.

**Fix:** declare `@typescript/native-preview` in `@voidzero-dev/vite-plus-core` (likely an `optionalDependency`); that's correct under every layout (yarn pnpm linker, PnP, pnpm). `packageExtensions` is a yarn-only band-aid.

PR #1863, opened 2026-06-18
fix(create): drop src-only tsconfig include

### LINKED ISSUE #1308: Remove `"include": ["src"]` from `tsconfig.json` in `vp create vite:library` template
### Description

Projects generated by `vp create vite:library` include a `tsconfig.json` like the following:

```json
{
  "compilerOptions": {
    "target": "esnext",
    "lib": ["es2023"],
    "moduleDetection": "force",
    "module": "nodenext",
    "moduleResolution": "nodenext",
    "resolveJsonModule": true,
    "types": ["node"],
    "strict": true,
    "noUnusedLocals": true,
    "declaration": true,
    "noEmit": true,
    "allowImportingTsExtensions": true,
    "esModuleInterop": true,
    "isolatedModules": true,
    "verbatimModuleSyntax": true,
    "skipLibCheck": true
  },
  "include": ["src"]
}
```

The key issue is that the `include` field is limited to the `src` directory only. This causes the following problems.

### Problem 1: Cannot auto-import items from `src` directory files outside of `src`

Files outside of the `src` directory cannot auto-import functions and other items from the `src` directory. For example, files in the `tests` directory cannot auto-import functions defined in `src`.

Steps to reproduce:

1. `git clone https://github.com/mizdra-sandbox/repro-vp-lib-tsconfig-is-missing`
2. `cd repro-vp-lib-tsconfig-is-missing`
3. `code --profile-temp .` (using VS Code)
4. Open `tests/index.test.ts`
5. Place the cursor after `fn` and trigger completion
    - Expected behavior: `fn` should be suggested
    - Actual behavior: `fn` is not suggested

Image

tsserver treats files in the `src` directory as part of the `tsconfig.json` project, but treats files in the `tests` directory as part of an "Inferred Project". An Inferred Project is one that tsserver automatically assigns to files that have no corresponding `tsconfig.json`. Since tsserver cannot auto-import items between files belonging to different projects, files in `tests` cannot auto-import items from files in `src`.

### Problem 2: Expected compiler options are not applied to files outside `src`

The `tsconfig.json` includes `"noUnusedLocals": true`. While the `src` directory is affected by this option, the `tests` directory, `vite.config.ts`, and similar files are not. As a result, unused local variables in those files are not reported as errors.

Steps to reproduce:

1. `git clone https://github.com/mizdra-sandbox/repro-vp-lib-tsconfig-is-missing`
2. `cd repro-vp-lib-tsconfig-is-missing`
3. `code --profile-temp .` (using VS Code)
4. Open `tests/index.test.ts`
5. Add code containing an unused local variable (e.g. `const unused = 1;`)
6. Run `vp check`
   ```console
   $ vp check
   
   pass: All 7 files are correctly formatted (158ms, 14 threads)
   error: Lint or type issues found
   × typescript(TS2304): Cannot find name 'fn'.
     ╭─[tests/index.test.ts:3:1]
   2 │ const unused = 1;
   3 │ fn();
     · ──
     ╰────
   
   Found 1 error and 0 warnings in 3 files (140ms, 14 threads)
   ```
   - No `noUnusedLocals` error is reported for `const unused = 1;`
   - Expected behavior: if `noUnusedLocals` were applied, an "unused variable" error for `unused` should be reported in addition to the "cannot find name" error for `fn`

Users would reasonably expect the same compiler options applied to `src` to also apply to files in `tests`, `vite.config.ts`, and similar locations. This behavior is therefore likely to cause confusion.

### Suggested solution

Remove the `"include"` field from `tsconfig.json` in projects generated by `vp create vite:library`. Omitting `"include"` is equivalent to specifying `"include": ["**/*"]`, which causes all TypeScript files to be treated as part of the `tsconfig.json` project. This resolves both problems described above.

### Alternative

None.

### About projects generated by `vp create vite:application`

The `tsconfig.json` in projects generated by `vp create vite:application` also contains `"include": ["src"]`, so the same problems occur. The `"include"` field should be removed there as well to broaden the scope.

### Compiler options used by Inferred Projects

Files in `tests`, `vite.config.ts`, and similar locations belong to an Inferred Project, so Inferred Project settings are applied to them. In tsserver, the options used are either the values sent by the editor via a `setCompilerOptionsForInferredProjects` request, or the default values defined in TypeScript's `src/server/project.ts`.

- https://github.com/microsoft/vscode/blob/8d415912364f9b29d76ae6bfc59234e59e03dbc7/extensions/typescript-language-features/src/typescriptServiceClient.ts#L622-L630, https://github.com/microsoft/vscode/blob/8d415912364f9b29d76ae6bfc59234e59e03dbc7/extensions/typescript-language-features/src/tsconfig.ts#L23
  - (when tsserver version >= 5.4.0) `"module": "Preserve"`, `"moduleResolution": "Bundler"`, `"target": "ES2022"`, `"allowImportingTsExtensions": true`, ...
  - (when tsserver version < 5.4.0) `"module": "ESNext"`, `"moduleResolution": "Bundler"`, `"target": "ES2022"`, `"allowImportingTsExtensions": true`, ...
- https://github.com/microsoft/TypeScript/blob/7b8cb3bdf82f400642b73173f941335775d6f730/src/server/project.ts#L590-L594, https://github.com/microsoft/TypeScript/blob/7b8cb3bdf82f400642b73173f941335775d6f730/src/services/services.ts#L1384
  - `"module": "ESNext"`, `"moduleResolution": "Bundler"`, `"target": "ESNext"`, `"allowImportingTsExtensions": false`, ...

In oxlint, the values defined in the `CreateInferredProjectProgram` function are used.

- https://github.com/oxc-project/tsgolint/blob/757a9968aeb3ab543515738d1a26aa95176b84ee/internal/utils/create_program.go#L124-L139
  - `"module": "ESNext"`, `"moduleResolution": "Bundler"`, `"target": "ES2022"`, `"allowImportingTsExtensions": true`, ...

### Side effects of broadening the scope of `tsconfig.json`

Broadening the scope of `tsconfig.json` has some side effects.

#### Side effect 1: `"module": "nodenext"` and `"moduleResolution": "nodenext"` are applied to files outside `src`

The compiler options intended for `src` will also be applied to files in other directories. Of particular note are `"module": "nodenext"` and `"moduleResolution": "nodenext"`. It is worth considering whether applying these to the `tests` directory, config files (e.g. `vite.config.ts`), and a `scripts` directory (e.g. https://github.com/vitest-dev/vitest/tree/main/scripts) is acceptable.

In my view, applying these options to `tests` and config files should not be a problem. The `src` directory is built by `vp pack` and `tests` is run by `vp test`; both are backed by Vite, which supports building code that assumes `"module": "nodenext"` and `"moduleResolution": "nodenext"`.

Applying these options to config files should also be fine. Config files are either executed via Node.js Type Stripping (e.g. `prettier.config.ts`) or bundled before execution (e.g. `vite.config.ts`), and both execution models support code that assumes these module settings. The same applies to a `scripts` directory.

Therefore, this side effect is not a significant concern.

#### Side effect 2: Per-directory compiler options become harder to configure

In some cases, users may want to apply different compiler options per directory. For example, they may want `"allowImportingTsExtensions": true` for `src` and `tests`, but not for `scripts`.

If that is needed, a "solution-style" `tsconfig.json` can be added later to achieve per-directory configuration.

- https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#support-for-solution-style-tsconfigjson-files

```json
// tsconfig.json
{
  "files": [],
  "references": [
    { "path": "./tsconfig.default.json" }, // For `src` and `tests` directories
    { "path": "./tsconfig.scripts.json" } // For `scripts` directory
  ]
}
```

```json
// tsconfig.default.json
{
  "compilerOptions": {
    "target": "esnext",
    "lib": ["es2023"],
    "moduleDetection": "force",
    "module": "nodenext",
    "moduleResolution": "nodenext",
    "allowImportingTsExtensions": true
    // ...
  }
}
```

```json
// tsconfig.scripts.json
{
  "compilerOptions": {
    "target": "esnext",
    "lib": ["es2023"],
    "moduleDetection": "force",
    "module": "nodenext",
    "moduleResolution": "nodenext"
    // "allowImportingTsExtensions": true,
    // ...
  }
}
```

Since a workaround exists, this side effect is not a significant concern.

### Validations

wyf027:
## Summary

- remove generated `tsconfig.json` `include: ["src"]` when Vite+ creates builtin application/library projects
- apply the same cleanup to the default app/library created by `vite:monorepo`
- add unit coverage for the src-only include cleanup helper

Closes #1308

## Verification

- `git diff --check`
- `pnpm exec oxnode tmp/check-remove-src-include.ts` (temporary smoke script; removed before commit)

## Notes

I could not run `pnpm -F vite-plus test src/create/__tests__/utils.spec.ts` in this checkout because local workspace dependencies are not built: Vitest resolves the local Vite package and fails on missing `vite/dist/vite/node/index.js`; trying `pnpm -F vite build` then fails on missing local `rolldown/packages/rolldown/dist/cli.mjs`.

### DIFF
--- a/packages/cli/snap-tests-global/new-create-vite/snap.txt
+++ b/packages/cli/snap-tests-global/new-create-vite/snap.txt
@@ -2,6 +2,9 @@
 > ls vite-plus-application/package.json # check package.json
 vite-plus-application/package.json
 
+> node -e "const fs = require('node:fs'); const tsconfig = JSON.parse(fs.readFileSync('vite-plus-application/tsconfig.json', 'utf8')); if ('include' in tsconfig) { console.error('ERROR: src-only include still exists'); process.exit(1); } console.log('src-only include removed');" # check default tsconfig include removed
+src-only include removed
+
 > test -f vite-plus-application/.vscode/settings.json && echo '.vscode/settings.json exists' # check VS Code settings created
 .vscode/settings.json exists
 
--- a/packages/cli/snap-tests-global/new-create-vite/steps.json
+++ b/packages/cli/snap-tests-global/new-create-vite/steps.json
@@ -5,6 +5,7 @@
       "ignoreOutput": true
     },
     "ls vite-plus-application/package.json # check package.json",
+    "node -e \"const fs = require('node:fs'); const tsconfig = JSON.parse(fs.readFileSync('vite-plus-application/tsconfig.json', 'utf8')); if ('include' in tsconfig) { console.error('ERROR: src-only include still exists'); process.exit(1); } console.log('src-only include removed');\" # check default tsconfig include removed",
     "test -f vite-plus-application/.vscode/settings.json && echo '.vscode/settings.json exists' # check VS Code settings created",
     "test -f vite-plus-application/.vscode/extensions.json && echo '.vscode/extensions.json exists' # check VS Code extensions created",
     "node -e \"const { spawnSync } = require('node:child_process'); const file = '.vscode/settings.json'; const result = spawnSync('git', ['-C', 'vite-plus-application', 'check-ignore', '--no-index', file], { stdio: 'ignore' }); const status = result.status; if (status === 1) { console.log(file + ' trackable'); } else { console.log(status === 0 ? 'ERROR: ' + file + ' ignored' : 'ERROR: git check-ignore failed with status ' + status); process.exit(status || 1); }\" # check VS Code settings are trackable",
--- a/packages/cli/src/create/__tests__/utils.spec.ts
+++ b/packages/cli/src/create/__tests__/utils.spec.ts
@@ -1,8 +1,8 @@
-import fs from 'node:fs';
-import os from 'node:os';
-import path from 'node:path';
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
 
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
 import {
   deriveDefaultPackageName,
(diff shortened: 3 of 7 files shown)


wyf027:
Updated the branch to parse scaffolded tsconfig.json with the existing JSONC parser before removing the default src-only include, and added a regression test for comments/trailing commas.\n\nLocal checks:\n- git diff --check\n- pnpm exec oxfmt --check packages/cli/src/utils/json.ts packages/cli/src/create/utils.ts packages/cli/src/create/__tests__/utils.spec.ts\n- pnpm exec oxlint packages/cli/src/utils/json.ts packages/cli/src/create/utils.ts packages/cli/src/create/__tests__/utils.spec.ts\n- pnpm exec tsc -p packages/cli/tsconfig.json --noEmit\n\nNote: the focused Vitest command still cannot start in this local checkout because Vitest resolves a missing local vite/dist/vite/node/index.js module after pnpm install.

fengmk2 (changes requested):
Could you add a snap test that can reproduce the issue? Also, please make sure CI passes before setting the PR to ready for review status.

wyf027:
Pushed a follow-up formatting fix for the vp check failures in packages/cli/src/create/__tests__/utils.spec.ts, packages/cli/src/create/utils.ts, and packages/cli/src/utils/json.ts. I reproduced the formatter with the root fmt config via oxfmt 0.55.0; targeted oxfmt --check and git diff --check pass locally. The local vp binary itself still cannot run until the CLI package dist is rebuilt.

wyf027:
Added the requested snap coverage in `274b7b9`: the global `new-create-vite` snap test now verifies the generated default application `tsconfig.json` no longer keeps the src-only `include`.\n\nLocal validation:\n- `node -e` parse check for `packages/cli/snap-tests-global/new-create-vite/steps.json`\n- `git diff --check`\n\nI also attempted the focused snap test, but this local checkout cannot build/run global snap tests because `packages/cli/dist` is absent and `pnpm -F vite-plus build-ts` currently fails resolving `rolldown/dist/index.mjs` from the local pnpm install. CI should now validate the snap job.

fengmk2:

PR #1864, opened 2026-06-18
fix(migrate): summarize skipped git hooks

### LINKED ISSUE #1854: vp migrate: precommit migration is silently skipped for lefthook/simple-git-hooks/yorkie even when opted-in
### Summary

When running `vp migrate` and answering **yes** to the git-hooks / precommit prompt, the migration silently does nothing for projects that use **lefthook** (or `simple-git-hooks` / `yorkie`). The existing lefthook setup is left fully in place, no `staged` config is written to `vite.config.ts`, and `.vite-hooks/` / `core.hooksPath` are never set up. The only signal is a single `⚠ Detected lefthook — skipping git hooks setup` warning line, which is very easy to miss.

The net result is a "silent-ish" no-op: the user believes precommit was migrated to Vite+, but in reality the old lefthook + old tools keep running, and Vite+ precommit is never configured.

### Environment

- `vite-plus@^0.1.24`
- bun `1.3.14`, monorepo (turbo)
- Pre-existing precommit: **lefthook** (running `oxlint --fix` / `oxfmt`)

### Steps to reproduce

1. In a repo that already uses lefthook (`lefthook` in devDependencies, `lefthook.yml`, `prepare: "lefthook install"`).
2. Run `vp migrate`.
3. Answer **yes** to the precommit / git-hooks prompt.

### Expected

Answering yes to precommit migrates lefthook the same way husky is handled — i.e. either:

- migrate it: write `staged` config to `vite.config.ts`, run `vp config` (generate `.vite-hooks/`, set `core.hooksPath`, rewrite `prepare`), and remove the old lefthook setup (`lefthook.yml`, devDependency, old `.git/hooks/pre-commit`); **or**
- if intentionally unsupported, make the skip impossible to miss (e.g. surface it in the final migration summary under a "skipped / manual action required" section, not just an inline warning).

### Actual

- `core.hooksPath` not set → git still uses `.git/hooks/`, which still calls lefthook.
- `lefthook.yml`, `prepare: "lefthook install"`, and the `lefthook` devDependency all remain.
- No `.vite-hooks/`, no `staged` config in `vite.config.ts`.
- Only a single inline `⚠ Detected lefthook — skipping git hooks setup. Please configure git hooks manually.` warning is printed.

### Root cause (from reading the bundled `dist/` of 0.1.24)

These references are from the **minified** bundle (`node_modules/vite-plus/dist/...`), so symbol/line numbers are approximate — please cross-check against source:

- `agent-*.js` defines two lists:
  ```js
  const OTHER_HOOK_TOOLS = ["simple-git-hooks", "lefthook", "yorkie"];
  const REPLACED_HOOK_PACKAGES = ["husky", "lint-staged"]; // only these are ever removed
  ```
- `preflightGitHooksSetup()` scans deps and, if it finds any `OTHER_HOOK_TOOLS` entry, returns a reason string (a "skip" signal):
  ```js
  for (const tool of OTHER_HOOK_TOOLS)
    if (deps?.[tool] || prodDeps?.[tool] || pkgContent[tool])
      return `Detected ${tool} — skipping git hooks setup. Please configure git hooks manually.`;
  ```
- In `migration/bin.js` `collectMigrationPlan()`, even when the user answered yes, that reason forces `shouldSetupHooks = false`:
  ```js
  let shouldSetupHooks = await promptGitHooks(options); // user said yes
  if (shouldSetupHooks) {
    const reason = preflightGitHooksSetup(rootDir, packageManager);
    if (reason) { log.warn(`⚠ ${reason}`); shouldSetupHooks = false; }
  }
  ```
- As a consequence, both follow-ups are skipped:
  - `const skipStagedMigration = !plan.shouldSetupHooks;` → no `staged` config is written.
  - `if (plan.shouldSetupHooks) { installGitHooks(...) }` → never runs (`vp config` equivalent skipped).
- lefthook is never removed because removal only targets `REPLACED_HOOK_PACKAGES` (`husky` / `lint-staged`).

So this appears to be **intentional** (only husky + lint-staged are auto-migrated), but the UX makes it look like the precommit migration succeeded when it didn't.

### Suggestions

1. When the user explicitly opted into precommit but preflight skips it, surface the skip prominently in the final migration summary (and at exit), not just as an inline warning.
2. For detected lefthook/simple-git-hooks/yorkie, offer an opt-in prompt to migrate (remove the old tool + set up Vite+ hooks), mirroring the husky migration path.
3. At minimum, document that lefthook / simple-git-hooks / yorkie users must migrate precommit manually.

### Manual migration that worked (for reference)

1. Add `staged` to `vite.config.ts` (translating `lefthook.yml`'s `oxlint --fix` / `oxfmt` to `vp lint --fix` / `vp fmt`).
2. `package.json`: change `prepare` to `vp config`, remove the `lefthook` devDependency.
3. Delete `lefthook.yml` and the old `.git/hooks/pre-commit`.
4. Run `vp config` to generate `.vite-hooks/` and set `core.hooksPath`.
5. Reinstall to drop lefthook from the lockfile (had to manually `rm -rf node_modules/lefthook node_modules/.bin/lefthook` afterward).

---

_Note: I'm not a native English speaker — this report was written with translation assistance, so apologies for any awkward phrasing._

wyf027:
## Summary

- keep the inline preflight warning when git hook migration is skipped
- carry the skip reason into the migration plan
- show the skipped hook setup in the final Manual follow-up summary

Closes #1854

## Verification

- `git diff --check`

## Notes

`pnpm -F vite-plus test src/migration/__tests__/migrator.spec.ts -- --runInBand` could not start in this checkout because Vitest resolves the local Vite workspace package and `vite/dist/vite/node/index.js` has not been built. This is the same local workspace-dist blocker seen while preparing #1863.

### DIFF
--- a/packages/cli/snap-tests-global/migration-existing-husky-v8-hooks/snap.txt
+++ b/packages/cli/snap-tests-global/migration-existing-husky-v8-hooks/snap.txt
@@ -5,6 +5,8 @@
 ◇ Migrated . to Vite+
 • Node <semver>  pnpm <semver>
 • 1 config update applied
+→ Manual follow-up:
+  - Git hooks were not migrated: Detected husky <9.0.0 — please upgrade to husky v9+ first, then re-run migration.
 
 > cat package.json # husky/lint-staged should remain in devDeps, prepare should stay as husky
 {
--- a/packages/cli/snap-tests-global/migration-existing-husky-v8-multi-hooks/snap.txt
+++ b/packages/cli/snap-tests-global/migration-existing-husky-v8-multi-hooks/snap.txt
@@ -5,6 +5,8 @@
 ◇ Migrated . to Vite+
 • Node <semver>  pnpm <semver>
 • 1 config update applied
+→ Manual follow-up:
+  - Git hooks were not migrated: Detected husky <9.0.0 — please upgrade to husky v9+ first, then re-run migration.
 
 > cat package.json # husky/lint-staged should remain in devDeps, prepare should stay as husky
 {
--- a/packages/cli/snap-tests-global/migration-husky-latest-dist-tag/snap.txt
+++ b/packages/cli/snap-tests-global/migration-husky-latest-dist-tag/snap.txt
@@ -5,6 +5,8 @@
 ◇ Migrated . to Vite+
 • Node <semver>  pnpm <semver>
 • 1 config update applied
+→ Manual follow-up:
+  - Git hooks were not migrated: Could not determine husky version from "latest" — please specify a semver-compatible version (e.g., "^9.0.0") and re-run migration.
 
 > cat package.json # husky should still be in devDeps
 {
--- a/packages/cli/snap-tests-global/migration-husky-v8-preserves-lint-staged/snap.txt
+++ b/packages/cli/snap-tests-global/migration-husky-v8-preserves-lint-staged/snap.txt
@@ -5,6 +5,8 @@
 ◇ Migrated . to Vite+
 • Node <semver>  pnpm <semver>
 • 1 config update applied
+→ Manual follow-up:
+  - Git hooks were not migrated: Detected husky <9.0.0 — please upgrade to husky v9+ first, then re-run migration.
 
 > cat package.json # lint-staged config should still be in package.json
 {
--- a/packages/cli/snap-tests-global/migration-lint-staged-ts-config/snap.txt
+++ b/packages/cli/snap-tests-global/migration-lint-staged-ts-config/snap.txt
@@ -5,6 +5,8 @@
 ◇ Migrated . to Vite+
 • Node <semver>  pnpm <semver>
 • 1 config update applied
+→ Manual follow-up:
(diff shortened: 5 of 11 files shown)


wyf027:
Updated the migration snap fixtures for the skipped git-hook manual follow-up output shown by CI.\n\nValidation:\n- git diff --check\n- Verified the changed files match the 10 failing snap fixtures reported in the CI diff\n\nNote: I tried a targeted local snap-test invocation, but this checkout's snap-test runner rejected the guessed --update/--case options, so I kept validation to the CI-provided golden diff plus whitespace checks.

naokihaba:
Discussions regarding this matter are still ongoing, and there are several points that need to be finalized before we modify the code. Specifically, we need to decide whether the next step should be preparing the documentation, improving the CLI UX, or adjusting the migration behavior.

naokihaba:
@leno23 

Thank you for your work on this.

  After the discussion in #1854, the agreed direction is to document the current behavior in the migration guide
  rather than change the migration behavior here. `vp migrate` currently supports `lint-staged`, while other Git
  hook tools are intentionally skipped and should be configured manually.

  I’ll take it from here.

PR #1866, opened 2026-06-18
fix(global): ignore project node version for global pm

### LINKED ISSUE #1621: Do not resolve project-level Node.js version for global pm commands

wyf027:
## Summary

- Adds a managed-global-package Node.js resolver that skips project-level `.node-version` and `package.json` engine fields.
- Uses that resolver for `vp install -g` / `vp add -g`, global update node-mismatch checks, and global outdated registry lookups.
- Keeps explicit `--node`, `VP_NODE_VERSION`, session file, user default, and LTS fallback behavior intact.

Closes #1621

## Test Plan

- `git diff --check`
- `RUSTUP_TOOLCHAIN=stable rustfmt --edition 2024 --check crates/vite_global_cli/src/cli.rs crates/vite_global_cli/src/commands/env/config.rs crates/vite_global_cli/src/commands/global/install.rs crates/vite_global_cli/src/commands/global/mod.rs`

Attempted but blocked locally:

- `cargo fmt --check` and `cargo test -p vite-global-cli test_resolve_global_package_version -- --nocapture` were blocked because the repo nightly toolchain (`nightly-2026-06-10`) was partially installed and `rustup` stalled while downloading the required components.
- Running the focused cargo test with `RUSTUP_TOOLCHAIN=stable` reached dependency resolution but failed because this workspace needs nightly `-Z bindeps` for `fspy` from `vite-task`.


wyf027:
Pushed a targeted clippy fix for the unused current_dir import in crates/vite_global_cli/src/commands/global/install.rs. git diff --check passes locally; cargo fmt could not run because the local nightly rustfmt binary is missing its librustc_driver dylib.

wyf027:
Follow-up pushed: updated the failing global snap fixtures from the latest CI diffs. The pnpm project-node fixture now uses Node 22.13.0 so latest pnpm can still exercise the intended inheritance behavior, and the global-install nodeVersion snapshots match the current CI runtime. Local checks: `git diff --check`. Targeted snap-test could not run locally because this checkout is missing `packages/cli/dist`, and `pnpm -F vite-plus build-ts` is blocked by the local tsdown/rolldown install state (missing `rolldown/dist/index.mjs`).

wyf027:
Pushed another CI fix in `2b563c8`: pinned the devEngines precedence test to Node 22.13.0 so the musl job does not chase a Node release without a checksum entry, and updated the pnpm project-node-version snapshot to match the current CLI output. Local `git diff --check` passes; the targeted cargo test could not start locally because the configured Rust toolchain manifest is missing on this machine, so I am letting CI validate it.

wyf027:
Addressed the Codex corepack fallback finding in `4794213`: the auto-install path now passes the already resolved project Node version into the managed `corepack` install instead of resolving with `None`/global defaults.

Local validation:
- `git diff --check` passed

I attempted `cargo fmt --check -- crates/vite_global_cli/src/shim/corepack.rs`, but this checkout's configured nightly rustfmt cannot start because its `librustc_driver` dylib is missing, so CI will need to cover Rust formatting/tests.

liangmiQwQ:
I think this is an issue that still needs further discussion and confirmation.

fengmk2:

Issue #770, opened 2026-03-10
Should `vp dev` fail in monorepo root?


TheAlexLichter:
### Describe the bug

1. Create monorepo repo via `vp create`
2. go into root
3. run `vp dev`

Expected: Wouldn't work (would be `vp dev apps/website`)
Actual: Starts dev server but fails when opening the page.

### Reproduction

(explained above)

### System Info

```shell
VITE+ - The Unified Toolchain for the Web

Environment:
  Version       24.14.0
  Source        engines.node
  Source Path   /home/alichter/Programming/vp-demo/package.json
  Project Root  /home/alichter/Programming/vp-demo

Tool Paths:
  node  /home/alichter/.vite-plus/js_runtime/node/24.14.0/bin/node
  npm   /home/alichter/.vite-plus/js_runtime/node/24.14.0/bin/npm
  npx   /home/alichter/.vite-plus/js_runtime/node/24.14.0/bin/npx
VITE+ - The Unified Toolchain for the Web

vp v0.1.5

Local vite-plus:
  vite-plus  v0.1.5

Tools:
  vite             v8.0.0-beta.18
  rolldown         v1.0.0-rc.8
  vitest           v4.0.18
  oxfmt            v0.36.0
  oxlint           v1.52.0
  oxlint-tsgolint  v0.16.0
  tsdown           v0.21.1
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/voidzero-dev/vite-plus/issues) for the same bug.
- [x] Confirm this is a Vite+ issue and not an upstream issue (Vite, Vitest, tsdown, Rolldown, or Oxc).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example).

isfk:
> Same problem.
> 
> `vp run dev` and `vp dev apps/website` work fine, but `vp dev` is hitting a 404.

also in wsl

cpojer:
No, it should not fail, it should list all the projects where you can start a Vite dev server in, and run it there, interactively just like `vp run`.

mythmon:
[...]

PR #1870, opened 2026-06-18
fix(cli): warn for bare dev at monorepo root

wyf027:
## Summary
- warns when `vp dev` is run bare from a monorepo root
- suggests `vp run dev` or an explicit package directory like `vp dev apps/website`
- leaves the existing Vite dev behavior unchanged while reducing the confusing 404/no-app case

Refs #770.

## Verification
- `git diff --check`

## Blocked
- `pnpm -F vite-plus test src/__tests__/monorepo-root-hint.spec.ts` fails at Vitest startup because the local Vite dist file is missing: `vite/dist/vite/node/index.js`.

### DIFF
--- a/packages/cli/src/__tests__/monorepo-root-hint.spec.ts
+++ b/packages/cli/src/__tests__/monorepo-root-hint.spec.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  devCommandHasExplicitRoot,
+  formatDevMonorepoRootHint,
+  shouldWarnDevFromMonorepoRoot,
+} from '../monorepo-root-hint.js';
+import type { WorkspaceInfoOptional } from '../types/index.js';
+
+const workspaceInfo: WorkspaceInfoOptional = {
+  rootDir: '/repo',
+  packageManager: undefined,
+  packageManagerVersion: 'latest',
+  isMonorepo: true,
+  monorepoScope: '',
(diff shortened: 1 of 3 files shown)


wyf027:
Pushed a targeted TypeScript fix for packages/cli/src/__tests__/monorepo-root-hint.spec.ts by removing fixture fields that are not part of WorkspacePackage. Targeted oxfmt check and git diff --check pass locally.

wyf027:
Follow-up pushed: formatted the CLI entry import order reported by the failing `vp check` step. Local verification: `oxfmt --check packages/cli/src/bin.ts` and `git diff --check`.

wyf027:
Addressed the Codex dev preflight finding in `ef62411`: the monorepo-root dev hint now treats `detectWorkspace()` as best-effort and falls through to the Rust core command runner if workspace/package-manager detection fails. That keeps `vp dev` from being blocked by warning-only hint logic for package manager specs such as `pnpm@latest`.

Local validation:
- `git diff --check` passed

I intentionally did not keep the default `oxfmt` rewrite because, without this repo's formatting config, it changed unrelated file-wide quote style; CI should run the configured formatter/checks.

naokihaba:
I think this is an issue that still needs further discussion and confirmation.

Since it is still being discussed in #770, I believe it would be best to get a consensus from the stakeholders regarding your proposed solution.

fengmk2:

PR #1871, opened 2026-06-18
fix(setup): reject cli packages without provenance

### LINKED ISSUE #1826: Reject Vite+ install script if the `@voidzero-dev/vite-plus-cli-*` package does not have provenance
### Description

Reject Vite+ install script (`https://vite.plus`, `https://vite.plus/ps1`) if the `@voidzero-dev/vite-plus-cli-*` package does not have provenance.

Also reject `vp upgrade` for the same case.

These will make attacking viteplus more difficult.

### Suggested solution

Add checks here:
- https://github.com/voidzero-dev/vite-plus/blob/b48e0a0b6219a34664fb40c7f9dcbd0a6e150cd8/packages/cli/install.sh#L927-L934
- https://github.com/voidzero-dev/vite-plus/blob/b48e0a0b6219a34664fb40c7f9dcbd0a6e150cd8/packages/cli/install.ps1#L524-L531

And for update:
- https://github.com/voidzero-dev/vite-plus/blob/b48e0a0b6219a34664fb40c7f9dcbd0a6e150cd8/crates/vite_global_cli/src/commands/upgrade/mod.rs#L126

### Alternative

_No response_

### Additional context

_No response_

### Validations

wyf027:
## Summary

Fixes #1826.

- reject platform-specific `@voidzero-dev/vite-plus-cli-*` packages when npm metadata lacks `dist.attestations.provenance`
- apply the same check to `vp upgrade`, the POSIX install script, and the PowerShell install script
- keep local tarball and pkg.pr.new install paths unchanged

## Verification

- `git diff --check`
- `bash -n packages/cli/install.sh`
- `RUSTUP_TOOLCHAIN=stable rustfmt --edition 2024 --check crates/vite_setup/src/registry.rs`
- checked current npm metadata for `@voidzero-dev/vite-plus-cli-darwin-arm64@0.2.0` exposes `dist.attestations.provenance`

## Blocked locally

- `pwsh` is not installed in this environment, so I could not run a PowerShell parse check
- `cargo test -p vite_setup registry::tests::test_dist_info --lib` fails before compiling because the pinned `nightly-2026-06-10` toolchain has a missing manifest locally; retrying with `RUSTUP_TOOLCHAIN=stable` reaches workspace dependency resolution and fails on the nightly-only `fspy` dependency from `vite-task`

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -361,6 +361,7 @@
     "test": "vitest run"
   },
   "dependencies": {
+    "@napi-rs/cli": "catalog:",
     "@oxc-project/types": "catalog:",
     "@oxlint/plugins": "catalog:",
     "@vitest/browser": "catalog:",
@@ -379,7 +380,6 @@
     "vitest": "catalog:"
   },
   "devDependencies": {
-    "@napi-rs/cli": "catalog:",
     "@nkzw/safe-word-list": "catalog:",
     "@oxc-node/core": "catalog:",
     "@types/cross-spawn": "catalog:",
--- a/crates/vite_setup/src/registry.rs
+++ b/crates/vite_setup/src/registry.rs
@@ -20,6 +20,24 @@ pub struct PackageVersionMetadata {
 pub struct DistInfo {
     pub tarball: String,
     pub integrity: String,
+    #[serde(default)]
+    pub attestations: Option<DistAttestations>,
+}
+
+/// npm attestation metadata from `dist.attestations`.
+#[derive(Debug, Deserialize)]
+pub struct DistAttestations {
+    #[serde(default)]
+    pub provenance: Option<serde_json::Value>,
+}
+
+impl DistInfo {
+    fn has_provenance(&self) -> bool {
+        self.attestations
+            .as_ref()
+            .and_then(|attestations| attestations.provenance.as_ref())
+            .is_some()
+    }
 }
 
 /// Resolved version info with URLs and integrity for the platform package.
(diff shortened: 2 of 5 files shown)


wyf027:
Addressed the Codex review findings in `60980b2`:

- POSIX installer now checks provenance under `dist.attestations.provenance` instead of matching arbitrary top-level metadata fields
- PowerShell installer now parses raw string metadata responses before checking `dist.attestations.provenance`, and reports metadata errors consistently

Local validation:
- `bash -n packages/cli/install.sh`
- `git diff --check`
- metadata grep smoke check verifies valid `dist.attestations.provenance` passes and top-level-only attestations fail

`pwsh` is not installed in this macOS checkout, so the PowerShell change was validated by code inspection only and left for CI.

wyf027:
Pushed a follow-up shell installer fix in `f1f2b62`: the first grep-based path check was too narrow because real npm metadata can contain nested `dist.signatures` before `dist.attestations`, which caused the install.sh matrix to reject valid alpha metadata. The shell path now uses a small awk scanner to verify `dist.attestations.provenance` without matching top-level-only fake fields.

Local validation:
- `bash -n packages/cli/install.sh`
- `git diff --check`
- live alpha metadata for `@voidzero-dev/vite-plus-cli-darwin-arm64@alpha` passes
- top-level-only `attestations.provenance` fixture fails
- nested `dist.attestations.provenance` fixture passes

wyf027:
Updated this branch to address the macOS CLI E2E failure where `vp check` could not resolve `@napi-rs/cli` from the installed CLI package context. Moved the CLI package's `@napi-rs/cli` entry into runtime dependencies and refreshed the lockfile.\n\nLocal validation:\n- `pnpm install --lockfile-only`\n- `pnpm -F vite-plus exec tsgo --noEmit`\n- `git diff --check`

packages/cli/package.json
```diff
@@ -361,6 +361,7 @@
     "test": "vitest run"
   },
   "dependencies": {
+    "@napi-rs/cli": "catalog:",
```
fengmk2, line 364: Why?

fengmk2:
Please reply your proposal to https://github.com/voidzero-dev/vite-plus/issues/1826, and then submit a PR.

PR #1872, opened 2026-06-18
fix(install): block provenance downgrades for managed pms

### LINKED ISSUE #1806: Reject installing package managers without provenance that had provenance previously
### Description

Reject installing package managers that had provenance previously. It is similar to [pnpm's `trustPolicy: no-downgrade`](https://pnpm.io/settings#trustpolicy), but for the package managers installed by viteplus internally.

This makes it difficult to attack the package managers.

### Suggested solution

Add the provenance check here: https://github.com/voidzero-dev/vite-plus/blob/491cd3f488fe057b4627a4ea850f6c5c56d50883/crates/vite_install/src/request.rs#L236

- [npm](https://www.npmjs.com/package/npm?activeTab=versions): does not have provenance
- [yarn](https://www.npmjs.com/package/@yarnpkg/cli-dist?activeTab=versions): has provenance for `>=4.9.3`
- [pnpm](https://www.npmjs.com/package/pnpm?activeTab=versions): has provenance for `>= 10.20.0`
- [bun](https://github.com/oven-sh/bun/releases): does not have provenance

### Alternative

_No response_

### Additional context

_No response_

### Validations

wyf027:
## Summary

Fixes #1806.

- require npm provenance attestation for managed `pnpm` installs from `10.20.0` onward
- require npm provenance attestation for managed Yarn `@yarnpkg/cli-dist` installs from `4.9.3` onward
- preserve the existing behavior for npm, bun, pre-threshold pnpm/Yarn releases, and missing-version 404 mapping

## Verification

- `git diff --check`
- `RUSTUP_TOOLCHAIN=stable rustfmt --edition 2024 --check crates/vite_install/src/package_manager.rs`
- checked live npm metadata: `pnpm@10.20.0` and `@yarnpkg/cli-dist@4.9.3` include `dist.attestations.provenance`; `pnpm@10.19.0` does not

## Blocked locally

- `cargo test -p vite_install package_manager::tests::test_package_manager_provenance --lib` fails before compiling because the pinned `nightly-2026-06-10` toolchain has a missing manifest locally
- retrying with `RUSTUP_TOOLCHAIN=stable` reaches workspace dependency resolution and fails on the nightly-only `fspy` dependency from `vite-task`

### DIFF
--- a/crates/vite_install/src/package_manager.rs
+++ b/crates/vite_install/src/package_manager.rs
@@ -668,10 +668,94 @@ struct RegistryPackument {
     versions: HashMap<String, serde::de::IgnoredAny>,
 }
 
+/// npm package version metadata used before downloading managed package managers.
+#[derive(Deserialize)]
+struct RegistryPackageVersion {
+    #[serde(default)]
+    dist: RegistryDistInfo,
+}
+
+#[derive(Default, Deserialize)]
+struct RegistryDistInfo {
+    #[serde(default)]
+    attestations: Option<RegistryDistAttestations>,
+}
+
+#[derive(Deserialize)]
+struct RegistryDistAttestations {
+    #[serde(default)]
+    provenance: Option<serde_json::Value>,
+}
+
+impl RegistryPackageVersion {
+    fn has_provenance(&self) -> bool {
+        self.dist
+            .attestations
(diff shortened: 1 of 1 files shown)


fengmk2:

PR #1873, opened 2026-06-18
refactor(cli): share tool resolver env assembly

wyf027:
## Summary

- add a small `createToolResolution` helper for resolver return values
- reuse it from the lint, fmt, doc, and pack tool resolvers
- keep resolver bin paths and extra lint envs unchanged

## Verification

- `git diff --check`


wyf027:
Pushed the vp check formatting fix for packages/cli/src/resolve-lint.ts. Targeted oxfmt with the root fmt config and git diff --check pass locally.

fengmk2:

PR #1885, opened 2026-06-19
feat: support yarn v1 detection from lockfile

situ2001:
I found `vp i` is not working in my projects that use yarn v1 as its pkg manager. 

After digging into yarn v1 source code, I think we can distinguish from v1 or berry from lock file shape.

https://github.com/yarnpkg/yarn/blob/c2dda503f3759b5be5f0e24ecd9cf5c97a540147/src/lockfile/parse.js#L28

https://github.com/yarnpkg/yarn/blob/c2dda503f3759b5be5f0e24ecd9cf5c97a540147/src/lockfile/parse.js#L185-L194

### DIFF
--- a/crates/vite_install/src/package_manager.rs
+++ b/crates/vite_install/src/package_manager.rs
@@ -2,7 +2,7 @@ use std::{
     collections::HashMap,
     env, fmt,
     fs::{self, File},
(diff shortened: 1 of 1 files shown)


fengmk2:
@codex review

PR #1890, opened 2026-06-19
fix(check): preserve oxlint GitHub reporter

### LINKED ISSUE #925: Preserve oxlint GitHub reporter output in `vp check` on GitHub Actions
## Context

PR #914 fixed `vp check` failing on GitHub Actions by forcing `--format=default` for oxlint. The root cause was that when `GITHUB_ACTIONS=true`, oxlint auto-switches to the GitHub reporter, which omits the human-readable summary output that `vp check` parses to determine pass/fail status.

While the fix resolves the immediate breakage, it loses the advantage of oxlint's customized GitHub Actions reporter output (e.g., inline annotations on PR diffs).

## Problem

Currently `vp check` parses oxlint's human-readable summary output to produce unified pass/fail lines. This is incompatible with oxlint's GitHub reporter format, forcing us to disable it.

## Desired Behavior

When running on GitHub Actions, `vp check` should:
1. Still produce correct pass/fail output (not break as before)
2. Preserve oxlint's GitHub Actions reporter benefits (inline annotations, structured error reporting)

## Possible Approaches

- Parse the GitHub reporter output format in addition to the default format
- Run oxlint with the GitHub reporter and extract results from exit code / structured output instead of parsing stdout
- Use oxlint's `--format=github` explicitly and emit our own summary alongside it

## References

- Fix PR: #914
- Original issue: https://github.com/voidzero-dev/setup-vp/issues/7

wyf027:
## Summary

- stop forcing `--format=default` for the oxlint phase in `vp check`
- allow oxlint to use its GitHub Actions reporter again now that upstream includes the summary output needed by `vp check`
- add a parser regression test for GitHub annotation output before the oxlint summary

## Context

Closes #925.

The workaround from #914 was needed because oxlint's GitHub reporter did not emit the human-readable summary that `vp check` parses. The upstream fix landed in oxc-project/oxc#20404, and this repo currently pins oxlint 1.70.0, so the workaround should no longer be needed.

I noticed an older comment saying this would be handled later; I did not find an open PR for #925, so I kept this as the smallest revert-style change.

## Validation

- `gh pr view 20404 -R oxc-project/oxc --json state,mergedAt,mergeCommit`
- `gh pr list -R voidzero-dev/vite-plus --state open --search "925 OR oxlint GitHub reporter OR format=default OR github reporter"`
- `git diff --check -- packages/cli/binding/src/check/mod.rs packages/cli/binding/src/check/analysis.rs`
- `cargo +1.95.0 fmt --check -- packages/cli/binding/src/check/mod.rs packages/cli/binding/src/check/analysis.rs`

Not fully validated locally: `cargo test -p vite-plus-cli lint_output_allows_github_reporter_annotations_before_summary` could not run in this checkout. The pinned nightly toolchain fails before compilation with a missing rustup manifest / rustfmt dylib, and `cargo +1.95.0 test` fails while resolving the pinned `vite-task` git dependency `fspy`.

### DIFF
--- a/packages/cli/binding/src/check/analysis.rs
+++ b/packages/cli/binding/src/check/analysis.rs
@@ -247,7 +247,7 @@ pub(super) fn analyze_lint_output(output: &str) -> Option<Result<LintSuccess, Li
 mod tests {
     use serde_json::json;
 
-    use super::{LintMessageKind, analyze_lint_output, lint_config_type_check_enabled};
+    use super::{LintMessageKind, lint_config_type_check_enabled};
 
     #[test]
     fn lint_message_kind_defaults_to_lint_only_without_typecheck() {
@@ -307,22 +307,5 @@ mod tests {
         assert!(!lint_config_type_check_enabled(Some(&json!({
             "options": { "typeAware": true, "typeCheck": false }
         }))));
-    }
-
-    #[test]
-    fn lint_output_allows_github_reporter_annotations_before_summary() {
-        let output = "\
-::warning file=src/example.ts,line=1,col=1,title=oxlint::example annotation
-Found 0 warnings and 0 errors.
-Finished in 12ms on 1 file using 12 threads.
-";
-
-        let result = analyze_lint_output(output)
-            .expect("summary should be parsed")
-            .expect("zero warnings and errors should pass");
-
-        assert_eq!(result.summary.duration, "12ms");
-        assert_eq!(result.summary.files, 1);
-        assert_eq!(result.summary.threads, 12);
     }
 }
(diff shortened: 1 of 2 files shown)


fengmk2:
According to the reply at https://github.com/voidzero-dev/vite-plus/issues/925#issuecomment-4067032265, simply removing the `--format=default` parameter will solve the problem.

wyf027:
Updated this PR to match the maintainer guidance: it now simply stops forcing oxlint `--format=default` and leaves reporter selection to oxlint. I also removed the extra GitHub Actions format branch/tests from the prior revision.

Validation:

```bash
cargo fmt --check
git diff --check
cargo test -p vite-plus-cli check::
```

PR #1900, opened 2026-06-21
fix(upgrade): run pinned pnpm with managed Node

### LINKED ISSUE #1876: Force to use a compatible version for `vp upgrade`
### Describe the bug

Reference https://github.com/voidzero-dev/vite-plus/pull/1865#issuecomment-4740475248

`vp upgrade` now uses the node version resolved from session env var, project configure, which may be not compatible with Vite+'s requirement in `package.json#engines.node`. That will cause `pnpm` skip installing the optional binary files that makes Vite+ broken.

We should always use a compatible node version for `vp upgrade`, just like `delegate_to_global_cli`. Maybe we can reuse `ensure_cli_runtime` function but it also needs changing (Embed requirement at buildtime instead of reading at runtime, similar to #1412

### Reproduction

This can be reproduced by following the steps given in the vite-plus's source code repo.

### Steps to reproduce

1. Cd into the current `vite-plus` repo
2. Run `pnpm boostrap-cli && vp env use 20.0.0`
3. Run `vp upgrade`
4. Run vp commands like `vp pack`, `vp build`

liangmiQwQ:
Close #1876

The dependency install spawned by `vp upgrade` currently re-enters `vp install`, which can select a session, project, or system-first Node.js runtime. An incompatible runtime can make pnpm skip optional native binaries and leave the upgraded CLI broken.

This PR removes that recursive path. The setup flow resolves and downloads the latest managed Node.js LTS runtime, downloads the pinned pnpm version, and runs `node pnpm.cjs install` directly. This bypasses `VP_NODE_VERSION`, `.session-node-version`, project runtime configuration, and `vp env off` while keeping the install script and upgrade flow on the same managed bootstrap path.

🤖 Generated with Codex

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7976,9 +7976,7 @@ dependencies = [
  "tokio",
  "tracing",
  "vite_install",
- "vite_js_runtime",
  "vite_path",
- "vite_shared",
  "vite_str",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7976,6 +7976,7 @@ dependencies = [
  "tokio",
  "tracing",
  "vite_install",
+ "vite_js_runtime",
  "vite_path",
  "vite_str",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7976,7 +7976,9 @@ dependencies = [
  "tokio",
  "tracing",
  "vite_install",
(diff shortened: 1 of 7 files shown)


liangmiQwQ:
@codex

fengmk2 (approved):

Issue #1881, opened 2026-06-18
Indicate current installed node version when running `vp env list-remote`


jpbnetley:
### Description

When running `vp env list-remote`, it would be nice if the current installed versions can also be listed in a different color.

### Suggested solution

Similar to [nvm](https://github.com/nvm-sh/nvm), the current installed version(s) is in a different color.
In the screenshot, the left side is `vp` and the right side is `nvm`

<img width="1032" height="734" alt="Image" src="https://github.com/user-attachments/assets/99a315ad-f283-4145-a94d-1e2bab04c45c" />

### Alternative

could maybe indicate the current installed version with an asterisk `*` rather if colors needs to be avoided.

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/voidzero-dev/vite-plus/blob/main/CONTRIBUTING.md).
- [x] Confirm this request is for Vite+ itself and not for Vite, Vitest, tsdown, Rolldown, or Oxc.
- [x] Check that there isn't already an issue requesting the same feature.

semimikoh:
I'd like to work on this.

Quick question on the desired behavior before I open a PR: should we distinguish only "installed" versions, or also highlight the current/active one separately (like nvm's `->` and `default`)?

For the styling, I'm thinking: color the installed versions when stdout is a TTY, and fall back to an `*` marker (or `(installed)` suffix) when colors are disabled (piped/redirected output). Does that sound good?

jong-kyung:
> Quick question on the desired behavior before I open a PR: should we distinguish only "installed" versions, or also highlight the current/active one separately (like nvm's `->` and `default`)?

I think `vp env list-remote` should still primarily list registry versions, but it may be useful to annotate the version resolved for the current project/cwd using the same logic as `vp env current`.

The global default could be shown as a separate marker if present, since it can differ from the current project-resolved version.

@fengmk2 WDYT?

PR #1907, opened 2026-06-22
feat(env): mark installed/current/default versions in `vp env list-remote`

semimikoh:
### Description

Implements #1881: annotate `vp env list-remote` so installed/current/default Node.js versions are easy to spot in the registry listing.

This follows the direction discussed in the issue: `list-remote` still primarily lists registry versions, but now annotates the version resolved for the current project/cwd (same logic as `vp env current`) and the global default as a separate marker.

### What changed

- **Installed versions** are highlighted in green (blue for the current project-resolved version) when stdout supports color, and prefixed with `*` when colors are disabled (piped / `NO_COLOR`) so the distinction survives non-TTY output.
- **Current** (project-resolved, via `config::resolve_version`) and **default** (global `default_node_version`) versions are annotated with trailing `current` / `default` labels. They can differ, so each gets its own label.
- **`--json`** output gains `installed`, `current`, and `default` boolean fields.
- All local lookups degrade gracefully — if any fail, the registry listing still renders.

### Example

```
# TTY (color)
v24.2.0
v22.13.0 (Jod)   current     # blue + label (project-resolved)
v20.18.0 (Iron)  default     # green (installed) + label (global default)

# piped / NO_COLOR
  v24.2.0
* v22.13.0 (Jod) current
* v20.18.0 (Iron) default
```

### Testing

- `cargo test -p vite_global_cli list_remote` (added coverage for installed and current/default marking)
- `cargo clippy -p vite_global_cli` — clean
- `cargo fmt --check`


jong-kyung:
@codex review

jong-kyung:
Could we add a snap test?

semimikoh:
@jong-kyung Added a `command-env-list-remote` snap test. It also doubles as a regression guard for the alias fix — without it, `default marked` would be `false`.

jong-kyung:
@codex review

jong-kyung:
@codex review

semimikoh:
Good catch — marked the snap case `serial: true` since it mutates the shared VP_HOME config via `vp env default`, matching the other default-mutating snap tests.

jong-kyung (approved):
It looks good to me 👍

packages/cli/snap-tests-global/command-env-list-remote/snap.txt
```diff
@@ -0,0 +1,11 @@
+> vp env install lts # Install an LTS Node.js version locally
+Installing Node.js v<semver>...
+Installed Node.js v<semver>
+
+> vp env default lts # Set it as the global default (stored as the `lts` alias)
+✓ Default Node.js version set to lts (currently <semver>)
+
+> vp env list-remote --lts --json | node -e "const {versions}=JSON.parse(require('fs').readFileSync(0,'utf8')); console.log('installed marked:', versions.some(v=>v.installed)); console.log('current marked:', versions.some(v=>v.current)); console.log('default marked:', versions.some(v=>v.default));" # installed/current/default flags should all resolve, including the `lts` default alias
```
fengmk2, line 8: Just need to execute the command `vp env list-remote --lts --json`, and the subsequent judgment logic code can be removed. We only need to ensure that the snapshot content is consistent.

fengmk2 (approved):

PR #1919, opened 2026-06-23
fix(pack): bundle @tsdown/exe and @tsdown/css into core

### LINKED ISSUE #1586: `vp pack --exe` fails with "Failed to import module @tsdown/exe" when `@tsdown/exe` is installed as documented

fengmk2:
## Problem

`@tsdown/exe` and `@tsdown/css` hard-peer-depend on `tsdown` and import `tsdown/internal`, but Vite+ bundles tsdown internally with no resolvable top-level `tsdown` package. Installing them at the project level fails with `Failed to import module "@tsdown/exe"` (and the equivalent for CSS bundling).

## Fix

Bundle both extensions into core (`tsdown-exe.js`, `tsdown-css.js`) so `tsdown/internal` resolves at build time and `vp pack --exe` / CSS bundling work with no extra install. `lightningcss` (native, cannot be bundled) becomes an optional peer, loaded lazily with an actionable error when it is missing.

## Tests

- `command-pack-css`: CSS transforms run through the bundled `@tsdown/css` + lightningcss.
- `command-pack-tsdown-extensions`: the bundled exe/css chunks load without a top-level `tsdown`.

Closes #1586


docs/guide/pack.md
```diff
@@ -58,4 +58,12 @@ export default defineConfig({
 });
 ```
 
+Executable support is bundled into Vite+, so you do not need to install `@tsdown/exe` separately.
+
+Building executables uses Node's [Single Executable Applications](https://nodejs.org/api/single-executable-applications.html) support and requires Node.js 25.7.0 or later. Switch the active runtime with `vp env use 25` if `vp pack --exe` reports an unsupported version.
```
fengmk2: change to `vp env use 26` instead of `vp env use 25` for document will be better.

fengmk2: Done in 12560ea, changed to `vp env use 26`.

fengmk2:
@codex review

fengmk2:
cc @sxzz

cpojer (approved):

sxzz (approved):

Issue #1916, opened 2026-06-23
bug: vp i pins npm@11 into devEngines on low Node, breaking manual npm with "Cannot find module node:path"


l246804:
### Describe the bug

After switching Node to a low version (14.15.0) with `vp env use`, running `vp i` fails — but before failing it writes `devEngines.packageManager = npm@11.17.0` into the project's `package.json` (with `onFail: "download"`), downloads npm 11.17.0, and routes every subsequent `npm` invocation to it. Because npm 11 requires Node `^20.17.0 || >=22.9.0` and uses `require('node:path')` (the `node:` prefix is not supported by Node 14's CommonJS loader), **every `npm` command then crashes** — including manual `npm i` / `npm run dev`, not just `vp i`. This breaks the project's toolchain connectivity even though the user never asked for an npm upgrade.

```
$ npm -v
internal/modules/cjs/loader.js:883
  throw err;
  ^

Error: Cannot find module 'node:path'
Require stack:
- ~/.vite-plus/package_manager/npm/11.17.0/npm/lib/cli.js
- ~/.vite-plus/package_manager/npm/11.17.0/npm/bin/npm-cli.js
```

The same `Cannot find module 'node:path'` is thrown by `vp i`, `npm i`, and `npm run dev` — they all hit the vp-managed npm 11.17.0.

### Root cause

1. **No node↔npm compatibility check.** When `vp env use` pins Node to 14.15.0 (which ships npm 6.14.8), vp does not honor the bundled npm. Instead, on `vp i` it pins/writes npm@11.17.0 into `devEngines`, whose `engines` field is `^20.17.0 || >=22.9.0` — fundamentally incompatible with Node 14.
2. **Failed `vp i` still writes side-effect config.** `vp i` fails, yet leaves `devEngines.packageManager` behind in `package.json`, permanently routing `npm` to the incompatible npm 11.

Verified on disk:
- `package.json` now contains:
  ```json
  "devEngines": { "packageManager": { "name": "npm", "version": "11.17.0", "onFail": "download" } }
  ```
- npm 11's `cli.js` line 10: `const cliEntry = require('node:path').resolve(...)`
- npm 11's `package.json`: `"engines": { "node": "^20.17.0 || >=22.9.0" }`
- Node 14.15.0: `require('node:path')` → `MODULE_NOT_FOUND`

### Expected

[...]

PR #1921, opened 2026-06-23
feat: passthrough fallback for vpr/vp i on low Node (#1909 #1916)

### LINKED ISSUE #1909: bug: vite-plus@0.2.1 crashes on Node < 20.12 with "node:util has no export named styleText" before engine check
### Describe the bug

Running `vpr <script>` in a project pinned to a Node.js version that does not yet have `util.styleText` (added in Node 20.12 / 22) crashes at module load time, before any of vp's own engine-compatibility checks can run:

```
file:///~/.vite-plus/0.2.1/node_modules/.pnpm/vite-plus@0.2.1_vite@8.0.16/node_modules/vite-plus/dist/terminal-uTv0ZaMr.js:2
import { styleText } from "node:util";
         ^^^^^^^^^
SyntaxError: The requested module 'node:util' does not provide an export named 'styleText'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:98:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:143:5)
    at async Loader.import (internal/modules/esm/loader.js:165:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
```

Source:

```js
// vite-plus/dist/terminal-uTv0ZaMr.js
import { styleText } from "node:util";
```

`styleText` was added in Node 20.12 and only stabilized in v22, so any project whose `.node-version` resolves to Node 14 / 16 / 18.x < 20.12 will hit this `SyntaxError` during ESM instantiation of the terminal helper — i.e. before vp's own "Node.js X is incompatible with Vite+ CLI" message has a chance to fire, and before the script-runner soft-fallback work tracked in #1875 (PR #1865 / #1874) can take effect.

This makes `vpr` completely unusable in older repos even as a thin script forwarder, which is exactly the use case #1875 was supposed to preserve.

### Suggested fix

Either:

1. Avoid the static `import { styleText } from "node:util"` in any module that can be loaded before the engine check (in particular `terminal-*.js`, which the CLI bootstrap pulls in eagerly). A small `pc`-style fallback or a dynamic import gated on `process.versions.node >= 20.12` would let the runner reach the `vpr` soft-fallback path; or
2. Build the CLI bootstrap chunk with a downlevel target so static imports don't reference Node-22-only named exports.

### Reproduction

Any project with a `.node-version` pinning Node ≤ 20.11 (reproduced on 14.15.0). Steps:

1. In a directory with `.node-version` containing `14.15.0` (or use `vp env use 14.15.0`).
2. `vpr dev` (or any `vpr <script>` — fails before reaching the script).

Result: the `SyntaxError` above. Expected: the soft-fallback behavior from #1875, or at minimum the previous "Node.js X is incompatible with Vite+ CLI" diagnostic.

### Steps to reproduce

```
echo 14.15.0 > .node-version
vpr dev
```

### System Info

```
$ vp env current

Environment:
  Version       14.15.0
  Source        .node-version
  Source Path   /home/<me>/<project>/.node-version
  Project Root  /home/<me>/<project>

Tool Paths:
  node  ~/.vite-plus/js_runtime/node/14.15.0/bin/node
  npm   ~/.vite-plus/js_runtime/node/14.15.0/bin/npm
  npx   ~/.vite-plus/js_runtime/node/14.15.0/bin/npx

$ vp --version
vp v0.2.1

Local vite-plus:
  vite-plus  Not found

Environment:
  Package manager  npm latest
  Node.js          v14.15.0 (.node-version)
```

### Used Package Manager

npm

### Related

- #1875 — `vpr` should soft-fallback on incompatible Node versions. This issue is what blocks that fallback from ever being reached on 0.2.1, because the failure happens during ESM instantiation of `terminal-uTv0ZaMr.js`, prior to any version-guard logic.

### Validations

- [x] Checked existing issues; #1875 is related but distinct (engine guard vs. unconditional `node:util` named import).
- [x] This is a Vite+ issue (the offending import lives in the published `vite-plus` package).
- [x] Reproduction is minimal (any low-Node project + `vpr <script>`).

l246804:
## Summary

- Adds low-Node passthrough: when the project's Node < 20.19.0, `vpr`/`vp run` and the package-manager family bypass the Vite+ JS CLI and run the project's own package manager directly.
- Fixes #1909 (ESM `styleText` crash on Node < 20.12 — never loads Vite+ JS) and the low-Node case of #1916 (no longer pins/downloads an incompatible npm, no longer writes `devEngines`).
- npm without an explicit version uses the Node-bundled npm (compatible); pnpm/yarn/bun without an explicit version error with a "specify a version" message.
- Scope: passthrough only. The `devEngines` "fail leaves side effects" general fix (deferred pin) is split into a follow-up PR.

Closes #1909 (low-Node passthrough). Refs #1916.

## Changes

| Crate | Change |
|---|---|
| `vite_shared` | Add `MIN_SUPPORTED_NODE = "20.19.0"` constant + `is_node_below_min()` helper |
| `vite_install` | Add `PackageManagerBuilder::detect_only()` — builds PM without writing devEngines |
| `vite_pm_cli` | Extract `dispatch_with_pm()` + 12 `*_with_pm` handler variants for external PM injection |
| `vite_global_cli` | Add passthrough module (`is_eligible`, `should_passthrough`, `execute`) + precheck in `run_command_with_options` |

## How it works

1. In `run_command_with_options`, before the command match, check if the command is eligible (Run / PackageManager) and the project's resolved Node < 20.19.0.
2. If yes: print a one-line notice, resolve the PM via `detect_only` (no devEngines pin), and dispatch via `dispatch_with_pm` (PM commands) or `run_command` (Run commands).
3. If no: original path unchanged.

## Test plan

- [x] `cargo test -p vite_shared --lib` → 92 passed
- [x] `cargo test -p vite_install --lib` → 552 passed
- [x] `cargo test -p vite_pm_cli --lib` → 12 passed
- [x] `cargo test -p vite_global_cli --bin vp passthrough` → 6 passed
- [x] `cargo clippy` on all 4 crates → clean
- [ ] Manual: `.node-version=14.15.0` + `vpr`/`vp i` passthrough, no `devEngines` written
- [ ] `pnpm bootstrap-cli && pnpm test && git status` (requires cmake)


crates/vite_global_cli/src/commands/passthrough.rs
```diff
@@ -0,0 +1,212 @@
...
+        command,
+        Commands::Run { .. } | Commands::PackageManager(_)
+    )
+}
+
+/// Print the one-line passthrough notice. Pure I/O, no version check.
+#[allow(dead_code)] // wired up in Task 4
+fn print_passthrough_notice(node_version: &str, min: &str) {
+    // Reuse shared output style; keep to a single concise line.
+    vite_shared::output::info(&format!(
```
liangmiQwQ:
```suggestion
    vite_shared::output::warn(&format!(
```
Is it better to show a warning?

liangmiQwQ (commented):
Thanks!

[plans/2026-06-23-low-node-passthrough.md‎](https://github.com/voidzero-dev/vite-plus/pull/1921/changes#diff-8cf82637daa261ef7013fab324b45ba541ef0dd53d244e3af27ee73354ece435) and [specs/2026-06-23-low-node-passthrough-design.md‎](https://github.com/voidzero-dev/vite-plus/pull/1921/changes#diff-93575101f8c4658584ae071384a274297098d87552d920069a8641b96e5f4538) seem left by mistake?

Could you please also clean up outdated comments and dead code like `Wired in Task X`?

liangmiQwQ:
I tested locally, and I found it only works for `vp run` command but not for `vpr` command. We'd better handle passthrough fallback for shims as well.

@codex

liangmiQwQ:
@codex review

crates/vite_global_cli/src/commands/passthrough.rs
l246804: Sorry, I haven't tested this PR locally yet; it was mistakenly submitted by AI yesterday.

liangmiQwQ:
Sorry, I have to close it as it is not planned anymore.

PR #1925, opened 2026-06-23
fix(cli): inline lint-staged version into staged debug bundle

### LINKED ISSUE #1922: `vp staged --debug` crashes by reading `vite-plus/dist/package.json` instead of lint-staged's package.json
### Describe the bug

`vp staged` succeeds, but adding `--debug` causes Vite+ to crash while trying to read `vite-plus/dist/package.json`.

Expected:
- `vp staged --debug` should succeed and print the bundled `lint-staged` version correctly

Actual:
- `vp staged --debug` throws `ENOENT` for `vite-plus/dist/package.json`

Example with `--debug`:

```txt
$ vp staged --debug
node:internal/fs/promises:640
  return new FileHandle(await PromisePrototypeThen(
                        ^

Error: ENOENT: no such file or directory, open '/Users/rokuosan/.local/share/mise/installs/viteplus/0.2.1/lib/node_modules/vite-plus/dist/package.json'
    at async open (node:internal/fs/promises:640:25)
    at async Object.readFile (node:internal/fs/promises:1290:14)
    at async getVersion (file:///Users/rokuosan/.local/share/mise/installs/viteplus/0.2.1/lib/node_modules/vite-plus/dist/staged/bin.js:6490:20)
    at async lintStaged (file:///Users/rokuosan/.local/share/mise/installs/viteplus/0.2.1/lib/node_modules/vite-plus/dist/staged/bin.js:6542:59)
    at async file:///Users/rokuosan/.local/share/mise/installs/viteplus/0.2.1/lib/node_modules/vite-plus/dist/staged/bin.js:6747:18 {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/rokuosan/.local/share/mise/installs/viteplus/0.2.1/lib/node_modules/vite-plus/dist/package.json'
}

Node.js v24.17.0
```

Example without `--debug`:

```txt
$ vp staged
✔ Backed up original state in git stash (825fd1e)
✔ Hiding unstaged changes to partially staged files...
✔ Running tasks for staged files...
✔ Applying modifications from tasks...
✔ Restoring unstaged changes...
✔ Cleaning up temporary files...
```

This appears to be a Vite+ packaging/bundling issue rather than a `lint-staged` issue. The bundled staged entry contains `lint-staged@16.4.0`, but `getVersion()` resolves `../package.json` relative to the bundled file, so it ends up reading `vite-plus/dist/package.json`.

### Local-only workaround

As a local workaround, adding `dist/package.json` avoids the crash:

```diff
diff --git a/dist/package.json b/dist/package.json
new file mode 100644
index 0000000..653bf6d
--- /dev/null
+++ b/dist/package.json
@@ -0,0 +1,3 @@
+{
+  "version": "16.4.0"
+}
```

### Reproduction

https://github.com/rokuosan/repro-vp-staged-debug

### Steps to reproduce

1. Create a project with `vite-plus@0.2.1`
2. Configure `staged` in `vite.config.ts`
3. Stage at least one file
4. Run:

```sh
$ vp staged --debug
```

### System Info

`vp env current` is mentioned in the issue template, but on `vite-plus@0.2.1` it is not available in this environment:

```shell
```

`vp --version` output:

```
$ vp --version

vp v0.2.1

Local vite-plus:
  vite-plus  v0.2.1

Tools:
  vite             v8.0.16
  rolldown         v1.1.1
  vitest           v4.1.9
  oxfmt            v0.55.0
  oxlint           v1.70.0
  oxlint-tsgolint  v0.23.0
  tsdown           v0.22.3


`sw_vers` output:


$ sw_vers
ProductName:            macOS
ProductVersion:         26.3
BuildVersion:           25D125
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

rokuosan:
## Problem

- #1922

`vp staged --debug` crashes in the published package.

When debug logging is enabled, bundled `lint-staged` calls `getVersion()`(`lib/version.js`), which reads its own manifest via`readFile(new URL('../package.json', import.meta.url))`. The CLI bundleslint-staged into `dist/staged/bin.js`, so that relative `../package.json` does not exist at runtime and the command throws instead of printing the debug log.

This is exactly what the pinned `lint-staged@16.4.0` does in `lib/version.js`:

```js
import fs from 'node:fs/promises'

export const getVersion = async () => {
  const packageJson = JSON.parse(await fs.readFile(new URL('../package.json', import.meta.url)))
  return packageJson.version
}
```

Source: https://github.com/lint-staged/lint-staged/blob/445f9dd042b88528c798b2e25c21c9adbc69a732/lib/version.js#L1-L6

## Fix

Add a tsdown plugin (`inline-lint-staged-version`) that rewrites the
`./version.js` import coming from `lint-staged/lib/index.js` to a virtual module
exposing the dependency version resolved at build time:

```js
export const getVersion = async () => "16.4.0";
```

Keeps the `getVersion` named-export shape identical, so no lint-staged internals change.

## Tests

- Extend the `command-staged-with-config` global snap test with a `vp staged --debug` step that must succeed. A regression flips the recorded exit code (`>` → `[1]>`), failing the snapshot.

Verified against a real build: `dist/staged/bin.js` now contains the inlined `const version = "16.4.0"` and no longer references `readFile(new URL('../package.json', ...))`.

```bash
vscode ➜ /workspaces/vite-plus (fix-debug-staged) $ vp staged --debug
lint-staged: Running `lint-staged@16.4.0` on Node.js v22.18.0 (linux) +126ms
lint-staged:execGit: Running git command: [ 'version', '--build-options' ] +128ms
lint-staged: git version 2.51.1
cpu: aarch64
no commit associated with this build
```

---

closes: #1922


jong-kyung (approved):
Thanks, that matches the direction I had in mind.

One small alternative: how about replacing `lint-staged/lib/version.js` directly? 

It could make the change a bit shorter, since the debug path calls `getVersion()` here:
https://github.com/lint-staged/lint-staged/blob/5f3b8f28e895972bd5a2cdba733327b49859b91f/lib/index.js#L106-L115

and `getVersion()` is isolated in `lib/version.js`:
https://github.com/lint-staged/lint-staged/blob/5f3b8f28e895972bd5a2cdba733327b49859b91f/lib/version.js#L3-L6

Not blocking though — the current approach also looks clear to me.

jong-kyung:
Also, could you please look into the CI failure?

rokuosan:
Thanks, that makes sense.
Replacing /lint-staged/lib/version.js directly is simpler here, so I'll update it.

I'll also investigate the CI failure.

jong-kyung (approved):

packages/cli/snap-tests-global/command-staged-with-config/steps.json
```diff
@@ -9,6 +9,12 @@
     },
     "vp staged # should succeed with staged .ts files",
     { "command": "git add -A && git commit -m 'second'", "ignoreOutput": true },
+    {
+      "command": "echo 'export const bar = 2;' >> src/index.ts && git add src/index.ts",
+      "ignoreOutput": true
+    },
+    "vp staged --debug > /dev/null 2>&1 # should succeed with debug enabled",
```
jong-kyung: Should we keep the command output visible/captured for the snapshot here?

rokuosan: The intent here is only to verify `vp staged --debug` doesn't crash.

I considered snapshotting the output too.
One thing I considered asserting was the inlined `lint-staged` version, but that would likely end up normalized to `<semver>` anyway, and beyond that the output mostly comes from `lint-staged --debug`.

Since the regression was the ENOENT crash while reading `dist/package.json`, I thought checking successful completion was enough.

Happy to capture the output if you think it adds useful signal.

jong-kyung: How about using `ignoreOutput: true` instead of redirecting to `/dev/null` here?
It would still keep the snapshot stable on success, but preserve the command output if it fails, which should make regressions easier to debug.

rokuosan: That makes sense. 
I originally followed an existing example that redirected stdout/stderr to `/dev/null`, but `ignoreOutput: true` seems like a better fit here.

Thanks for the suggestion!

Updated: [97cf339](https://github.com/voidzero-dev/vite-plus/pull/1925/commits/97cf33985300dcd91e5019eb82f3d9d4c042c1a0)

jong-kyung (approved):
Thanks! 👍

fengmk2 (approved):

fengmk2:
@rokuosan @jong-kyung Thanks!

PR #1927, opened 2026-06-23
Validate npm update lockfiles

### LINKED ISSUE #1926: vp update can produce a package-lock that vp install rejects
### Summary

`vp update` can produce a `package-lock.json` that a subsequent `vp install` rejects with npm `ERESOLVE`.

The concrete case is an npm project with Storybook and Vite+:

- starting valid graph: `storybook@10.3.5`, `@storybook/react-vite@10.3.5`, `vite-plus@0.2.0`
- important starting metadata: `storybook@10.3.5` has no `vite-plus` peer dependency; its peer dependencies are only optional `prettier`
- `vp update` succeeds and floats the lockfile to `storybook@10.4.6`, `@storybook/react-vite@10.4.6`, `vite-plus@0.2.1`
- `vp install` then fails because `storybook@10.4.6` has `peerOptional vite-plus@^0.1.15`, which rejects `vite-plus@0.2.1`

This is not just a `vite-plus@0.2.1` patch-level issue: semver `^0.1.15` accepts `0.1.x` only and does not include any `0.2.x` release. So pinning Vite+ from `0.2.1` back to `0.2.0` would still leave the Storybook `10.4.6` peer graph invalid.

I would expect `vp update && vp install` to preserve a valid installable project when the starting lockfile was valid.

### Environment

Observed with:

- `vp v0.2.1`
- local `vite-plus v0.2.1`
- npm `11.17.0`
- Node `24.17.0`
- package manager: npm / `package-lock.json`

### Minimal repro

```sh
mkdir /tmp/vite-plus-repro-storybook
cd /tmp/vite-plus-repro-storybook

npm init -y
npm install --save-dev --save-prefix='^' --ignore-scripts \
  storybook@10.3.5 \
  @storybook/react-vite@10.3.5 \
  vite-plus@0.2.0 \
  react@18.3.1 \
  react-dom@18.3.1

# Valid starting graph.
node -e "const lock=require('./package-lock.json'); for (const k of ['node_modules/storybook','node_modules/@storybook/react-vite','node_modules/vite-plus']) console.log(k, lock.packages[k]?.version)"

# Keystone detail: storybook@10.3.5 has no vite-plus peer.
node -e "const p=require('./node_modules/storybook/package.json'); console.log(p.version, p.peerDependencies, p.peerDependenciesMeta)"

vp install --ignore-scripts

# Succeeds, but writes a lockfile that install cannot reproduce.
vp update

node -e "const lock=require('./package-lock.json'); for (const k of ['node_modules/storybook','node_modules/@storybook/react-vite','node_modules/vite-plus']) console.log(k, lock.packages[k]?.version, lock.packages[k]?.peerDependencies || {})"

# Fails.
vp install --ignore-scripts
```

The starting Storybook metadata is:

```text
10.3.5 { prettier: '^2 || ^3' } { prettier: { optional: true } }
```

There is no `vite-plus` entry in either `peerDependencies` or `peerDependenciesMeta` for `storybook@10.3.5`.

### Actual result

The first `vp install --ignore-scripts` succeeds.

`vp update` succeeds and resolves:

```text
node_modules/storybook 10.4.6 {
  '@types/react': '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0',
  prettier: '^2 || ^3',
  'vite-plus': '^0.1.15'
}
node_modules/@storybook/react-vite 10.4.6 {
  react: '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0',
  'react-dom': '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0',
  storybook: '^10.4.6',
  vite: '^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0'
}
node_modules/vite-plus 0.2.1 {
  '@vitest/browser-playwright': '4.1.9',
  '@vitest/browser-webdriverio': '4.1.9'
}
```

The subsequent `vp install --ignore-scripts` fails:

```text
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: storybook@10.4.6
npm error Found: vite-plus@0.2.1
npm error node_modules/vite-plus
npm error   dev vite-plus@"^0.2.0" from the root project
npm error
npm error Could not resolve dependency:
npm error peerOptional vite-plus@"^0.1.15" from storybook@10.4.6
```

### Expected result

`vp update` should not successfully leave the project in a state where `vp install` immediately fails.

Reasonable outcomes would be:

- `vp update` avoids selecting the Storybook update that creates the invalid peer graph
- `vp update` performs a post-update install/peer-resolution validation and exits non-zero
- ideally, `vp update` rolls back the lockfile/package changes when post-update validation fails

### Source investigation

The current implementation appears to delegate update and install through different npm paths without validating the updated graph afterward.

At `053fc9337884716b8ac27df63b19dc2ea4c6295e`:

- `crates/vite_pm_cli/src/handlers.rs`: `run_update` builds a package manager and calls `pm.run_update_command(...)`.
- `crates/vite_install/src/commands/update.rs`: `run_update_command` resolves and executes one package-manager command.
- In the npm branch of `resolve_update_command`, Vite+ maps `vp update` directly to `npm update`.
- `crates/vite_install/src/commands/install.rs`: `vp install` maps npm to `npm install` / `npm ci`, which catches this peer conflict.

So the underlying issue seems to be that npm's `update` command can write a graph that npm's `install` command later refuses. Vite+ exposes both as a unified workflow, so `vp update` probably needs a post-update validation step, at least for npm.

One possible minimal validation for npm after a successful update:

```sh
npm install --package-lock-only --ignore-scripts
```

That command fails on the repro above and catches the invalid peer graph without needing to run lifecycle scripts.

dale-lakes:
## Summary

- validate npm update results with `npm install --package-lock-only --ignore-scripts`
- restore `package.json`, `package-lock.json`, and `npm-shrinkwrap.json` when npm accepts an update that npm install later rejects
- leave pnpm, yarn, and bun update behavior unchanged

## Root Cause

`vp update` delegates npm projects to `npm update`, while `vp install` delegates to `npm install`. npm can update the lockfile into a graph that `npm install` later rejects for peer dependency resolution, so `vp update && vp install` can fail after `vp update` reports success.

## Repro Used

A minimal npm project with:

- `storybook@^10.3.5`
- `@storybook/react-vite@^10.3.5`
- `vite-plus@^0.2.0`

The starting lockfile resolves Storybook `10.3.5` and Vite+ `0.2.0`. That starting graph is valid because `storybook@10.3.5` has no `vite-plus` peer dependency; its peer dependencies are only optional `prettier`.

Current `npm update` floats the lockfile to Storybook `10.4.6` and Vite+ `0.2.1`, but Storybook `10.4.6` adds optional peer `vite-plus@^0.1.15`, so the subsequent install rejects the lockfile with `ERESOLVE`.

This is structural to Storybook's declared peer range, not just a patch float problem: semver `^0.1.15` accepts `0.1.x` only and does not include `0.2.0` or `0.2.1`.

With this patch, the same repro makes `vp update` exit `1`, restores the lockfile to Storybook `10.3.5` / Vite+ `0.2.0`, and `npm install --package-lock-only --ignore-scripts` succeeds afterward.

Fixes #1926

## Validation

- `cargo fmt --all --check`
- `RUST_MIN_STACK=8388608 cargo test -p vite_install`
- `cargo clippy -p vite_install --all-targets --all-features -- --deny warnings -A clippy::byte_char_slices -A clippy::manual_assert_eq -A clippy::needless_return_with_question_mark -A clippy::unused_async_trait_impl -A clippy::useless_borrows_in_formatting`
- local Storybook repro against `target/debug/vp update`

### DIFF
--- a/crates/vite_install/src/commands/update.rs
+++ b/crates/vite_install/src/commands/update.rs
@@ -1,12 +1,18 @@
-use std::{collections::HashMap, process::ExitStatus};
+use std::{
+    collections::HashMap,
+    fs, io,
+    path::{Path, PathBuf},
+    process::ExitStatus,
+};
 
 use vite_command::run_command;
 use vite_error::Error;
 use vite_path::AbsolutePath;
 use vite_shared::output;
 
-use crate::package_manager::{
-    PackageManager, PackageManagerType, ResolveCommandResult, format_path_env,
+use crate::{
+    commands::install::InstallCommandOptions,
+    package_manager::{PackageManager, PackageManagerType, ResolveCommandResult, format_path_env},
 };
 
 /// Options for the update command.
@@ -35,9 +41,49 @@ impl PackageManager {
         options: &UpdateCommandOptions<'_>,
         cwd: impl AsRef<AbsolutePath>,
     ) -> Result<ExitStatus, Error> {
+        let cwd = cwd.as_ref();
+        let snapshot = if self.requires_update_validation() {
+            Some(ProjectFileSnapshot::capture(cwd, NPM_UPDATE_PROJECT_FILES)?)
+        } else {
+            None
+        };
         let resolve_command = self.resolve_update_command(options);
-        run_command(&resolve_command.bin_path, &resolve_command.args, &resolve_command.envs, cwd)
-            .await
+        let status = run_command(
+            &resolve_command.bin_path,
+            &resolve_command.args,
+            &resolve_command.envs,
+            cwd,
+        )
+        .await?;
+        if !status.success() {
+            if let Some(snapshot) = snapshot {
+                snapshot.restore()?;
+            }
+            return Ok(status);
+        }
+
+        if let Some(validate_command) = self.resolve_update_validation_command() {
+            let status = run_command(
+                &validate_command.bin_path,
+                &validate_command.args,
+                &validate_command.envs,
+                cwd,
+            )
+            .await?;
+
+            if !status.success() {
+                if let Some(snapshot) = snapshot {
+                    snapshot.restore()?;
+                    output::warn(
+                        "npm update produced package metadata that npm install could not resolve. Restored package.json and lockfile state.",
+                    );
+                }
+            }
+
+            return Ok(status);
+        }
+
+        Ok(status)
     }
 
     /// Resolve the update command.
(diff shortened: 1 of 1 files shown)


dale-lakes:
Closing this PR after narrowing the root cause. `vp update` and `vp install` are wrapping npm here; the raw npm commands reproduce the same invalid-lockfile behavior, so this should be fixed/tracked in npm rather than patched in Vite+.

PR #1931, opened 2026-06-24
fix(pm): allow cache flags without separator

jong-kyung:
## Summary

 Improve `vp pm cache` pass-through argument parsing so cache commands can accept package-manager flags without
requiring a `--` separator, such as `vp pm cache clean --force`.


jong-kyung:
@codex review

jong-kyung:
@codex review

fengmk2:
Why do we need to change this? The current requirement to use `--` to pass parameters not managed by `vite+` is intentional. Similar designs can be seen in other subcommands of `vp pm`. This is designed to avoid adding parameters later. If `--` were removed, any differentiated parameters would confuse users.

jong-kyung:
> Why do we need to change this? The current requirement to use `--` to pass parameters not managed by `vite+` is intentional. Similar designs can be seen in other subcommands of `vp pm`. This is designed to avoid adding parameters later. If `--` were removed, any differentiated parameters would confuse users.

Thanks, that makes sense. I opened this because `docs/guide/install.md` currently shows `vp pm cache clean --force`, so I assumed that was the intended UX.

If requiring `--` for package-manager-specific flags is intentional, I agree this PR should not change the parser behavior. I will update the docs to use `vp pm cache clean -- --force` instead.

jong-kyung:
I’ll close this PR and update `docs/guide/install.md` instead. 🙇‍♂️

PR #1934, opened 2026-06-24
fix: correct error type mismatch

liangmiQwQ:
In codebase, `ErrorConfig` seems overused a lot. There are a lot of errors should not be wrapped and should be other error types.

It may be created by mistake, and be treated as a project standard by contributors and agents and copied over and over again.

This PR corrects them.


liangmiQwQ:
@codex

fengmk2 (approved):
@liangmiQwQ Good job!

PR #1939, opened 2026-06-24
fix(cli): skip lazyPlugins only while resolving config metadata

fengmk2:
## Problem

`lazyPlugins` decided whether to run the user's plugin factory from `VP_COMMAND`. That signal is:

- **overloaded** — `vp run`'s task-discovery config read and a verbatim build child both see `run`;
- **unnormalized** — the `vp format` alias is `format`, not `fmt`;
- **inherited** by spawned children.

So a Vite build spawned under `vp run` (a verbatim `node build.mjs` task) or `vp exec` built **without** the user's plugins, while `vp format` wrongly **loaded** them. (Both were raised in review.)

## Fix

Decide from the resolution intent instead of the command name. The metadata resolvers (`resolveViteConfig`, migrate's rolldown-compat check, and the oxlint/oxfmt subprocess) mark that the config is being loaded only to read a config block, and `lazyPlugins` skips the factory **only while that marker is set**.

The default is to load, so `dev`/`build`/`test`/`preview` and any spawned build keep their plugins, regardless of which command launched them. The marker is scoped to the resolution, so it never leaks to task children, and it's immune to aliases and not-yet-known commands.

`VP_COMMAND` had no other reader, so it's removed.

Adds snap tests for the two fixed cases (verbatim `vp run` build loads plugins; `vp format` skips them) plus updated unit tests and docs.

### DIFF
--- a/docs/guide/troubleshooting.md
+++ b/docs/guide/troubleshooting.md
@@ -69,7 +69,7 @@ export default defineConfig({
 
 When `vite.config.ts` imports plugins at the top level, they are evaluated for every command, including `vp lint`, `vp fmt`, editor integrations, and long-lived background processes. This can make config loading slow and may trigger plugin setup side effects, such as reading files, starting watchers, or connecting to services.
 
-Use `lazyPlugins` to load plugins only when the Vite pipeline actually runs (`dev`, `build`, `test`, `preview`):
+Use `lazyPlugins` to skip the plugin factory when vite-plus loads your config only to read a metadata block (`lint`, `fmt`, `check`, `staged`, `pack`, `create`, the `run`/`cache` task lookup, and editor tooling). The plugins still load whenever Vite actually runs, `dev`, `build`, `test`, `preview`, and any build your own scripts spawn (a `vp run` task, `vp exec`):
 
 ```ts [vite.config.ts]
 import { defineConfig, lazyPlugins } from 'vite-plus';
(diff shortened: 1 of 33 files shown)


fengmk2:
@codex review

fengmk2:
@codex review

fengmk2:
Thanks for the review. Status on the four comments:

- **Verbatim `vp run` build loses plugins** and **`vp format` alias loads plugins** — both fixed by the approach change in this PR: instead of a `VP_COMMAND` command-name check, `lazyPlugins` now skips only while a *config-metadata resolution* is in flight (`withConfigMetadataResolution`). A verbatim `node build.mjs` under `vp run` runs after the marker is cleared, so it loads plugins; `vp format` skips via the oxfmt subprocess marker (alias-agnostic). Added snap tests `vite-plugins-run-verbatim-build` and `vite-plugins-skip-on-format` covering exactly these.

- **Migration dynamic import uses `.ts`** — fixed: `checkRolldownCompatibility` now imports `../define-config.js` (matching the neighboring `../index.js`), so it resolves the bundled entry instead of throwing into the silent `catch`.

- **Marker not preserved across overlapping resolutions** — fixed: `withConfigMetadataResolution` is now ref-counted (depth counter + saved baseline), so the marker stays set until the last overlapping/nested resolution finishes. Added a test that overlaps two resolutions and asserts `lazyPlugins` still skips while one is in flight.

fengmk2:
@codex review

fengmk2:
@codex review

cpojer (approved):

sapphi-red (approved):

PR #1943, opened 2026-06-25
fix(global): clean stale global package installs on upgrade

liangmiQwQ:
The current global package install flow can leave inactive `packages/<name>#<uuid>` directories behind when an install is interrupted before metadata activation. Older legacy installs can also leave inactive `packages/<name>` directories without an install ID.

This PR makes `vp upgrade` run a best-effort cleanup for stale global package install directories. It preserves every active install directory recorded in metadata, including legacy installs where `installId` is empty, removes inactive identified and legacy install directories once they are old enough, and leaves malformed `#...` names alone.

Recent inactive install directories are kept so a concurrent `vp install -g` that has created its package prefix but not saved metadata yet cannot be deleted by upgrade cleanup.

🤖 Generated with Codex.

### DIFF
--- a/crates/vite_global_cli/src/commands/global/install.rs
+++ b/crates/vite_global_cli/src/commands/global/install.rs
@@ -745,10 +745,11 @@ async fn cleanup_stale_installation_dir(
     path: &std::path::Path,
     active_install_dirs: &HashSet<std::path::PathBuf>,
 ) {
-    let has_install_id = name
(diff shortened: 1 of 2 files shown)


liangmiQwQ:
@codex

liangmiQwQ:
I have a better solution, we can maintain a `File::lock()` while running global installation, and all packages will be removed except the new install one and the locked one.

PR #1944, opened 2026-06-25
feat(docker): official Vite+ toolchain image

### LINKED ISSUE #1490: Feature request: official Docker image that honors .node-version
### Feature request

Please consider publishing an official Vite+ Docker image that can keep the container's Node.js version aligned with the project's pinned Node version.

### Problem

When building a Vite+ project into a Docker image, I need the runtime Node.js version to match the project's `.node-version` file.

In this project, `.node-version` contains `24.15.0`, and `vp --version` correctly reports:

```text
Environment:
  Package manager  pnpm v10.33.2
  Node.js          v24.15.0 (.node-version)
```

While optimizing the image, I can use plain `alpine:3.23` with `apk add nodejs` to get a much smaller runtime image. However, Alpine currently installs Node `v24.14.1`, which does not match the project requirement. The safer workaround is to use an official Node image such as `node:24-alpine`, but that roughly doubles the image size in this case.

### Desired behavior

It would be helpful if Vite+ provided an official Docker image or documented Docker image pattern that honors the project's `.node-version` file.

For example, a Vite+ image could either:

- read `.node-version` and install/use that exact Node version during the build/runtime flow, or
- provide image tags that map to exact Node versions managed by Vite+, or
- document the recommended Docker pattern for keeping container Node versions aligned with `.node-version`.

### Why this matters

Without an official Vite+ Docker image or recommended pattern, projects that need an exact Node version have to choose between:

- a larger official Node base image that matches the version more closely, or
- a smaller distro package such as Alpine's `nodejs`, which may lag behind the project's `.node-version` patch version.

### Workaround

Use an official Node image such as `node:24-alpine` when exact patch-version alignment matters, or accept the distro-packaged Node version when image size matters more.

### Validations

- [x] I searched existing issues for Docker/image/node-version related reports.
- [x] I checked the closest existing Node version management issue (#1200), which appears to be about slow Node installation/runtime management rather than Docker image version alignment.

fengmk2:
Implements the official Vite+ Docker image from the RFC (`rfcs/docker-image.md`).

`vp` already provisions the exact Node.js from `.node-version`, so one toolchain image builds any project (no Node-version-keyed tags). It is not a production runtime image: a documented multi-stage build copies the resolved Node.js into a small, vp-free runtime stage.

**Changes**
- `docker/Dockerfile`: glibc (`debian:bookworm-slim`) image bundling `vp` + native build toolchain, non-root user.
- `release.yml`: `publish-docker` job, multi-arch (amd64/arm64) push to `ghcr.io/voidzero-dev/vite-plus` after npm publish, version-tagged.
- `docs/guide/docker.md` + sidebar: multi-stage runtime, static SPA, CI, devcontainer, ad-hoc usage.

**Verified locally**: `.node-version=24.15.0` resolves and installs exactly v24.15.0; the copied Node runs standalone in a plain `debian:bookworm-slim` stage.

**Note**: ~~the first publish needs the GHCR package made public / linked to the repo in org settings.~~

Closes #1490


fengmk2:
## Manual verification commands for the Docker preview images

Preview images on GHCR (from the `pkg.pr.new` build):

- `ghcr.io/voidzero-dev/vite-plus:pr-1944` (Debian)
- `ghcr.io/voidzero-dev/vite-plus:pr-1944-alpine` (Alpine/musl)

### 1. Basic functionality

```bash
docker run --rm ghcr.io/voidzero-dev/vite-plus:pr-1944 vp --version
docker run --rm ghcr.io/voidzero-dev/vite-plus:pr-1944 sh -c \
  'id -un; command -v vp; vp help >/dev/null && echo "vp help ok"; git --version'
```

### 2. Common `vp` commands (install / build / fmt / lint / test / check)

```bash
docker run --rm ghcr.io/voidzero-dev/vite-plus:pr-1944 bash -s <<'EOF'
set -e
cd /app
echo "24.15.0" > .node-version
printf 'node_modules/\ndist/\n' > .gitignore
cat > package.json <<'JSON'
{ "name": "vp-verify", "private": true, "type": "module", "devDependencies": { "vite-plus": "0.2.1" } }
JSON
mkdir -p src
cat > vite.config.ts <<'TS'
import { defineConfig } from 'vite-plus'
export default defineConfig({
  build: { ssr: 'src/server.ts', outDir: 'dist', rollupOptions: { output: { entryFileNames: 'server.js' } } },
})
TS
cat > src/greeting.ts <<'TS'
export function greeting(name: string): string {
  return `hello, ${name}`;
}
TS
cat > src/server.ts <<'TS'
import { createServer } from 'node:http';
import { greeting } from './greeting';
createServer((_req, res) => res.end(greeting('world'))).listen(3000);
TS
cat > src/greeting.test.ts <<'TS'
import { expect, test } from 'vitest';
import { greeting } from './greeting';
test('greeting', () => { expect(greeting('vp')).toBe('hello, vp'); });
TS
vp install --no-frozen-lockfile
node --version          # -> v24.15.0 (from .node-version)
vp build
vp fmt
vp lint
vp test
vp check
EOF
```

Swap the tag to `ghcr.io/voidzero-dev/vite-plus:pr-1944-alpine` to verify the Alpine image (same results, musl Node).

### 3. End-to-end multi-stage deploy (build app, run on a slim runtime)

```bash
mkdir -p vp-docker-e2e/src && cd vp-docker-e2e
echo "24.15.0" > .node-version
printf 'node_modules/\ndist/\n' > .gitignore
cat > package.json <<'JSON'
{ "name": "e2e", "private": true, "type": "module", "devDependencies": { "vite-plus": "0.2.1" } }
JSON
cat > vite.config.ts <<'TS'
import { defineConfig } from 'vite-plus'
export default defineConfig({
  build: { ssr: 'src/server.ts', outDir: 'dist', rollupOptions: { output: { entryFileNames: 'server.js' } } },
})
TS
cat > src/server.ts <<'TS'
import { createServer } from 'node:http'
createServer((_req, res) => { res.writeHead(200); res.end('vite-plus docker OK\n') })
  .listen(Number(process.env.PORT) || 3000)
TS
cat > Dockerfile <<'EOF'
# syntax=docker/dockerfile:1
FROM ghcr.io/voidzero-dev/vite-plus:pr-1944 AS build
WORKDIR /app
COPY --chown=vp:vp package.json .node-version ./
RUN vp install --no-frozen-lockfile
COPY --chown=vp:vp . .
RUN vp build
RUN cp "$(vp env which node | head -1)" /tmp/node

FROM ghcr.io/voidzero-dev/vite-plus:pr-1944 AS deps
WORKDIR /app
COPY --chown=vp:vp package.json .node-version ./
RUN vp install --no-frozen-lockfile --prod

FROM debian:bookworm-slim AS runtime
WORKDIR /app
ENV NODE_ENV=production
COPY --from=build /tmp/node /usr/local/bin/node
COPY --from=build /app/dist ./dist
COPY --from=deps /app/node_modules ./node_modules
COPY --from=build /app/package.json ./
USER nobody
EXPOSE 3000
CMD ["node", "dist/server.js"]
EOF

docker build -t vp-e2e .
cid=$(docker run -d -p 3000:3000 vp-e2e)
sleep 2 && curl -s localhost:3000          # -> vite-plus docker OK
docker exec "$cid" /usr/local/bin/node --version   # -> v24.15.0
docker rm -f "$cid"
```

Note `COPY --chown=vp:vp` in the build/deps stages: the image runs as the non-root `vp` user, so without it `vp install` hits `Permission denied`. For an Alpine end-to-end build, use the `-alpine` tag and an `alpine` runtime stage (a musl Node only runs on a musl base); see `docs/guide/docker.md`.

.github/workflows/publish-to-pkg.pr.new.yml
```diff
@@ -137,3 +137,129 @@ jobs:
...
+    steps:
+      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
+        env:
+          IMAGE: ghcr.io/voidzero-dev/vite-plus
+        with:
+          script: |
+            const image = process.env.IMAGE;
+            const pr = context.payload.pull_request.number;
+            const marker = '<!-- docker-preview -->';
+            const body = [
```
fengmk2, line 231: Could you show the final size of the Docker image? Please display this metadata in a table.

fengmk2: Done in 901e9726e. The `comment-docker-preview` job now pulls each published preview image, measures it (`docker image inspect`), and renders a size table (Image | Size) above the pull commands. Re-triggering the workflow to refresh the sticky comment with the table.

fengmk2:
An exception was found that may be related to the vite task. The lint command runs normally when executed directly, but throws an error after passing through the task runner.

```bash
docker run --rm ghcr.io/voidzero-dev/vite-plus:pr-1944 bash -c \
  'vp create vite:monorepo --directory hello --no-interactive && cd hello && vp lint && vp run ready'

◇ Scaffolded hello with Vite+ monorepo
• Node 24.18.0  pnpm 11.9.0
✓ Dependencies installed in 27s
→ Next: cd hello && vp run
Found 0 warnings and 0 errors.
Finished in 1.2s on 6 files with 111 rules using 16 threads.
$ vp check
pass: All 18 files are correctly formatted (1112ms, 16 threads)
Error running tsgolint: "exit status: exit status: 1"/app/hello/node_modules/.pnpm/oxlint-tsgolint@0.23.0/node_modules/oxlint-tsgolint/bin/tsgolint.js:18
    throw e;
    ^

<ref *1> Error: spawnSync /app/hello/node_modules/.pnpm/@oxlint-tsgolint+linux-x64@0.23.0/node_modules/@oxlint-tsgolint/linux-x64/tsgolint EINVAL
    at Object.spawnSync (node:internal/child_process:1143:20)
    at spawnSync (node:child_process:911:24)
    at Object.execFileSync (node:child_process:954:15)
    at Object.<anonymous> (/app/hello/node_modules/.pnpm/oxlint-tsgolint@0.23.0/node_modules/oxlint-tsgolint/bin/tsgolint.js:11:17)
    at Module._compile (node:internal/modules/cjs/loader:1871:14)
    at Object..js (node:internal/modules/cjs/loader:2002:10)
    at Module.load (node:internal/modules/cjs/loader:1594:32)
    at Module._load (node:internal/modules/cjs/loader:1396:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5) {
  errno: -22,
  code: 'EINVAL',
  syscall: 'spawnSync /app/hello/node_modules/.pnpm/@oxlint-tsgolint+linux-x64@0.23.0/node_modules/@oxlint-tsgolint/linux-x64/tsgolint',
  path: '/app/hello/node_modules/.pnpm/@oxlint-tsgolint+linux-x64@0.23.0/node_modules/@oxlint-tsgolint/linux-x64/tsgolint',
  spawnargs: [ 'headless' ],
  error: [Circular *1],
  status: null,
  signal: null,
  output: null,
  pid: 0,
  stdout: undefined,
  stderr: undefined
}

Node.js v24.18.0

Linting failed before analysis started
error: Linting could not start
```

cc @wan9chi

fengmk2:
@codex review

docs/guide/docker.md
```diff
@@ -0,0 +1,228 @@
...
+
+```dockerfile [Dockerfile]
+# syntax=docker/dockerfile:1
+
+# --- build stage: the official Vite+ toolchain image ---
+FROM ghcr.io/voidzero-dev/vite-plus:latest AS build
+WORKDIR /app
+
+# Install dependencies first so this layer is cached across source changes.
+COPY --chown=vp:vp package.json pnpm-lock.yaml .node-version ./
```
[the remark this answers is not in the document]
fengmk2: Fixed in 3a61144d7. The dependency layer now copies only `package.json` + the lockfile (always present), so projects that pin Node via `engines.node`/`devEngines.runtime` (no `.node-version` file) no longer fail at this `COPY`. `.node-version`, when used, is picked up from the full `COPY . .` before `vp build`. Applied across the SSR, deps, static, and Alpine examples (and the RFC).

docker/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+# downloads its own pinned Node at build time, so the default is dead weight in a
+# builder image. The node/npm/npx shims remain and fetch the right version on
+# first use.
+#
+# Keep this install line in sync with docker/Dockerfile.alpine.
+RUN curl -fsSL https://vite.plus | VP_VERSION="${VP_VERSION}" VP_PR_VERSION="${VP_PR_VERSION}" bash \
+ && vp --version \
+ && rm -rf "$VP_HOME/js_runtime"
+
+WORKDIR /app
```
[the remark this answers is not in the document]
fengmk2: This doesn't reproduce for this image: `/app` is created by `WORKDIR /app` while `USER vp` is active, and under BuildKit (forced by the `# syntax=docker/dockerfile:1` directive) WORKDIR creates missing dirs owned by the current user, so the baked `/app` is vp-owned, not root-owned.

Verified on the published preview image:
```
$ docker run --rm ghcr.io/voidzero-dev/vite-plus:pr-1944 sh -c 'ls -ld /app; id -un; touch /app/probe && echo WRITABLE'
drwxr-xr-x 1 vp vp 0 ... /app
vp
WRITABLE
```
A downstream multi-stage build (`FROM …:pr-1944` → `WORKDIR /app` → `COPY --chown=vp:vp …` → `vp install`) creates `node_modules` under `/app` and runs end to end (HTTP 200). The 'WORKDIR creates dirs as root even after USER' behavior is the legacy (non-BuildKit) builder, which doesn't apply here. So no change needed; happy to add an explicit `mkdir -p /app && chown vp:vp /app` if you'd prefer it spelled out.

fengmk2:
@codex review

docs/guide/docker.md
```diff
@@ -0,0 +1,230 @@
...
+```dockerfile [Dockerfile]
+# syntax=docker/dockerfile:1
+
+# --- build stage: the official Vite+ toolchain image ---
+FROM ghcr.io/voidzero-dev/vite-plus:latest AS build
+WORKDIR /app
+
+# Install dependencies first so this layer is cached across source changes.
+COPY --chown=vp:vp package.json pnpm-lock.yaml ./
+RUN vp install --frozen-lockfile
```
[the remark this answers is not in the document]
fengmk2: Addressed by the `.node-version*` glob in the dependency `COPY` (commit a67c9d9db). It is optional (build succeeds when the file is absent, so engines.node/devEngines projects don't fail) and, when present, copies `.node-version` into every stage before `vp install`, including the separate `deps` stage, so the install resolves the pinned Node.

.github/workflows/release.yml
```diff
@@ -232,3 +232,70 @@ jobs:
...
             • Windows: `irm https://vite.plus/ps1 | iex`
           embed-url: https://github.com/${{ github.repository }}/releases/tag/v${{ env.VERSION }}
+
+  # Build and push the official toolchain Docker image to GHCR after the npm
+  # release is published (the image installs vp from npm, so the version must
+  # exist first). See docker/Dockerfile and docs/guide/docker.md.
+  publish-docker:
+    name: Publish Docker image (${{ matrix.variant }})
+    runs-on: ubuntu-latest
+    needs: [check, Release]
```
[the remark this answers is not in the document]
fengmk2: Fixed in 36989b2c2: the Discord notification step is now `continue-on-error: true`, so a webhook failure no longer fails the Release job or skips `publish-docker` after npm/GitHub release are published. (publish-docker still needs the Release job for the npm publish; needs is job-level, so making the notification non-blocking is the fix.)

docker/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+#   docker build --build-arg VP_PR_VERSION=1569 .
+ARG VP_PR_VERSION=
+
+# Toolchain image: include git and a C/C++ build toolchain so native addons
+# (for example better-sqlite3 or sharp) can compile during `vp install`.
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends \
+      ca-certificates \
+      curl \
+      git \
```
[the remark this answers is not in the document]
fengmk2: Fixed in 85b56924b: added `openssh-client` alongside `git` in both the Debian (apt) and Alpine (apk) toolchain layers, so `git+ssh` dependencies resolve during `vp install`. Verified `ssh` is on PATH (`/usr/bin/ssh`) in both images.

docker/Dockerfile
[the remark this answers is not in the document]
fengmk2: Made explicit in 85b56924b: the toolchain layer now `mkdir -p /app && chown vp:vp /app` before `USER vp`. (`/app` was already vp-owned because BuildKit creates the WORKDIR as the current user, verified writable end to end, but making it explicit removes any builder dependency.) Verified `/app` is `vp:vp` in both images.

fengmk2:
@codex review

docs/guide/docker.md
```diff
@@ -0,0 +1,231 @@
...
+The image is a toolchain image, not a production runtime image. Because `vp`
+already reads your project's pinned Node.js version (`.node-version`,
+`engines.node`, or `devEngines.runtime`) and downloads that exact version, you do
+not need a Node-version-specific base image: one image builds any project against
+its own Node.
+
+For production, you do not ship this image. Instead you use a multi-stage build
+where this image builds the app, and the exact Node.js it resolved is copied into
+a small, vp-free runtime image. That keeps the deployed image small while
+matching your project's Node version exactly.
```
fengmk2: should be `Node.js version exactly`.

We should use `node` or `Node.js`, not `Node`.

fengmk2: Agreed, and fixed in 7fb11897b. `Node.js` is the trademarked product name (OpenJS Foundation; the [trademark policy](https://nodejs.org/trademark-policy.pdf) says the mark must not be abbreviated or shown in parts), so bare `Node` is incorrect. Convention now applied: `Node.js` for the runtime/version in prose, lowercase `node` for the binary/command and image tags (`node:24`). Swept the whole Docker guide, the RFC, and the Dockerfile comments (line 19 is now "Node.js version exactly").

fengmk2:
@Brooooooklyn I think is good to go for this first version.

Brooooooklyn (approved):

docs/guide/docker.md
```diff
@@ -0,0 +1,231 @@
...
+| `ghcr.io/voidzero-dev/vite-plus:<major>.<minor>.<patch>` | Exact version  |
+
+The examples use `:latest` to track the newest release; pin an exact tag or a
+digest if you need reproducible builds. The image is published for `linux/amd64`
+and `linux/arm64` and runs as a non-root user by default.
+
+Browse all published versions and digests on the [GitHub package page](https://github.com/voidzero-dev/vite-plus/pkgs/container/vite-plus).
+
+The default image is Debian (glibc). An Alpine (musl) variant is published under
+the same versions with an `-alpine` suffix (`:latest-alpine`,
```
fengmk2: Do not provide Alpine version for now.

TheAlexLichter: Why not?

fengmk2: According to @Brooooooklyn's experience, maintenance costs will be very high.

TheAlexLichter: Lets start w/o it then and see if there is demand.

fengmk2: Done in 3ebdc238, dropped the Alpine variant for now. Removed `docker/Dockerfile.alpine`, collapsed the release + pkg.pr.new preview Docker jobs back to a single Debian build (no matrix), removed the Alpine section and `-alpine` tag references from this guide, and moved Alpine back to Future Work in the RFC. The image is glibc-only for the first version.

docs/guide/docker.md
```diff
@@ -0,0 +1,176 @@
+# Docker
+
+Vite+ publishes an official Docker image that bundles the `vp` CLI for the
```
fengmk2: The intro repeats the same idea a few times. I think we can make it shorter: state once that this is a build/CI/devcontainer image, once that `vp` resolves Node from the project so we don't need Node-version-specific image tags, and then let the multi-stage examples carry the details.

Suggested replacement for the intro:

````md
Vite+ publishes an official Docker image with the `vp` CLI preinstalled:

```bash
ghcr.io/voidzero-dev/vite-plus
```

Use it for builds, CI, and devcontainers. It is not intended as a production runtime image.

`vp` resolves the Node.js version from your project (`.node-version`, `devEngines.runtime`, or `engines.node`) and downloads that exact version during install/build. This means the image does not need Node-version-specific tags.

For production, use a multi-stage build: build the app with the Vite+ image, then copy only the resolved Node.js binary, build output, and production dependencies into a smaller runtime image.
````

**Review by fengmk2 — COMMENTED** [review]
Left one inline doc suggestion for the Docker guide intro.

File: .github/workflows/publish-to-pkg.pr.new.yml [hunk 10]
```diff
@@ -137,3 +137,137 @@ jobs:
...
+              '## 🐳 Docker preview image',
+              '',
+              "Built from this PR's pkg.pr.new build:",
+              '',
+              '| Image | Compressed size |',
+              '| --- | --- |',
+              `| \`${image}:pr-${pr}\` | ${process.env.COMPRESSED_SIZE} |`,
+              '',
+              '```bash',
+              `docker pull ${image}:pr-${pr}`,
```
fengmk2: Deed to delete the local image cache first, then pull it again to avoid using the cache from the previous run when conducting continuous testing.

wan-kong:
Looks like the `latest` tag for `ghcr.io/voidzero-dev/vite-plus` has not been published yet.

Will it be published as part of the next Vite+ release?

fengmk2:
@wan-kong Yes, the new version will be released today.

PR #1945, opened 2026-06-25
fix(cli): clean stale global package installs

liangmiQwQ:
The current global install cleanup only removes the install directory recorded by previous metadata. If a reinstall is interrupted, unrecorded package directories can stay behind, and the legacy empty install-id package directory can also survive.

This PR keeps a file lock for each in-progress global package install directory, then replaces the previous-install cleanup with a same-package stale directory sweep. The sweep skips locked directories so concurrent global installs are protected, and removes both `package#installId` stale directories and the empty install-id layout.

Related:
- https://github.com/voidzero-dev/vite-plus/pull/1906#issuecomment-4794286796
- https://github.com/voidzero-dev/vite-plus/pull/1943#issuecomment-4796576465

🤖 Generated with Codex.


liangmiQwQ:
@codex

liangmiQwQ:
Adding hooks in `vp upgrade` was not chosen at last, cleaning while upgrading sounds odd and this approach might be better.

fengmk2 (approved):

PR #1949, opened 2026-06-25
fix(install): clean up the temp dir when a package-manager install fails

shulaoda:
#### Problem

Both `download_package_manager` (pnpm/npm/yarn) and `download_bun_package_manager` created their temporary install directory like this:

```rust
let target_dir_tmp = tempfile::tempdir_in(parent_dir)?.path().to_path_buf();
```

`tempfile::tempdir_in` returns a `TempDir` guard whose `Drop` deletes the directory. Here the guard is an **unnamed temporary** that drops at the end of the `let`, so the directory is deleted immediately and only the `PathBuf` survives.

It still worked on the happy path because `download_and_extract_tgz_with_hash` recreates the directory (`create_dir_all`) before writing. But on any failure after that — a download/extract error, or the `package` → bin rename — the function returns early via `?`, leaving the recreated temp dir behind. With the guard already gone, nothing cleans it up, so a `.tmpXXXX` directory leaks into `$VP_HOME/package_manager/<name>/`. Every failed install (flaky network, corrupt download, …) leaks one, and they accumulate forever.

#### Fix

Bind the `TempDir` to a named `tmp_dir` that lives for the whole function:

```rust
let tmp_dir = tempfile::tempdir_in(parent_dir)?;
let target_dir_tmp = tmp_dir.path().to_path_buf();
```

Now any early return drops the guard and removes the temp dir. On success the dir is renamed into its final location first, so the end-of-scope drop hits a `NotFound` on the original path, which `tempfile` ignores — success is unaffected. Applied to both sites.

#### Test

`test_download_package_manager_cleans_temp_dir_on_failure` drives `download_package_manager(Pnpm, "10.0.0", None)` against a mock registry serving a non-gzip tarball (the download succeeds, extraction fails), then asserts no leftover directory remains under `package_manager/pnpm/`. It fails on the old code (a leaked `.tmp*` dir) and passes on the fix. The full `package_manager` module stays green (91 passed).

#### Reproduce

```bash
cargo test -p vite_install --lib \
  package_manager::tests::test_download_package_manager_cleans_temp_dir_on_failure
# If you run a local/system HTTP proxy, prefix with NO_PROXY=127.0.0.1,localhost
# so the loopback mock isn't intercepted (same requirement as the other httpmock
# tests in this crate).
```

To watch the bug itself: temporarily revert the two sites to `let target_dir_tmp = tempfile::tempdir_in(parent_dir)?.path().to_path_buf();` and rerun — the test fails with a leaked `.tmpXXXX` directory under `package_manager/pnpm/`.

### DIFF
--- a/crates/vite_install/src/package_manager.rs
+++ b/crates/vite_install/src/package_manager.rs
@@ -871,7 +871,9 @@ pub async fn download_package_manager(
     // Use tempfile::TempDir for robust temporary directory creation
     let parent_dir = target_dir.parent().unwrap();
     tokio::fs::create_dir_all(parent_dir).await?;
-    let target_dir_tmp = tempfile::tempdir_in(parent_dir)?.path().to_path_buf();
+    // Keep the TempDir guard alive so a failure path cleans up the temp dir.
+    let tmp_dir = tempfile::tempdir_in(parent_dir)?;
+    let target_dir_tmp = tmp_dir.path().to_path_buf();
 
     download_and_extract_tgz_with_hash(&tgz_url, &target_dir_tmp, expected_hash).await.map_err(
         |err| {
@@ -987,7 +989,9 @@ async fn download_bun_package_manager(
 
     // Download the platform-specific package directly
     let platform_tgz_url = get_npm_package_tgz_url(platform_package_name, version);
-    let target_dir_tmp = tempfile::tempdir_in(parent_dir)?.path().to_path_buf();
+    // Keep the TempDir guard alive so a failure path cleans up the temp dir.
+    let tmp_dir = tempfile::tempdir_in(parent_dir)?;
+    let target_dir_tmp = tmp_dir.path().to_path_buf();
 
     download_and_extract_tgz_with_hash(&platform_tgz_url, &target_dir_tmp, None).await.map_err(
         |err| {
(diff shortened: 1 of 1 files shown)


crates/vite_install/src/package_manager.rs
```diff
@@ -987,7 +989,9 @@ async fn download_bun_package_manager(
 
     // Download the platform-specific package directly
     let platform_tgz_url = get_npm_package_tgz_url(platform_package_name, version);
-    let target_dir_tmp = tempfile::tempdir_in(parent_dir)?.path().to_path_buf();
+    // Keep the TempDir guard alive so a failure path cleans up the temp dir.
+    let tmp_dir = tempfile::tempdir_in(parent_dir)?;
```
fengmk2, line 993: Are there any clippy rules that can be configured to ensure that subsequent code doesn't repeat the same mistakes?

fengmk2 (approved):
@shulaoda Thanks!

crates/vite_install/src/package_manager.rs
fengmk2: Good question, I dug into this. Short answer: there's no clippy rule that reliably catches this specific footgun, and the one config that could would over-fire on a lot of correct code.

What I checked, by temporarily reverting one site back to the buggy `tempdir_in(parent_dir)?.path().to_path_buf()` form and running clippy:

- Default workspace clippy (`all` + `pedantic` + `nursery`): nothing.
- `significant_drop_tightening` / `significant_drop_in_scrutinee`: nothing.
- `let_underscore_must_use` / `let_underscore_drop`: don't apply, the value is bound to a named variable, not `let _`.

Why nothing fires: the chain *uses* the `TempDir` (so its `#[must_use]` is satisfied) and `.to_path_buf()` copies the path out before the guard drops, so there's no borrow-after-drop either. It's a valid program that just drops the guard earlier than intended.

The only `disallowed-methods` candidate would be `tempfile::TempDir::path`, but that's the same method we call legitimately on a kept guard (every `dir.path()` in the tests), so disallowing it would fire on ~all the correct usages too. Scoping it to "only when chained off `tempdir`" isn't expressible in `.clippy.toml`.

So the structural fix (binding the guard) plus the regression test is the real guard here. If you'd like, I can additionally disallow the deliberate-leak methods `TempDir::into_path` / `TempDir::keep` (both currently unused in the repo) to block the related "persist a temp dir forever" footgun, but that wouldn't have caught this particular drop bug.

PR #1961, opened 2026-06-27
fix(cli): skip local packages during global update

liangmiQwQ:
The current global update flow can treat packages installed from a local path or tarball like normal registry packages. That means `vp update -g some-package` may try to update a package that was originally installed from `./pkg`.

This PR records whether a managed global package was installed from local filesystem content. Global outdated/update checks now skip packages with that metadata, including named updates like `vp update -g some-package`, which follows npm's behavior for local global installs.

Legacy metadata stays readable and defaults to non-local.

🤖 Generated with Codex

### DIFF
--- a/crates/vite_global_cli/src/cli.rs
+++ b/crates/vite_global_cli/src/cli.rs
@@ -727,7 +727,7 @@ async fn managed_update(
         }
         current_node_version = get_current_node_version().await?;
 
-        for metadata in &all {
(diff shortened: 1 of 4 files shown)


liangmiQwQ:
This PR has some blockers (Vite+'s global alias bug, global packages version handling bug), and to follow npm's behavior there are tons of things need to deal with, `npm` itself's behavior is also strange, close it for now.

PR #1962, opened 2026-06-27
fix(windows): improve install.ps1 errors for missing VC++ runtime

cheezone:
## Summary

On clean Windows machines without the MSVC runtime, `vp.exe` fails to start during `irm https://vite.plus/ps1 | iex` with exit code `0xC0000135` (STATUS_DLL_NOT_FOUND). The installer previously reported a generic "Failed to install dependencies" with an empty `install.log`, and `exit 1` closed the entire PowerShell window when run via `iex`.

This PR:
- Detects `0xC0000135` after `vp.exe install` fails and prints VC++ 2015–2022 Redistributable install guidance (x64/arm64)
- Replaces bare `exit 1` with a non-CI halt path so interactive `irm | iex` installs keep the shell open without duplicating the error message

All `Write-Error-Exit` paths now keep the interactive shell open when not running under `CI=true`.

## Context

Similar report elsewhere: https://github.com/kirodotdev/Kiro/issues/9752

## Test plan

- [x] Clean Windows without VC++: single clear error, shell stays open (no duplicate message, no window close)
- [ ] Windows with VC++ installed: install completes normally
- [ ] CI path (`CI=true`): still exits with non-zero code

### DIFF
--- a/packages/cli/install.ps1
+++ b/packages/cli/install.ps1
@@ -90,12 +90,33 @@ function Test-IsInstallStopException {
     return $ErrorRecord.Exception.Message -eq $script:InstallStopSignal
 }
 
+function Test-ShouldKeepShellOpenAfterFailure {
+    # Only `irm ... | iex` typed in an already-open interactive shell should keep the
+    # session alive. CI, script files, and `powershell -Command "..."` must exit non-zero.
+    if ($env:CI -eq "true") {
+        return $false
+    }
+    if ($PSCommandPath) {
+        return $false
+    }
+    if (-not [Environment]::UserInteractive) {
+        return $false
+    }
+    try {
+        $commandLine = (Get-CimInstance Win32_Process -Filter "ProcessId=$PID").CommandLine
+        if ($commandLine -match '(^|\s)-Command(\s|$)') {
+            return $false
+        }
+    } catch {
+        return $false
+    }
+    return $true
+}
+
 function Exit-Installer {
     param([int]$Code = 1)
     $global:LASTEXITCODE = $Code
-    # CI and script-file invocations must propagate a non-zero process exit code.
-    # `irm ... | iex` has no $PSCommandPath; use a catchable stop signal to keep the shell open.
-    if ($env:CI -eq "true" -or $PSCommandPath) {
+    if (-not (Test-ShouldKeepShellOpenAfterFailure)) {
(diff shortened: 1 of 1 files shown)


fengmk2:
@codex review

ljlsasuke:
I’ve come across this problem too; it’s a great fix.

cheezone:
Thanks for the Codex review — addressed in the latest push:

1. **Non-CI exit code**: `Exit-Installer` now calls `exit $Code` when `$PSCommandPath` is set (e.g. `powershell -File install.ps1`), not only when `CI=true`. The catchable stop signal is only used for `irm | iex` where there is no script file path.

2. **NTSTATUS comparison**: `Test-IsDllNotFoundExitCode` no longer casts arbitrary negative exit codes to `[uint32]`; it compares the signed value and uses `-band 0xFFFFFFFF` for hex matching.

3. **Nested catch**: `Get-PackageMetadata` inner `catch` blocks rethrow `VP_INSTALL_STOP` so registry/JSON fallback does not mask an already-printed error.

packages/cli/install.ps1
```diff
@@ -749,4 +820,10 @@ exec "`$VP_HOME/current/bin/vp.exe" "`$@"
     Write-Host ""
 }
 
-Main
+try {
+    Main
+} catch {
+    if (-not (Test-IsInstallStopException $_)) {
+        throw
```
fengmk2: Blocking: this also swallows failures for non-CI `powershell -Command "irm ... | iex"` callers, so the host process can exit 0 after a failed install. Please only swallow this sentinel for truly interactive `irm | iex`; non-interactive callers should still exit non-zero.

fengmk2 (commented):
Left one blocking inline comment.

cheezone:
@fengmk2 Good catch — updated in fba7366c.

`Test-ShouldKeepShellOpenAfterFailure` now only swallows the stop signal when all of these hold:
- not `CI=true`
- no `$PSCommandPath` (not `-File`)
- `[Environment]::UserInteractive`
- current process command line does **not** include `-Command`

So `powershell -Command "irm ... | iex"` exits non-zero again, while typing `irm ... | iex` in an open interactive shell still keeps the window open.

fengmk2 (approved):
LGTM

PR #1963, opened 2026-06-27
fix(global): handle bins with uppercase name

liangmiQwQ:
There are two problems:

- Vite+ now transforms all bin names to lower case. When there is a bin including uppercase letter (like `vitePlus`), it can stop the dispatch part from getting the real package and the real bin, and report bin not found.
- Different from macOS / Windows, Linux is case-sensitive, means users can install bins like `vP` or `nPM`. Force to transform them into lowercase break these packages.

This PR fixes them by reading the real bin filename on Windows and macOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8365,6 +8365,7 @@ dependencies = [
  "node-semver",
  "owo-colors",
  "oxc_resolver",
+ "same-file",
  "serde",
  "serde_json",
  "serial_test",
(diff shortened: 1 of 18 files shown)


liangmiQwQ:
@codex

liangmiQwQ:
@codex review

liangmiQwQ:
Codex actually reports some issues about #1808, this PR makes its influenced scope bigger and I will fix them while I'm in there.

liangmiQwQ:
@codex

fengmk2 (approved):

PR #1970, opened 2026-06-28
test(env): snapshot list-remote json markers

jong-kyung:
### Description

Follow-up to #1907. Preserve the raw `vp env list-remote --lts --json` output in the global snap test so the JSON marker fields stay covered.

 ### Testing
 - `pnpm -F vite-plus snap-test-global command-env-list-remote`


fengmk2:
I tested this before merging, and it caused unstable output. Also, there was too much content. So I'll stick to using scripts for assertion verification.

PR #1973, opened 2026-06-28
chore(deps): bump vite-task to pick up the Windows ps1 stdin gate

### LINKED ISSUE #1971: Spawn vp-managed package managers via `node <pm>.cjs` directly to remove the Windows .cmd/.ps1 shim hazards
## Summary

Proposal: invoke vp-managed package managers (and `npm`/`npx`) by spawning the real JS entry with the managed Node directly (`node <pm>.cjs <args>`), instead of resolving to the Windows `.cmd` shim and then routing through PowerShell. This removes the root cause behind both the Ctrl+C terminal corruption and the CI stdin hang in every stdio configuration, and drops the PowerShell dependency.

## Background

- #1414 routed `.cmd` shims through PowerShell to fix the `cmd.exe` "Terminate batch job (Y/N)?" Ctrl+C corruption.
- That regressed CI: on Windows runners stdin is not a TTY, and the npm/pnpm/yarn `.ps1` wrappers introspect stdin (`$MyInvocation.ExpectingInput`) and block forever on an open, empty stdin pipe (#1489, #1176, downstream tauri-apps/tauri-action#1308 and voidzero-dev/setup-vp#90). Jobs hung ~6h until GitHub cancelled them.
- Current fix in `crates/vite_command/src/ps1_shim.rs`: skip the PowerShell rewrite when `stdin.is_terminal()` is false. It is sound and well-tested.

## Why the current approach has structural limits

- It couples two unrelated concerns through a heuristic: terminal Ctrl+C corruption is a *console* property, but it is gated on *stdin* TTY-ness.
- It keeps a hard dependency on PowerShell being present plus `-ExecutionPolicy Bypass` for the interactive path.
- It still leaves the `cmd.exe` corruption in place for non-TTY-but-console cases (e.g. `vp install < input.txt` from a real terminal).

## Proposal: bypass the shims for vp-managed binaries

vp already has everything needed to skip the shim layer for the binaries it manages:

- The installer writes the real JS entry next to the shims (`crates/vite_install/src/package_manager.rs`, `crates/vite_install/src/shim.rs`): `<pm>.cjs`, `<pm>.cmd`, and `<pm>.ps1` all just run `node <pm>.cjs`, e.g. `$VP_HOME/package_manager/pnpm/<ver>/pnpm/bin/pnpm.cjs`.
- vp ships its own Node: `$VP_HOME/js_runtime/node/<ver>/node(.exe)`.

So for the managed scope, spawn:

```
$VP_HOME/js_runtime/node/<ver>/node(.exe)   .../pnpm/bin/pnpm.cjs   <args>
```

This is byte-for-byte what the shims do internally, minus the wrapper.

Benefits:

- No `cmd.exe`, so no Ctrl+C "Terminate batch job (Y/N)?" corruption, in every stdio config.
- No `.ps1`, so no `$MyInvocation.ExpectingInput` stdin hang, in every stdio config.
- No PowerShell dependency or execution-policy concerns.
- Removes the TTY heuristic for this path; identical behavior in CI and interactive. vp keeps full control of stdio inheritance and signal handling (`execute_with_terminal_guard`, `fix_stdio_streams`).

## Scope and fallback

Apply this only to what vp owns (the `$VP_HOME` managed shims: package managers plus `npm`/`npx` under `js_runtime`), where the `.cjs` target is known. For arbitrary third-party `node_modules/.bin/*.cmd`, parsing the shim to find the node target is more involved, so the existing TTY-gated PowerShell rewrite in `ps1_shim.rs` can stay as the fallback there. Net result: the highest-value path (the package-manager invocation that caused the CI hang) becomes bulletproof, and the general case keeps today's behavior.

## Implementation sketch

- In `crates/vite_command/src/lib.rs` `resolve_program` (or a layer above it): when the resolved bin is a vp-managed PM / npm / npx shim, resolve to the sibling `<name>.cjs` plus the managed `node`, and return `(node_path, [cjs_path, ...args])` instead of the `.cmd` / PowerShell rewrite.
- Keep `ps1_shim::rewrite_cmd_to_powershell` for the `node_modules/.bin/*` fallback.

## Notes

- setup-vp also ships a defensive mitigation (close child stdin so the `.ps1` wrappers see EOF) for users who pin vp < 0.1.21: voidzero-dev/setup-vp#91.

fengmk2:
## Problem

vite-task's plan-time `.cmd` -> `.ps1` rewrite was ungated and hung Windows CI builds via `vp run <script>` (e.g. `vp run tauri build`): the `.ps1` wrappers block draining a non-TTY stdin pipe, so the process never exits after the build finishes and the runner waits on the open pipe until the job is cancelled hours later.

## Fix

Bump vite-task to `fix/ps1-stdin-tty-gate`, which gates that rewrite on `is_stdin_terminal()` (matching `vite_command`'s ps1_shim). Rev `8daa9bb7` -> `0c5c6236`.

Depends on voidzero-dev/vite-task#491
Closes #1971

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,7 +145,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -156,7 +156,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1865,7 +1865,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -2099,7 +2099,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2299,7 +2299,7 @@ dependencies = [
 [[package]]
 name = "fspy"
 version = "0.1.0"
-source = "git+https://github.com/voidzero-dev/vite-task.git?rev=8daa9bb72faa89b745cb58c087b416b15d3bddc5#8daa9bb72faa89b745cb58c087b416b15d3bddc5"
+source = "git+https://github.com/voidzero-dev/vite-task.git?rev=0c5c6236c9ae3b16c7fd02574d3f87537a938eee#0c5c6236c9ae3b16c7fd02574d3f87537a938eee"
 dependencies = [
  "anyhow",
  "bstr",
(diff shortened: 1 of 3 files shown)


fengmk2:
The verification at https://github.com/fengmk2/digital-signature/pull/1 was successful.

fengmk2:
voidzero-dev/vite-task#491 is merged (squash commit \`6cfa6e47\`), so the rev here is re-pinned from the branch commit to the merged `main` commit. Marking ready.

Verified end-to-end with stock actions (upstream open-stdin `tauri-apps/tauri-action@v0` + `setup-vp@v1`) on two of the originally-affected apps from tauri-apps/tauri-action#1308, using the pkg.pr.new build of this PR:
- fengmk2/digital-signature: full matrix green; Windows build proceeds past `Finished N bundles` instead of hanging.
- fengmk2/redis-me: full matrix green (win32 x64 + arm64).

Both confirm the plan-time ps1 stdin gate resolves the Windows hang on its own, with no tauri-action/setup-vp workaround.

fengmk2:
@codex review

wan9chi (approved):

Issue #1935, opened 2026-06-24
npm shim is not used when running the npm command from a script


Adrian-Baran-GY:
### Describe the bug

When running `npm` command from a script by using `childProcess.execSync` `~/.vite-plus/js_runtime/node/22.22.0/bin/npm` is used instead of `~/.vite-plus/bin/npm`, becuse `~/.vite-plus/js_runtime/node/22.22.0/bin` is added before `/home/ch4cha/.vite-plus/bin` in the `PATH`. Is this expected behaviour? This is confusing as the output from `npm --version` command and  `childProcess.execSync("which npm", { cwd: projectDir, encoding: "utf-8", stdio: "pipe" })` differs. 

### Steps to reproduce

Given following `package.json`:

```json
{
  "name": "test",
  "version": "1.0.0",
  "type": "module",
  "devEngines": {
    "packageManager": {
      "name": "npm",
      "version": "11.17.0",
      "onFail": "download"
    },
    "runtime": {
      "name": "node",
      "version": "22.22.0",
      "onFail": "download"
    }
  }
}
```

we have:

```bash
npm --version
# 11.17.0
which npm
# ~/.vite-plus/bin/npm
```

which is **expected**.

Given `index.js`:

```js
import childProcess from "node:child_process";
import path from "node:path";

const projectDir = path.resolve(import.meta.dirname, ".");

try {
  const output = childProcess.execSync(`which npm`, {
    cwd: projectDir,
    encoding: "utf-8",
    stdio: "pipe",
  });
  console.log(output);
} catch (error) {
  console.error(String(error));
}

try {
  const output = childProcess.execSync(`npm --version`, {
    cwd: projectDir,
    encoding: "utf-8",
    stdio: "pipe",
  });
  console.log(output);
} catch (error) {
  console.error(String(error));
}
```

`node index.js` output is:

```
~/.vite-plus/js_runtime/node/22.22.0/bin/npm

11.14.1
```

`npm` version reported from the script is `11.14.1` instead of `11.17.0`. This is because `~/.vite-plus/js_runtime/node/22.22.0/bin/npm` is used instead of `~/.vite-plus/bin/npm`. Is this expected? I would assume running commands via the script would use the same env as regular commands in the same directory.

### System Info

```shell
VITE+ - The Unified Toolchain for the Web

Environment:
  Version       22.22.0
  Source        devEngines.runtime
  Source Path   ~/ps/test/package.json
  Project Root  ~/ps/test

[...]

Issue #540, opened 2026-02-04
`vp env` shouldn't intercept `npm i -g`


hyf0:
Some notes:

- binaries installed by `npm i -g` will disappear is an already knonw behavior of fnm and nvm. This is a already a solid mental model for node users. I don't see a solid reason to break this mental model. https://github.com/Schniz/fnm/issues/109 also introduce.
- Volta introduces different behavior with an explicitly different command `volta install` instead of `npm i -g`.
  - Notice Votal also intercepts `npm i -g`. For Volta, it's also the same between `volta install` and `npm i -g`

```
  Comparison Table
  ┌───────┬───────────────────────────────────┬──────────────────────────────────────────────────────────┐
  │ Tool  │         Default Behavior          │                   Built-in Workaround                    │
  ├───────┼───────────────────────────────────┼──────────────────────────────────────────────────────────┤
  │ fnm   │ Isolated per-version              │ None (users complain)                                    │
  ├───────┼───────────────────────────────────┼──────────────────────────────────────────────────────────┤
  │ nvm   │ Isolated per-version              │ nvm reinstall-packages to copy from another version      │
  ├───────┼───────────────────────────────────┼──────────────────────────────────────────────────────────┤
  │ mise  │ Isolated per-version              │ .default-npm-packages auto-install, npm backend          │
  ├───────┼───────────────────────────────────┼──────────────────────────────────────────────────────────┤
  │ Volta │ Isolated + pinned to Node version │ Explicit design - volta install with clear communication │
  └───────┴───────────────────────────────────┴──────────────────────────────────────────────────────────┘
```

# My proposal:  Do not make `npm i -g` special

### 1. Principle of Least Surprise

`npm i -g` has had the same behavior for 15+ years. Every Node developer knows:
- Global packages go to the current Node version's directory
- Switching Node versions means different global packages

Intercepting and changing this behavior silently violates user expectations. Even Volta, which does intercept, prints `note: using Volta to install Yarn` — acknowledging it's doing something unexpected.

### 2. Explicit is Better Than Implicit

- `vp env install` explicitly says "I want vite-plus to manage this"
- `npm i -g` should mean "I'm using npm directly"

Users who want isolation should ask for it. Users who want standard npm behavior shouldn't have to learn why their mental model is broken.

### 3. Migration Path from fnm/nvm

fnm and nvm users have muscle memory: `npm i -g` installs to current version, packages disappear when switching. This is annoying but understood.

[...]

Issue #528, opened 2026-02-02
Support `vp env` to manage node version


hyf0:
Behavior about managing global binary

### fnm

- fnm itself doesn't manage global binary. It just relies on `npm install -g`.
- Global binaries installed by `npm install -g` will be cleaned up after switching global node version
- `fnm use/install` is used to install different node

### mise

- Same behavior as fnm for global binaries installed by `npm install -g`
- Global binaries installed by `mise use -g ..` won't be cleaned up after switching global node version
- `mise install`  is used to install different node

## Current proposal for vite-plus

- Same behavior as fnm for global binaries installed by `npm install -g`

### Questions

- What command will us used for installing/switching between node versions?

If we refer to fnm, probably we'll use `vp env use/install`.

- What command will us used for installing global binary isolately

If `vp env use/install` is smart enough, we could still stick with `vp env use/install`

Like: `vp env install -g @anthropic-ai/claude-code`. But in this case, what does `vp env install @anthropic-ai/claude-code` mean?

- How it's different with `vp install -g @anthropic-ai/claude-code` ~~`vite install -g @anthropic-ai/claude-code`~~

IHEL:
vp env支持为单独打开的命令窗口设置node版本，不影响其他窗口吗？

fengmk2:
@IHEL yes, `vp env use <version>` sets a Node.js version for the current shell session

IHEL:
但是我现在在windows上开多个PowerShell窗口执行vp env use <version>后，窗口之间来回切换node -v查看一直显示最后执行env use的版本，是不是现在还不稳定有问题呢？

fengmk2:
@IHEL There is indeed an issue. I discovered that on Windows it's using the global configuration file mode. You can resubmit an issue to report this problem, and I will follow up on it.

fengmk2:
@IHEL Will try to fix in https://github.com/voidzero-dev/vite-plus/pull/1577

PR #1974, opened 2026-06-28
fix(env): only prepend core tools from Node.js runtime

liangmiQwQ:
This could be a follow-up to #1935.

Vite+ now prepends node's bin directory to `$PATH`, in this directory, it includes not only core Node.js tools, but other installed tools, like `npm install -g typescript`'s `tsc`. It can't be access directly from shells, but this prepending makes it accessible from scripts.

Another edge case: if you use Node.js to launch your terminal / editor (like [`@liangmi/mo`](https://github.com/liangmiQwQ/mo)), the shell's `$PATH` will include the Node's bin as well. When this runtime's bin is included, `npm install -g` will stop prompting users to add bins. Make you can't access this bin anymore after you exist this terminal.

After this PR, Vite+ will no longer directly prepend the `bin` directory to the `$PATH`. Instead, Vite+ will lazily create a `core` directory, including symlinks to Node.js's core tools (`npm`, `node`, `npx`), and prepend `PATH` with this directory. It makes sure installed tools won't appear and disappear all the time.

🤖 Generated with Codex

### DIFF
--- a/crates/vite_global_cli/src/commands/env/exec.rs
+++ b/crates/vite_global_cli/src/commands/env/exec.rs
@@ -9,7 +9,7 @@
 
 use std::process::ExitStatus;
 
-use vite_js_runtime::NodeProvider;
+use vite_js_runtime::{NodeProvider, ensure_node_core_bin_prefix};
 use vite_shared::{env_vars, format_path_prepended};
 
 use crate::{
@@ -137,10 +137,9 @@ async fn execute_with_version(
         std::env::remove_var(env_vars::VP_TOOL_RECURSION);
     }
 
-    // 4. Build PATH with node bin dir first (uses platform-specific separator)
-    // Always prepend to ensure the requested Node version is first in PATH
-    let node_bin_dir = runtime.get_bin_prefix();
-    let new_path = format_path_prepended(node_bin_dir.as_path());
+    // 4. Build PATH with the limited core bin dir first (uses platform-specific separator)
+    let core_bin_dir = ensure_node_core_bin_prefix(&runtime.get_binary_path())?;
+    let new_path = format_path_prepended(core_bin_dir.as_path());
 
     // 5. Execute command
     let (cmd, args) = command.split_first().unwrap();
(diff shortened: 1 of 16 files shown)


liangmiQwQ:
Maybe I should learn https://fengmk2.com/posts/blog-qwin-windows-testing-on-macos carefully haha.

fengmk2:
> Maybe I should learn https://fengmk2.com/posts/blog-qwin-windows-testing-on-macos carefully haha.

don't do that, it was slop. I only test on real windows by hand.

liangmiQwQ:
@codex review

fengmk2:
This fix will require us to adapt any new executables added to the bin directory in Node.js in the future, and it will be difficult to spot the problem. Maintaining the current implementation is actually to be expected, since users install the global CLI via `npm install -g`, and it's expected they will have access to it.

liangmiQwQ:
Emm, so let's close it.

> Vite+ now prepends node's bin directory to `$PATH`, in this directory, it includes not only core Node.js tools, but other installed tools, like `npm install -g typescript`'s `tsc`. It can't be access directly from shells, but this prepending makes it accessible from scripts.
> 
> Another edge case: if you use Node.js to launch your terminal / editor (like [`@liangmi/mo`](https://github.com/liangmiQwQ/mo)), the shell's `$PATH` will include the Node's bin as well. When this runtime's bin is included, `npm install -g` will stop prompting users to add bins. Make you can't access this bin anymore after you exist this terminal.

@fengmk2 However, I still think there is room to improve the current behavior. Actually, in my opinion, the current `npm install -g` seems kind of strange, its current duty is similar to `vp install -g`, and our modify (hooks) make its behavior change in different cases. 

I've checked #528, #540. According to them, if I have a chance to re-design the `vp env` system, I will make Vite+ always inject runtime's `bin` directory to `$PATH`, let `npm` do not have any wrapper, and make the bin only usable when using the node used to install it. I believe it is a approach to keep code and behavior both clean.

fengmk2:
Alternatively, we could improve this by intercepting `npm install -g` and prompting the user to use `vp install -g` instead.

liangmiQwQ:
Trace in #1986

PR #1976, opened 2026-06-28
test: update pnpm vite alias snapshots

wyf027:
## Summary
- update affected snap outputs for the pnpm Vite alias lint rule
- make the Astro shim snap setup remove the full generated lint block instead of deleting individual lines

## Why
The new rule causes generated `rules` config to format across multiple lines. The old Astro snap setup only deleted single-line lint fields, which left invalid config once the rules block expanded.

## Validation
- `pnpm -F vite-plus snap-test-local command-init-inline-config create-org-bundled-monorepo`
- `git diff --check`

Note: global snap tests need built CLI dist artifacts; local `build-ts` is blocked here by missing upstream `rolldown/dist` build output.


wyf027:
Closing this helper PR to avoid adding noise while the original branch owner decides whether to continue with the fix.

PR #1978, opened 2026-06-28
fix(command): resolve fspy program before spawn

### LINKED ISSUE #1599: Windows: `cargo test -p vite_command` fspy tests fail when `node` resolves to Vite+ shim
## Summary

On Windows, `cargo test -p vite_command` can fail in the `run_command_with_fspy` tests when `node` resolves through the Vite+ trampoline shim in `~/.vite-plus/bin/node.exe`.

The failing tests call `run_command_with_fspy("node", ...)`. Unlike `run_command`, the fspy path does not resolve the binary through `resolve_program`, so PATH resolution happens inside `fspy::Command`. In an environment where Vite+ managed shims are earlier on PATH, `node` resolves to the trampoline shim, which dispatches to `~/.vite-plus/current/bin/vp.exe`. Under fspy, the trampoline fails to spawn `vp.exe`, causing the child status to be nonzero.

## Reproduction

On Windows with Vite+ shims on PATH:

```powershell
cargo test -p vite_command
```

Observed failures:

```text
vite-plus: failed to execute C:\Users\fengmk2\.vite-plus\current\bin\vp.exe

failures:
    tests::run_command_with_fspy_tests::test_run_command_with_fspy
    tests::run_command_with_fspy_tests::test_run_command_with_fspy_and_capture_path_accesses_write_file
    tests::run_command_with_fspy_tests::test_run_command_with_fspy_and_capture_path_accesses_write_and_read_file
```

The failed assertions are `assert!(cmd_result.status.success())` in `crates/vite_command/src/lib.rs`.

## Notes

`C:\Users\fengmk2\.vite-plus\current\bin\vp.exe --version` works outside fspy, so this appears specific to the fspy-wrapped child process plus Vite+ trampoline resolution.

The likely issue is that `run_command_with_fspy` is environment-sensitive on Windows because it passes the bare binary name directly to fspy:

```rust
let mut cmd = fspy::Command::new(bin_name);
```

while `run_command` first resolves the program via `resolve_program`, including Windows shim handling.

## Possible fix directions

- Make `run_command_with_fspy` resolve the binary consistently with `run_command` before spawning.
- In tests, avoid depending on ambient PATH by resolving a real Node executable explicitly or by filtering Vite+ shim paths out of PATH.
- Investigate why the trampoline cannot spawn `vp.exe` under fspy and whether that path should be supported directly.

wyf027:
## Summary

- Resolve `run_command_with_fspy` binaries through the same `resolve_program` path used by `run_command`.
- Preserve Windows `.cmd` to PowerShell prefix args before user args when fspy is enabled.
- Update the missing-binary test expectation to match the shared resolver error.

## Why

`run_command_with_fspy` passed the bare binary name directly to `fspy::Command::new`, so PATH resolution happened inside fspy instead of using Vite+'s resolver. On Windows this can make fspy pick up environment-sensitive shim paths differently from `run_command`. Reusing `resolve_program` keeps both command paths aligned and follows the fix direction described in #1599.

Fixes #1599.

## Validation

- `rustfmt --edition 2024 crates/vite_command/src/lib.rs`
- `git diff --check`
- Confirmed `fspy::Command::new<P: AsRef<OsStr>>` in `voidzero-dev/vite-task` at the locked fspy rev accepts the resolved path argument.

Cargo tests could not run in this local checkout because the workspace references `rolldown/crates/rolldown_binding/Cargo.toml`, but the `rolldown/` directory is not present here.

### DIFF
--- a/crates/vite_command/src/lib.rs
+++ b/crates/vite_command/src/lib.rs
@@ -206,16 +206,19 @@ where
     S: AsRef<OsStr>,
 {
     let cwd = cwd.as_ref();
+    let (program, prefix_args) = resolve_program(bin_name, envs, cwd)?;
     let args: Vec<OsString> = args.into_iter().map(|s| s.as_ref().to_owned()).collect();
     tracing::debug!(
         target: "vite_command::spawn",
-        bin_name,
+        program = %program.as_path().display(),
+        prefix_args = ?prefix_args,
         args = ?args,
         cwd = %cwd.as_path().display(),
         "spawn (fspy)",
     );
-    let mut cmd = fspy::Command::new(bin_name);
-    cmd.args(&args)
+    let mut cmd = fspy::Command::new(program.as_path());
+    cmd.args(&prefix_args)
+        .args(&args)
         // set system environment variables first
         .envs(std::env::vars_os())
         // then set custom environment variables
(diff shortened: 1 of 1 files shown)


wyf027:
Closing this draft PR to avoid duplicating prior unmerged work and adding extra review noise.

PR #1992, opened 2026-06-30
docs: Vite Task cache reuse and tool-reported caching

wan9chi:
- Add an experimental guide for reusing Vite Task cache with GitHub Actions cache.
- Add "Automatic Tracking" explainer that contains fs tracking and cache-reporting tools (previously called runner-aware tools)
- Add docs for `{ task: '...', from: 'depedencies' }` in `dependsOn`.


docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,232 @@
...
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: voidzero-dev/setup-vp@v1
+        with:
+          node-version: '24'
+          cache: true
```
fengmk2, line 86: `cache: true` can be removed, it's enable by default.

docs/guide/automatic-tracking.md
```diff
@@ -0,0 +1,133 @@
+# Automatic Tracking
```
TheAlexLichter:
```suggestion
# Automatic Dependency Tracking
```

Just "Automated tracking" feels a bit short / lacks context I think.

wan9chi: “Dependency” can be confused with package/task dependencies. And it's more than dependencies: output files are also tracked.

How about "Automatic Cache Tracking"?

docs/.vitepress/config.mts
```diff
@@ -16,6 +16,14 @@ const taskRunnerGuideItems = [
     text: 'Task Caching',
     link: '/guide/cache',
   },
+  {
+    text: 'Automatic Tracking',
```
cpojer: It's unclear what this means in the sidebar. How about "Automatic File Tracking"?

I also suggest moving this and the next one below "Running Binaries" as that is usually more useful for an initial user.

docs/config/run.md
```diff
@@ -120,9 +120,11 @@ Commands joined with `&&` (or supplied as an array) are automatically split into
 
 ### `dependsOn`
 
-- **Type:** `string[]`
+- **Type:** `Array<string | { task: string, from: DependsOnFrom }>`
 - **Default:** `[]`
 
+`DependsOnFrom` accepts `"dependencies"`, `"devDependencies"`, `"peerDependencies"`, or an array of those values.
```
cpojer: I don't really understand what this means. I don't think adding the type here is valuable. How about:

> `from` accepts any package name listed in the dependencies of the project.

cpojer: Oh, I misunderstood it. The added documentation is confusing to me. Why do we have to specify the type of dependencies? What is the value of doing that?

docs/config/run.md
```diff
@@ -270,26 +291,39 @@ String glob patterns are resolved relative to the package directory by default.
 
 ### `output`
 
-- **Type:** `Array<string | { pattern: string, base: "workspace" | "package" }>`
-- **Default:** `[]` (nothing is archived)
+- **Type:** `Array<string | { auto: boolean } | { pattern: string, base: "workspace" | "package" }>`
+- **Default:** automatic write tracking
+
+Files Vite Task archives after a successful run and restores on a cache hit.
```
cpojer: This sentence is harder to understand than the previous version. Maybe something like this:

"Vite Task automatically archives the files generated from a successful task run and restores them on a cache hit."

and then launch into the explanation of manually providing the `output` field.

docs/config/run.md
```diff
@@ -270,26 +291,39 @@ String glob patterns are resolved relative to the package directory by default.
...
 tasks: {
   build: {
-    command: 'vp build',
+    command: 'node build.mjs',
     output: ['dist/**', '!dist/cache/**'],
   },
 }
 ```
 
+Use `{ auto: true }` to keep automatic write tracking while adding explicit output globs:
```
cpojer: This could benefit from an explanation of why I'd wanna do this.

docs/config/run.md
```diff
@@ -298,6 +332,19 @@ tasks: {
 }
 ```
 
+Set `output: []` to disable output restoration for a cached task:
```
cpojer, line 337: Why is this desirable over disabling the cache?

docs/guide/automatic-tracking.md
```diff
@@ -0,0 +1,133 @@
...
+
+Automatic tracking is how Vite Task learns what a task needs for caching without explicit config.
```
cpojer:
```suggestion
Automatic tracking is how Vite Task learns what inputs a task needs for caching outputs without explicit config.
```

docs/guide/automatic-tracking.md
```diff
@@ -0,0 +1,133 @@
...
+    tasks: {
+      build: {
+        command: 'tsc',
+      },
+    },
+  },
+});
+```
+
+For this task, Vite Task records source files, config files, missing files the command checked, and directories the command scanned. A later run misses the cache when one of those tracked inputs changes.
```
cpojer:
```suggestion
For this task, Vite Task records source files, config files, missing files the command checked, and directories the command scanned. Subsequent runs re-run the task when one of those tracked inputs changes.
```

docs/guide/automatic-tracking.md
```diff
@@ -0,0 +1,133 @@
...
+
+File system tracking also tracks outputs. If you omit [`output`](/config/run#output), Vite Task archives files the command writes after a successful run and restores them on a cache hit.
+
+### Limitations
+
+Vite Task cannot track environment variable reads, and it cannot always tell which tracked paths are stable inputs, generated outputs, or tool-managed cache paths that should not become inputs or outputs.
+
+Use [Override Inputs And Outputs](#override-inputs-and-outputs) when file system tracking includes files that should not affect the cache, misses files that should, or restores the wrong outputs.
+
+Use [`env`](/config/run#env) when a command needs an environment variable and the value should affect the cache. Use [`untrackedEnv`](/config/run#untrackedenv) when a command needs an environment variable but the value should not affect the cache.
```
cpojer:
```suggestion
Use [`env`](/config/run#env) when a command needs an environment variable and the value should affect the cache, or [`untrackedEnv`](/config/run#untrackedenv) when the value should not affect the cache.
```

docs/guide/automatic-tracking.md
```diff
@@ -0,0 +1,133 @@
...
+
+These limitations do not apply to `vp build`: Vite reports [Cooperative Tracking](#cooperative-tracking) metadata automatically, including `VITE_*`, `NODE_ENV`, and Vite-managed cache paths that should not become inputs or outputs. A standard `vp build` task does not need manual `input`, `output`, or `env`.
+
+### Override Inputs And Outputs
+
+[`input`](/config/run#input) controls what invalidates the cache.
+[`output`](/config/run#output) controls which files Vite Task restores on a cache hit.
```
cpojer:
```suggestion
[`input`](/config/run#input) controls what invalidates the cache. [`output`](/config/run#output) controls which files Vite Task restores on a cache hit.
```

docs/guide/automatic-tracking.md
```diff
@@ -0,0 +1,133 @@
...
+tasks: {
+  lint: {
+    command: 'vp lint',
+    // Disable automatic input tracking and fingerprint only these files.
+    input: ['src/**', 'vite.config.ts'],
+  },
+}
+```
+
+Set `input: []` when no files should affect the cache fingerprint. Set `output: []` when no files should be restored on a cache hit.
```
cpojer: What's a real world example where `input: []` is appropriate?

docs/guide/automatic-data-tracking.md
```diff
@@ -0,0 +1,133 @@
...
+```ts [vite.config.ts]
+import { defineConfig } from 'vite-plus';
+
+export default defineConfig({
+  run: {
+    tasks: {
+      build: 'vp build',
+    },
+  },
+});
```
cpojer, line 126: This is where Vite+ becomes incredibly confusing because you get a cache hit with `vpr build` but never with `vp build`. It would be great if we could add caching to `vp build` and `vp pack` through a config option instead.

It might also make sense to rename the task name here to something like "frontendBuild" and then explain it can be run with `vpr frontendBuild` or similar.

docs/guide/cache.md
```diff
@@ -4,32 +4,19 @@ Vite Task can automatically track dependencies and cache tasks run through `vp r
 
 ## Overview
 
-When a task runs successfully (exit code 0), its terminal output (stdout/stderr) is saved. On the next run, Vite Task checks if anything changed:
+When a task runs successfully (exit code 0), its terminal output (stdout/stderr) and all files it writes (output files) are saved. On the next run, Vite Task checks if anything changed:
```
cpojer:
```suggestion
When a task runs successfully (exit code 0), its terminal output (stdout/stderr) and all written files (output files) are saved. On the next run, Vite Task checks if anything changed:
```

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
+# GitHub Actions Cache
+
+::: warning Experimental
+Reusing Vite Task cache across GitHub Actions runs is experimental. Test and measure it in your project before relying on it in CI.
```
cpojer:
```suggestion
Reusing Vite Task's cache across GitHub Actions runs is experimental. Test and measure it in your project before relying on it in CI.
```

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+:::
+
+Vite Task stores task results in `node_modules/.vite/task-cache` at the workspace root. Restore that directory in later GitHub Actions runs so Vite Task can reuse previous task results.
```
cpojer, line 7: Does this cause any issues with what package managers are caching?

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+
+1. `actions/cache` restores and saves the cache directory based on the key in your workflow.
+2. Vite Task uses the restored cache directory and replays only the tasks whose fingerprints still match.
+
+## Before You Start
+
+Use this workflow when all of these are true:
+
+- The command runs through [`vp run`](/guide/run).
+- The task hits on a second local run.
```
cpojer: I know what this means, but this sentence is probably unclear to most people.

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+- The task has stable input and output tracking for CI.
+- The workflow installs dependencies before restoring `node_modules/.vite/task-cache`.
+
+Fix local misses first. GitHub Actions cache can move Vite Task's local cache directory between runs, but it cannot make an unstable task cacheable.
```
cpojer: This reads too llm written and is not actionable. People won't know how to fix it, please add a link on how to figure out how to make the caching stable.

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Measure Before Caching Across Runs
+
+Measure before you add GitHub Actions cache for Vite Task. Compare workflow duration with and without the restore and save steps. Check both the GitHub cache step time and the `vp run` time.
```
cpojer: This section should explain first why measuring matters. Basically the bullet list at the end should probably just be a paragraph in the beginning instead.

Also, a typo:
```suggestion
Measure before you add a GitHub Actions cache for Vite Task. Compare workflow duration with and without the restore and save steps. Check both the GitHub cache step time and the `vp run` time.
```

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+Run each task twice:
+
+```bash
+vp run build
+vp run build # should print "cache hit"
+vp run lint
+vp run lint # should print "cache hit"
+```
+
+## 2. Restore The Cache After Install
```
cpojer, line 62: I don't really understand why this isn't just a flag on `setup-vp`? cc @fengmk2

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+vp run: cache hit, 1.10s saved.
+```
+
+If GitHub restores a cache but Vite Task prints a cache miss, the workflow restored the cache directory, but the task fingerprint changed.
+
+## Keep Task Tracking Stable
+
+If GitHub restores a cache but `vp run` prints a cache miss, fix the task fingerprint before changing the Actions cache key. See [Automatic Tracking](/guide/automatic-tracking) and [`run.tasks`](/config/run#tasks).
+
+## Choose A Cache Key
```
cpojer, line 134: If this was part of setup-vp, we could have a good default for this so not everyone has to customize the cache keys.

docs/guide/github-actions-cache.md
fengmk2: I think it can be built into setup-vp, but let's not block the merge of this PR for now? I'll look into how to integrate this feature into setup-vp later.

cpojer: Yes, I agree.

docs/.vitepress/config.mts
wan9chi: Envs are also tracked. How about  "Automatic Cache Tracking"?

This page and "GitHub Actions Cache" only make sense after the reader has already seen the "Task Caching" page. How about putting them under it?

image

cpojer: Yes, grouping makes sense! Maybe "Automatic Data Tracking" is best then since it discusses both inputs and outputs?

docs/config/run.md
wan9chi: In turborepo, `^build` includes all types of dependencies except `peerDependencies`. I believe they do this for a reason but I also find complaints about this and there being no way to configure it.

Since there isn't a default that makes everyone happy, I think we should let users be explicit about it.

docs/guide/github-actions-cache.md
```diff
@@ -0,0 +1,164 @@
...
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: voidzero-dev/setup-vp@v1
+        with:
+          node-version: '24'
```
fengmk2, line 85: @wan9chi Sorry, I'm wrong, cache is `false` by default, we need to set `cache: true` here.

docs/.vitepress/config.mts
wan9chi: 👌 updated in f0fe52832eb4793bc3d7d4f42185d4c97b17e542

docs/guide/github-actions-cache.md
wan9chi: It could if the whole `node_modules` is cached 🤔 I'll look into it (logged in https://github.com/voidzero-dev/vite-task/issues/502).

docs/config/run.md
wan9chi: Updated in 994f7e1e0adfd3d7250d5aed52515eb0836c37f1.

docs/config/run.md
wan9chi: Updated in 75f0f712ce81f006cddf4344280b1c1377cf524f.

docs/config/run.md
wan9chi: Could be useful for local-only cache where outputs are already there. Added this use case in ce24fd920172f49a4fc82724f4ae1fd6e0d10712.

docs/guide/automatic-tracking.md
wan9chi: fd7df85be1c5d71c05b685f19c18f03ad83680bd

docs/guide/automatic-tracking.md
wan9chi: d77b13600baa218484d1701fec0495f2928ebca7

docs/guide/automatic-tracking.md
wan9chi: 6417d4de9ef33fadd759b33661af449cfba940b9

docs/guide/automatic-tracking.md
wan9chi: efbe1ed9371a462bcacc1cb86b445d592d358c29

docs/guide/automatic-tracking.md
wan9chi: Could be useful for a download task.
Added the example in 888dc9ca624d22193d9163b5eeac4c469df43dda.

docs/guide/automatic-data-tracking.md
wan9chi: We decided not to cache non-`run` subcommands, but I did hear feedbacks about this confusion. #2011 is logged. Let's revisit it later.

Task name updated in 073f50f2e3165102034188bf024e5441585894ec.

docs/guide/cache.md
wan9chi: 3050091d826173065d117096d226c7daf707982c

docs/guide/github-actions-cache.md
wan9chi: 310a789c642b5fcad6ec8616acfc76b44a024e12

docs/guide/github-actions-cache.md
wan9chi: Rephrased in abdc11b2da4c3a0863c3ce136572c19946394f5e.

docs/guide/github-actions-cache.md
wan9chi: Rephrased and added link to the "When To Add Manual Config" section in bdd9135fdb325146e2e359c50bc4a6e80698ea2d

docs/guide/github-actions-cache.md
wan9chi: Reordered and fixed typo in a1c2d1be8996024db175236a34503fc76c5cd00c

docs/guide/github-actions-cache.md
wan9chi: Added back in 3491c31b3560850ce8d153881e44b9ef002b4732

docs/guide/github-actions-cache.md
wan9chi: https://github.com/voidzero-dev/vite-plus/issues/2013 logged

docs/config/run.md
wan9chi: Rephrased in ec324fd6f127a7d97f3878b147ecce0282435591

wan9chi:
@cpojer Thank you for such detailed reviews! I've addressed all of them. Let me know if you have any more concerns.

docs/guide/automatic-data-tracking.md
cpojer: Yeah we should definitely not cache by default, but it could be opt-in.

cpojer:
Nice work on the docs!

fengmk2 (approved):

fengmk2:
@wan9chi let's go!

docs/.vitepress/config.mts
```diff
@@ -15,6 +15,10 @@ const taskRunnerGuideItems = [
   {
     text: 'Task Caching',
     link: '/guide/cache',
+    items: [
+      { text: 'Automatic Data Tracking', link: '/guide/automatic-tracking' },
```
TheAlexLichter: can we update the filename + link here to be in line with the heading please?

TheAlexLichter (changes requested):
minor detail (but hard to fix later for SEO 🙈)

docs/.vitepress/config.mts
wan9chi: Good catch! Updated.

TheAlexLichter (approved):

PR #2022, opened 2026-07-02
rfc: global -C flag for working-directory switching

fengmk2:
RFC for a global `-C <dir>` flag and the monorepo app-command UX built on it:

- `vp -C <dir> <cmd>` behaves exactly like `cd <dir> && vp <cmd>`, for every vp command (the `git -C` / `make -C` convention). This is the first-class "run there" primitive vp lacks today; it also gives `vp pack` its missing directory form.
- Fully backward compatible: `vp dev <path>` keeps upstream Vite semantics (`root` only), pack positionals stay tsdown entries, no flags are added to the subcommands themselves.
- Built on `-C`, the app commands (`dev`/`build`/`preview`/`pack`) stop being silently wrong at a workspace root: an interactive fuzzy package picker in a TTY, a `defaultPackage` root config (also covers non-workspace framework repos like Laravel with a `frontend/` directory), and a clear package listing plus exit 1 when non-interactive. All three expand to an implicit `-C`.

Pain points are reproducible: https://github.com/why-reproductions-are-required/vite-plus-monorepo-app-commands-repro

Full UX mockups, behavior spec, decisions, and open questions are in `rfcs/cwd-flag.md`.


rfcs/monorepo-app-commands.md
```diff
@@ -0,0 +1,324 @@
...
+├── routes/
+├── composer.json
+├── vite.config.ts     (root config below)
+└── frontend/          (the Vite app)
+```
+
+There is no `pnpm-workspace.yaml` or `workspaces` field here, so the picker has no package list to enumerate. `defaultProject` is what makes `vp dev` at the root work in this shape:
+
+```ts
+import { defineConfig } from 'vite-plus'
```
fengmk2: There's a contradiction here, vite-plus is not installed in this directory, so it cannot be imported.

fengmk2 (commented):
Left a few inline RFC comments.

rfcs/monorepo-app-commands.md
```diff
@@ -0,0 +1,324 @@
...
+
+After this RFC the following holds and is documented:
+
+```
+vp <cmd> <path> [args...]  ===  cd <path> && vp <cmd> [args...]
+```
+
+for `cmd` in `dev`, `build`, `preview`, `pack`. Concretely: the child process spawn cwd is `<path>`, so config lookup, `.env` loading, `process.cwd()` reads in configs and plugins, relative CLI args, and local `vite-plus` resolution all behave as if the user had `cd`'d. The parent `vp` process never calls `process.chdir()`.
+
+### Picker contents
```
fengmk2: I think we should clarify the scope of the equivalence here. Given our supported monorepo model assumes a single `vite-plus` version across the workspace, resolving through the root local CLI is fine. But as written, including “local `vite-plus` resolution” can read like per-package local CLI resolution is also guaranteed to be byte-for-byte equivalent.

Maybe narrow this sentence to config lookup, `.env`, `process.cwd()` reads, and relative CLI args, then add something like:

```md
This assumes a workspace uses a single Vite+ version, which is already the supported monorepo model.
```

rfcs/monorepo-app-commands.md
```diff
@@ -0,0 +1,324 @@
...
+├── vite.config.ts     (root config below)
+└── frontend/          (the Vite app)
+```
+
+There is no `pnpm-workspace.yaml` or `workspaces` field here, so the picker has no package list to enumerate. `defaultProject` is what makes `vp dev` at the root work in this shape:
+
+```ts
+import { defineConfig } from 'vite-plus'
+
+export default defineConfig({
```
fengmk2: This example still feels contradictory for non-JS framework repos. The section says there is no JS workspace metadata, but the root config imports `vite-plus`. If `vite-plus` is only installed under `frontend/`, the root config cannot import it.

I think we should either make this example explicitly require a JS root that has `vite-plus` installed, or use a config location that works without importing `vite-plus` from the non-JS root.

rfcs/monorepo-app-commands.md
```diff
@@ -0,0 +1,324 @@
...
+vp <cmd> <path> [args...]  ===  cd <path> && vp <cmd> [args...]
+```
+
+for `cmd` in `dev`, `build`, `preview`, `pack`. Concretely: the child process spawn cwd is `<path>`, so config lookup, `.env` loading, `process.cwd()` reads in configs and plugins, relative CLI args, and local `vite-plus` resolution all behave as if the user had `cd`'d. The parent `vp` process never calls `process.chdir()`.
+
+### Picker contents
+
+- One row per workspace package: package name plus relative path, sorted by path.
+- Fuzzy search over name and path via `vite_select::fuzzy_match`, paging identical to the task picker.
+- v1 lists **all** workspace packages. Filtering or annotating rows by command relevance (has `vite.config.*` / `index.html` for `dev`, has pack config for `pack`) is a possible refinement, kept out of v1 to avoid guessing wrong and hiding valid targets.
```
fengmk2: Non-blocking: listing all packages equally may make `vp dev` noisy in large monorepos with many libraries. Could we sort or annotate likely runnable packages first, e.g. packages with `vite.config.*`, `index.html`, or relevant scripts, while still keeping all packages searchable so we don't hide valid targets?

fengmk2:
RFC moved into the implementation PR #2031 as rfcs/cwd-flag.md; review continues there.

PR #1991, opened 2026-06-30
ci: skip full CI for docs-only PRs

wan9chi:
## Summary

- Exclude `docs/**` from the main CI workflow's `code` path filter.
- Set `predicate-quantifier: every` so negative path patterns combine correctly.

## Why

Docs-only PRs can include non-Markdown docs files such as `docs/.vitepress/config.mts`. The old filter treated any non-`.md` file as code, which scheduled full CI, CLI E2E, and snapshot matrices for docs-only changes.

PR #2028, opened 2026-07-03
ci: fast docs format check for docs-only PRs

fengmk2:
Since #1991, docs-only PRs skip the full CI matrix, so nothing format-checks markdown (`vp check` in cli-e2e-test only runs for code changes). As suggested in [this comment](https://github.com/voidzero-dev/vite-plus/pull/1991#issuecomment-4839488485), add a fast `docs-fmt` job that runs `vp fmt --check` when `code-changed == 'false'`.

How it stays fast (about 15s total on the runner, the check itself is about 5s for 1,879 files):

- `setup-vp` installs the released Vite+ (`version: latest`, since the workspace `vite-plus` version can be ahead of the registry) with `run-install: false`.
- `vp fmt` needs the root `vite.config.ts` to load, and its runtime `import ... from 'vite-plus'` must resolve. Instead of installing the whole workspace (which also needs the vite/rolldown subrepo clones), the job symlinks the released `vite-plus` package out of the global install into `node_modules/`.
- Checks the whole repo, so root markdown (README, RFCs) is covered too, matching the `**/*.md` part of the CI skip filter.

Non-blocking as suggested: the job is not listed in the `done` job's needs.

Verified on this PR by temporarily lifting the skip condition (middle commits, since ci.yml itself is a code change the job would otherwise skip here):

- [Failing run](https://github.com/voidzero-dev/vite-plus/actions/runs/28635962494/job/84922274235) on an intentionally misformatted canary file
- [Passing run](https://github.com/voidzero-dev/vite-plus/actions/runs/28636039893/job/84922566220) after removing it

The final commit restores the `code-changed == 'false'` guard, so the job is expected to be skipped on this PR itself.


.github/workflows/ci.yml
```diff
@@ -59,6 +59,33 @@ jobs:
...
+          run-install: false
+
+      - name: Run vp fmt --check
+        run: |
+          # `vp fmt` loads the root vite.config.ts, whose `import ... from 'vite-plus'`
+          # must resolve at runtime. Link the released package out of the global
+          # install instead of installing the whole workspace.
+          mkdir -p node_modules
+          ln -s "$HOME/.vite-plus/current/node_modules/vite-plus" node_modules/vite-plus
+          vp fmt --check
```
fengmk2, line 87: Should find the real oxfmt version in our repo to run it?

fengmk2: Done in 9adcc4c1. It turns out vp resolves oxfmt project-first (`process.cwd()` before its bundled copy, see `packages/cli/src/utils/constants.ts` `resolve()`), so the job now installs the pnpm catalog's pinned oxfmt (`=0.57.0`) and links it into `node_modules/`; vp then runs exactly that version. I verified the resolution order empirically with a stub project-local oxfmt (vp executed the stub) and ran the full script locally against a bare checkout with the pinned install (all 1,878 files pass).

One extra wrinkle the commit handles: oxfmt itself loads the `fmt` block from `vite.config.ts` and resolves the config's `vite-plus` import from its own realpath, so the temp oxfmt install also gets a `vite-plus` sibling symlink.

Note the released vp currently bundles oxfmt 0.57.0 too, so behavior is identical today; the pin protects against future drift in either direction.

cpojer (approved):

.github/workflows/ci.yml
```diff
@@ -59,6 +59,33 @@ jobs:
...
+  docs-fmt:
+    name: Docs format check
+    needs: detect-changes
+    # Docs-only changes skip the full CI matrix, so nothing else format-checks
+    # them (code changes get `vp check` in cli-e2e-test). Non-blocking:
+    # intentionally not listed in the `done` job's needs.
+    if: needs.detect-changes.outputs.code-changed == 'false'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: taiki-e/checkout-action@7d1e50e93dc4fb3bba58f85018fadf77898aee8b # v1.4.2
```
cpojer, line 71: What is taiki-e??

fengmk2: https://github.com/taiki-e/checkout-action#why-not-actionscheckout A safer checkout action, love by @Boshen

PR #2029, opened 2026-07-03
fix(global): respect custom VP_HOME

liangmiQwQ:
At present, Vite+ allows users to specific `VP_HOME` as an installation directory. However, for any future `vp` runs without manual `VP_HOME` environment, we still use default `~/.vite-plus` or `vp`'s grandparent directory if its name is `.vite-plus` (It's related to #1185 by me).

It will cause mess, like multiple Vite+ instances, vp binary not finding, packages unusable and a lot of unexpected behavior.

This PR improves all related logic, include:

- Set `VP_HOME` in env files
- Improve prompt and implode check related to `VP_HOME`
- Respect `VP_HOME` in `pnpm boostrap-cli` at devtime.

The new `install.sh` prompt looks like that:
image

🤖 Generated by Codex


liangmiQwQ:
@codex review

liangmiQwQ:
@codex review

liangmiQwQ:
@fengmk2 Worth mentioning: after this change, Vite+ will start injecting `VP_HOME` in shellrc, just like Volta. (`VOLTA_HOME`). It may also help to check whether Vite+ is installed for other tools :)

fengmk2:
> @fengmk2 Worth mentioning: after this change, Vite+ will start injecting `VP_HOME` in shellrc, just like Volta. (`VOLTA_HOME`). It may also help to check whether Vite+ is installed for other tools :)

Good news, maybe after your pr merge, my one can be closed https://github.com/voidzero-dev/vite-plus/pull/1958 😄

fengmk2:
@nekomoyi @naokihaba I need your help to see if this added environment variable will have any side effects.

liangmiQwQ:
> Good news, maybe after your pr merge, my one can be closed #1958 😄

Oh? I haven't seen it, I would take a look at it later.

fengmk2:
If I have previously installed vite-plus, and this time I reinstall it and set `VP_HOME` to a new directory, should we prompt the user whether they need to remove the previous `~/.vite-plus` directory?

liangmiQwQ:
> should we prompt the user whether they need to remove the previous `~/.vite-plus` directory?

I think that makes sense, and I've implemented in 43f7d23

fengmk2:
@codex review

liangmiQwQ:
@codex review

packages/cli/install.sh
```diff
@@ -132,6 +137,103 @@ write_release_age_override() {
...
+      if [ ! -f "$vp_bin" ]; then
+        vp_bin="$old_dir/current/bin/vp.exe"
+      fi
+      if [ ! -f "$vp_bin" ]; then
+        warn "Could not remove previous Vite+ install at $old_dir: vp binary not found."
+        return 0
+      fi
+
+      local implode_output
+      if implode_output=$(VP_HOME="$old_dir" "$vp_bin" implode --yes 2>&1); then
```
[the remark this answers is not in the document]
liangmiQwQ: I won't fix this, because the new installing version can be customized. Using its own implode script is good practice I believe.

liangmiQwQ:
@codex review

liangmiQwQ:
@codex review

liangmiQwQ:
This PR lets me run `vp implode` for the first time lol.

fengmk2 (approved):
@liangmiQwQ Nice improve!

I will do a round simplify before I merge.

crates/vite_global_cli/src/commands/env/setup.rs
```diff
@@ -742,37 +746,56 @@ Register-ArgumentCompleter -Native -CommandName vpr -ScriptBlock $__vpr_comp
...
+        EnvShell::Fish => ENV_TEMPLATE_FISH
+            .replace("__VP_HOME__", &home_path_ref)
+            .replace("__VP_BIN__", &bin_path_ref),
         EnvShell::Nu => {
             // Nushell requires `~` instead of `$HOME` in string literals — `$HOME` is not
             // expanded at parse time, so PATH entries would contain a literal "$HOME/...".
             let bin_path_ref_nu = bin_path_ref.replace("$HOME/", "~/");
-            ENV_TEMPLATE_NU.replace("__VP_BIN__", &bin_path_ref_nu)
+            let home_path_ref_nu =
+                vite_plus_home.as_path().display().to_string().replace('\\', "/");
```
naokihaba: Minor consistency note, not a blocker.

In the Nu implementation, __VP_BIN__ is rendered as home-relative by converting $HOME/... to ~/..., but __VP_HOME__ remains an absolute install-time path.

This should work fine for normal installs since env.nu is generated under that VP_HOME. I just noticed a small inconsistency where if a generated env.nu is reused after the home path changes, PATH would resolve through the new home while VP_HOME would still point to the old absolute path.

It might be worth applying the same home-relative logic to VP_HOME as well, but I don't think this needs to block the PR.

For context, Nushell can expand paths such as "~/.vite-plus" at runtime with the `path expand` command.
https://www.nushell.sh/commands/docs/path_expand.html

crates/vite_global_cli/src/commands/env/setup.rs
```diff
@@ -622,6 +624,7 @@ complete -c vpr --keep-order --exclusive --arguments "(__vpr_complete)"
 // Completions delegate to Fish dynamically (VP_COMPLETE=fish) because clap_complete_nushell
 // generates multiple rest params (e.g. for `vp install`), which Nushell does not support.
 const ENV_TEMPLATE_NU: &str = r#"# Vite+ environment setup (https://viteplus.dev)
+$env.VP_HOME = "__VP_HOME__"
```
naokihaba: Nit / robustness, not a blocker.

It looks like __VP_HOME__ is being inserted into a Nushell double-quoted string without any escaping. Since Nushell expects quotes and backslashes to be escaped in these strings, a custom VP_HOME that includes a `"` or some backslash sequences might end up breaking the env.nu file.

The same thing applies to __VP_BIN__ as well, so it might be a good idea to add a simple Nu string-literal renderer that both placeholders can use.

Nushell string docs:
https://www.nushell.sh/book/working_with_strings.html#double-quoted-strings

naokihaba (approved):

crates/vite_global_cli/src/commands/env/setup.rs
fengmk2: I think we should add a Nushell real test case at github action

nekomoyi (approved):

fengmk2:
@liangmiQwQ I'm done, back to your control.

fengmk2:
@liangmiQwQ I ran it once under Windows and haven't found any issues so far.

crates/vite_global_cli/src/commands/env/setup.rs
naokihaba: I will add this separately later on.

naokihaba: https://github.com/voidzero-dev/vite-plus/issues/2049

liangmiQwQ:
@fengmk2 @nekomoyi @naokihaba Thank the team for inspecting, reviewing and simplifying!

I'm a bit sleepy tonight, and I will take a look to know if there is anything need doing on my side tomorrow.

@fengmk2 And I left a comment in your pull request, you can take a look https://github.com/voidzero-dev/vite-plus/pull/1958#discussion_r3519794416

crates/vite_global_cli/src/commands/env/setup.rs
liangmiQwQ: That makes a lot of sense! I've adjusted it in 139248a. I am not familiar to Nushell and its script syntax so you can review to confirm if my fix in the right way 💚. @naokihaba

naokihaba: Thanks!!

This should fix the consistency issue with the home-relative `VP_HOME`.

I've updated `VP_HOME` to render as `~/... | path expand --no-symlink`. This way, it follows the same home-relative handling we use for `VP_BIN` but still resolves to an absolute path at runtime.

fengmk2:
GO!

PR #2059, opened 2026-07-06
fix(docs): add missing pnpm-workspace.yaml to Dockerfile

wan-kong:
For pnpm workspace projects, omitting `pnpm-workspace.yaml` causes `vp install --frozen-lockfile` to fail because pnpm cannot resolve `catalog:` in dependencies without it.


fengmk2 (approved):

fengmk2:
@wan-kong Thanks!

PR #2060, opened 2026-07-06
fix(static_config): only trust `defineConfig` from`vite-plus`

liangmiQwQ:
At present, static config only checks if the `export default` statement's callee is `defineConfig`.

So, if I import `defineConfig` from other sources and modify the input in the function, or define a variable called `defineConfig`, the static config analyze will get broken.

For example, if you are using a preset which automatically generates tasks, you will get:
```bash
$ vpr build
Task "build" not found.
Select a task (↑/↓, Enter to run, type to search): build

  No matching tasks. Press Escape to clear search.

```

If the `defineConfig` modify its import structure, you may get:

```bash
$ vpr dev
error: Failed to load task graph
* Failed to load task config file for package at "/Users/liangmi/code/liangmiQwQ/is-github-bot/"
* unknown field `missions`, expected one of `cache`, `tasks`, `enablePrePostScripts`
```

This PR adds source check and only trust `defineConfig` from `vite-plus` package.

This problem is discovered when I am doing attempts in my personal Vite+ config preset.

🤖 Generated with Codex


liangmiQwQ:
@codex review

liangmiQwQ:
@codex review

liangmiQwQ:
@codex

liangmiQwQ:
@codex

liangmiQwQ:
@codex review

fengmk2 (approved):
@liangmiQwQ That was indeed a problem, thank you for fixing it.

fengmk2:
@liangmiQwQ Could you add the error message to the PR description? This way, people can get a clearer understanding of the specific issue being fixed. Even if I know what the error is, others who are unfamiliar with it should be able to understand the problem at a glance.

liangmiQwQ:
@fengmk2 It may just be simple `task not found` for most cases. Or if the function's input type is changed, it will prompt an error. I've added both of them to PR description

fengmk2:
@liangmiQwQ I will run a round simplify before I merge.

PR #2069, opened 2026-07-06
test(snapshot): reproduce vp run ctrl-c tearing tasks down mid-shutdown

fengmk2:
Reproduces #2036 with a PTY snapshot test.

The stray characters after killing `vp run` in fish are the terminal's replies to fish's own prompt probes (fish 4.x sends OSC 11, CSI 6n, and DA1 at every prompt draw). The vp-side defect that turns them into garbage: on Ctrl+C, `vp run` tears the task down within a few milliseconds instead of waiting for its graceful shutdown, so the dying task tree (vite dev server, managed node) still touches the terminal while the shell is already drawing its next prompt. Plain `vite dev` is clean because the shell waits for vite itself, which finishes its terminal cleanup before exiting. Verified end to end by driving real fish 4.8 in a scripted PTY: `vp run dev` reproduces the exact junk from the issue, a plain `node` process killed the same way does not.

The new `run_ctrlc_teardown` fixture pins this: a dev-server-like `vpt` payload whose shutdown takes 300 ms records whether that shutdown was allowed to finish. The verdict is written by a watcher process in its own process group (it survives vp's teardown the same way real dev-server grandchildren do) and read by a follow-up `vpt wait-file` step, which keeps the snapshot deterministic under parallel suite load. The recorded snapshot currently says:

```
task was torn down before its graceful shutdown finished
```

A fix should flip this line to `task completed its graceful shutdown`.

Also adds `vpt wait-file <path> [timeout-ms]`, a generic helper that polls until a file exists and prints it.

PR #2070, opened 2026-07-06
fix(cli): wait for the task tree's graceful shutdown on ctrl-c

### LINKED ISSUE #2036: Odd characters after killing `vp run` with fish
### Describe the bug

When I kill a `vp run` task in [fish shell](https://fishshell.com/), the terminal shows a number of unexpected characters in the prompt:

```text
❯ vp run

Selected task: dev
$ vp dev ⊘ cache disabled

  VITE+ v0.2.2

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
^C⏎

fish-repro on  main via  v25.5.0 on ☁️  (ca-central-1) took 3s
❯ ^[]11;rgb:2828/2c2c/3434^[\^[[9;1R^[[?62;22;52c
```

These disappear if I hit return/enter to get a fresh prompt.

This does not happen with `zsh` or `bash`, and I can't reproduce it with vanilla `vite` (`vite run` or `vite dev`).

### Reproduction

https://github.com/johnfraney/vite-plus-fish-repro

### Steps to reproduce

1. Use `fish` as your shell
1. Execute `vp run`
1. Choose a task, like `dev`
1. Kill the process with ctrl+c
1. See the extra characters in the prompt

### System Info

```shell
vp env current

Environment:
  Version  24.18.0
  Source   lts

Tool Paths:
  node  /Users/john/.vite-plus/js_runtime/node/24.18.0/bin/node
  npm   /Users/john/.vite-plus/js_runtime/node/24.18.0/bin/npm
  npx   /Users/john/.vite-plus/js_runtime/node/24.18.0/bin/npx

Package Manager:
  Name          pnpm
  Version       11.9.0
  Source Path   /Users/john/Code/vp-fish-repro/fish-repro/package.json
  Project Root  /Users/john/Code/vp-fish-repro/fish-repro
  Bin Path      /Users/john/.vite-plus/package_manager/pnpm/11.9.0/pnpm/bin/pnpm



vp --version

vp v0.2.1

Local vite-plus:
  vite-plus  v0.2.2

Tools:
  vite             v8.1.2
  rolldown         v1.1.4
  vitest           v4.1.9
  oxfmt            v0.57.0
  oxlint           v1.72.0
  oxlint-tsgolint  v0.24.0
  tsdown           v0.22.3

Environment:
  Package manager  pnpm v11.9.0
  Node.js          v24.18.0
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

fengmk2:
Closes #2036

Builds on #2065 by @forehalo: this branch carries both of its commits as-is (the `execute_with_terminal_guard` integration with terminal-state restore, applied to both delegation call sites), with follow-up commits adjusting the signal handling on top.

Background: Ctrl+C is delivered by the terminal to the whole foreground process group, and the delegating global `vp` is the PTY session leader. It used to die of SIGINT within milliseconds, handing the terminal back to the shell (and SIGHUP'ing the session) while the task tree (vite dev server, managed node) was still shutting down. The dying tree's late terminal writes and mode restores then raced the shell's prompt probes (fish sends OSC 11, CSI 6n, and DA1 at every prompt draw), turning the terminal's replies into the garbage input from the issue.

Adjustments on top of #2065:

- Interrupt swallowing moves from the guard's `SIG_IGN` sigaction into the wait loop as a signal handler. `SIG_IGN` survives exec into the entire child tree, so any task without its own SIGINT handler would become uninterruptible (node launders inherited dispositions at startup, which masks this for the delegation path today, but the guard is shared code and the binding call site spawns other children). A handler resets to default on exec and leaves children untouched, and also works when stdin is not a terminal, where the guard never installs.
- A second Ctrl+C force-kills the child, so a task that ignores the first interrupt cannot make vp unstoppable. Verified against the `SIG_IGN` variant: a stuck task left vp hanging until externally killed.
- Children killed by a signal now exit with the conventional `128 + signal` code (130 for SIGINT, 137 for the force-kill) instead of a generic 1.
- Windows gets the same swallow-and-escalate loop via the console Ctrl+C handler.

The `run_ctrlc_teardown` snapshot from #2069 flips to `task completed its graceful shutdown` and now guards the waiting behavior; a second case pins the escalation (task ignoring the interrupt, force-killed by the second Ctrl+C, exit 137). Verified end to end by driving real fish 4.8 in a scripted PTY: the stray escape sequences no longer appear on the prompt after killing `vp run dev`.

Known residual edge (follow-up, vite-task side): a shell-wrapped script line whose intermediate shell dies of SIGINT can still orphan grandchildren, since vite_task waits only for its direct child; waiting on the task's process group would close that.


fengmk2:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/2070?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2070** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/2070?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/2070?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2069** <a href="https://app.graphite.com/github/pr/voidzero-dev/vite-plus/2069?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/voidzero-dev?ref=merge-queue-instructions-in-stack-comment&prId=2070).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

fengmk2:
Consolidated into #2065: the fix commits (including the adjustments from this PR) have been pushed onto that branch so the fix lands under the original contributor's PR.

PR #2072, opened 2026-07-07
docs: refresh root agent guide

jong-kyung:
## Summary

Refresh the root `AGENTS.md` guidance to match the current repository structure and test workflow.

- Remove stale references to the deleted `packages/test` package and document that `vite-plus/test*` is now generated
from upstream Vitest shims in `packages/cli`.
- Update the repository map with current crates and package ownership.
- Clarify that new CLI snapshot coverage should use the PTY-based runner introduced in #2052, while legacy snap-test
trees are migration-only.
- Update validation guidance and references for the current snapshot test commands.


fengmk2 (approved):
@jong-kyung Thanks

PR #2075, opened 2026-07-07
fix(static_config): trust defineConfig from vite

fengmk2:
## Summary

Trust `defineConfig` imported from both `vite-plus` and `vite` during static config extraction, while keeping custom helpers on the runtime fallback path.

Also adds PTY snapshot coverage and migrates the legacy custom `defineConfig` snapshot.

## Root cause

Commit `b9b15c45` only trusted `defineConfig` from `vite-plus`. Migrated projects commonly retain imports from `vite`, so configs without a `run` block became non-static and were evaluated at runtime.

This broke vinext during install because task graph loading evaluated an example Vite config that imported the workspace `vinext` package before its `dist` entry existed.

Failure: https://github.com/voidzero-dev/vite-plus/actions/runs/28774700335/job/85318252437


fengmk2:
@codex review

wan9chi (approved):

liangmiQwQ (approved):

liangmiQwQ:
> This broke vinext during install because task graph loading evaluated an example Vite config that imported the workspace `vinext` package before its `dist` entry existed.

I agree on this change, but I think the result of analyzing static config and really run config should be the same? I'm not very familiar with this part and is there any other problem?

fengmk2:
> > This broke vinext during install because task graph loading evaluated an example Vite config that imported the workspace `vinext` package before its `dist` entry existed.
> 
> I agree on this change, but I think the result of analyzing static config and really run config should be the same? I'm not very familiar with this part and is there any other problem?

There is a chicken and egg problem here: `vinext`'s `build` depends on `vp run build`. If static analysis is not performed, `vite.config.ts` will depend on the build dist output of `vp run build`, leading to a deadlock.

liangmiQwQ:
okay, understood, thanks for explanation 💯

PR #2095, opened 2026-07-08
fix(create): ignore dotenv files in new projects

### LINKED ISSUE #2086: Add environment files to `.gitignore`
### Description

Hi, would it be possible to add these files to the default .gitignore when creating a new project with Vite+?

```sh
.env
.env.local
.env.*.local
```

### Suggested solution

Add `.env` files to the default `.gitignore`

### Alternative

_No response_

### Additional context

_No response_

### Validations

forehalo:
## Summary

- Add dotenv environment files to the built-in monorepo template `.gitignore`.
- Extend create-time `.gitignore` defaults so generated projects also ignore `.env` and `.env.*` while keeping `!.env.example` trackable.
- Update focused unit coverage and the migrated PTY snapshot for generated `.gitignore` content.

Closes #2086

## Validation

- `git diff --check -- packages/cli/src/create/utils.ts packages/cli/src/create/bin.ts packages/cli/src/create/__tests__/utils.spec.ts packages/cli/templates/monorepo/_gitignore crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/create_org_bundled_monorepo/snapshots/create_org_bundled_monorepo.md`
- `/tmp/vp-issue-2086-test/node_modules/.bin/vitest run src/create/__tests__/utils.spec.ts --config <temp config>`

Note: the normal workspace install is still not available in this worktree because the checkout lacks `vite/patches/*` files referenced by `pnpm-workspace.yaml`, so the focused test used a temporary dependency install.

### DIFF
--- a/crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/create_org_bundled_monorepo/snapshots/create_org_bundled_monorepo.md
+++ b/crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/create_org_bundled_monorepo/snapshots/create_org_bundled_monorepo.md
@@ -66,4 +66,9 @@ node_modules excluded even though tarball shipped no .gitignore
 
 ```
 node_modules
+
+# dotenv environment variable files
+.env
+.env.*
+!.env.example
 ```
--- a/packages/cli/src/create/__tests__/utils.spec.ts
+++ b/packages/cli/src/create/__tests__/utils.spec.ts
@@ -6,7 +6,7 @@ import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 
 import {
   deriveDefaultPackageName,
-  ensureGitignoreNodeModules,
+  ensureDefaultGitignoreEntries,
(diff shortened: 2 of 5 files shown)


crates/vite_js_runtime/src/providers/node.rs
```diff
@@ -323,7 +323,48 @@ impl NodeProvider {
     /// Returns an error if no version is found or the version index cannot be fetched.
     pub async fn resolve_absolute_latest_version(&self) -> Result<Str, Error> {
         let versions = self.fetch_version_index().await?;
-        find_absolute_latest_version(&versions)
+        #[cfg(target_env = "musl")]
```
fengmk2: I skip flaky musl test at https://github.com/voidzero-dev/vite-plus/pull/2099

fengmk2: Every time Node.js releases a new version, the corresponding musl file is delayed in its release, so I skip testing it directly.

fengmk2 (approved):

PR #2098, opened 2026-07-08
fix(pack): guard shims banner in unbundle mode to prevent ESM imports leaking into CJS output

### LINKED ISSUE #2090: pack: shims + unbundle emits raw ESM imports into .cjs, producing invalid CommonJS
### Describe the bug

With `pack.shims: true` **and** `pack.unbundle: true`, the generated CommonJS output (`.cjs`) contains bare ESM statements at the top of the file:

```js
import "node:path";
import "node:url";
import.meta.url;
```

This is not valid CommonJS, so **`require()`-ing the built `.cjs` throws at module load, before any code runs**:

```
SyntaxError: Cannot use import statement outside a module
```

The shim preamble is emitted verbatim into *both* the `.mjs` and the `.cjs`. In the ESM output those statements are legal; in the CommonJS output they are not, and they are neither transformed to their CJS equivalents nor omitted. `vp pack` reports success (exit 0) and writes the broken artifact.

The trigger is the combination of the two flags. With `shims: true` and `unbundle` left at its default, `import.meta.url` / `node:` builtins are correctly compiled to their CJS forms (`require("url").pathToFileURL(__filename).href`, `require("node:path")`); it is only `shims: true` + `unbundle: true` together that leaks the raw ESM preamble.

Expected: the CommonJS build must be valid CommonJS — the shim preamble should be compiled to CJS-safe equivalents (or omitted) in the `.cjs`, exactly as it is when `unbundle` is off.

### Reproduction

https://github.com/TheHolyWaffle/vite-plus-cjs-shim-leak

### Steps to reproduce

```sh
npm install
npx vp pack
# build succeeds (exit 0)

# require the CommonJS output the way a CJS consumer would:
node -e "require('./dist/index.cjs')"
# -> SyntaxError: Cannot use import statement outside a module

# the ESM output is fine:
node --input-type=module -e "import('./dist/index.mjs').then(m => console.log(m.under('/tmp')))"
# -> /tmp/sub
```

Generated `dist/index.cjs` (note lines 1-3):

```js
import "node:path";
import "node:url";
import.meta.url;
Object.defineProperty(exports, Symbol.toStringTag, { value: "Module" });
let node_path = require("node:path");
//#region src/index.ts
function under(dir) {
	return (0, node_path.resolve)(dir, "sub");
}
//#endregion
exports.under = under;
```

Removing `unbundle: true` (keeping `shims: true`) produces valid CommonJS
(`let node_path = require("node:path")`, no leading `import`), which loads without error.

### System Info

```shell
# `vp env current` is not a command in 0.2.3 (errors "Command 'env' not found"),
# so pasting `vp --version` (which includes the tool matrix) plus runtime versions:

$ npx vp --version
vp v0.2.3

Local vite-plus:
  vite-plus  v0.2.3

Tools:
  vite             v8.1.3
  rolldown         v1.1.4
  vitest           v4.1.9
  oxfmt            v0.57.0
  oxlint           v1.72.0
  oxlint-tsgolint  v0.24.0
  tsdown           v0.22.3

$ node --version
v24.16.0

$ npm --version
11.13.0

# OS: macOS (darwin 25.5.0)
```

### Used Package Manager

npm

### Logs

```shell
$ npx vp pack
ℹ entry: src/index.ts
ℹ Build start
ℹ [ESM] dist/index.mjs  ...
✔ Build complete
ℹ [CJS] dist/index.cjs  ...
✔ Build complete
# exit code: 0 — the invalid .cjs is written without any warning

$ node -e "require('./dist/index.cjs')"
.../dist/index.cjs:1
import "node:path";
^^^^^^
SyntaxError: Cannot use import statement outside a module
    at wrapSafe (node:internal/modules/cjs/loader:1691:18)
    at Module._compile (node:internal/modules/cjs/loader:1734:20)
    ...
```

### Validations

AmariahAK:
### What was the issue

When `pack.shims: true` and `pack.unbundle: true` are combined, the generated `.cjs` file contains raw ESM statements at the top:

```js
import "node:path";
import "node:url";
import.meta.url;
```

This is not valid CommonJS — `require()`-ing the built `.cjs` throws `SyntaxError: Cannot use import statement outside a module` at module load. The `.mjs` output is fine; only the CJS artifact is broken.

### Where was the issue

In tsdown's `resolveInputOptions()`, the unbundle code path pushes the `shimsPlugin` (which injects an ESM banner) unconditionally — without checking the output format. The bundle-mode path (`getShimsInject`) already has a `format === 'es' && platform === 'node'` guard, but the unbundle path never did.

The bug is upstream: [rolldown/tsdown#991](https://github.com/rolldown/tsdown/pull/991). The fix was merged but not included in the published `tsdown@0.22.4` release.

### How I fixed it

Added a build-time patch in `bundleTsdown()` inside `packages/core/build.ts`. It runs before the tsdown re-bundle and injects the same format guard that `getShimsInject` already uses:

```diff
 if (shims && !cjsDts) if (unbundle) {
+    if (format === 'es' && platform === 'node') {
         define = { ...define, ...shimsDefine };
         plugins.push(shimsPlugin);
+    }
 } else inject = getShimsInject(format, platform);
```

Also bumped the tsdown catalog from `^0.22.3` to `^0.22.4` (with `@tsdown/css` and `@tsdown/exe`), and updated `bundledVersions.tsdown` to `0.22.4`.

### Why this approach

- **Reuse-first:** the patch follows the same pattern as `brandTsdown()`, which already applies string replacements to the bundled tsdown output.
- **Minimal:** 35 lines, build-time only, zero runtime overhead.
- **Reversible:** marked with a comment referencing the upstream PR and a note to re-evaluate after `tsdown >=0.22.5`.
- **Alternative considered:** waiting for the upstream release — but the fix is one line and the build-time patch is the established mechanism in this codebase.

### How to test locally

1. Initialize submodules:
   ```bash
   git submodule update --init --recursive
   ```

2. Rebuild the core package (the patch runs here):
   ```bash
   pnpm -C packages/core build
   ```
   Expected output: `Patched shims+unbundle CJS leak in .../build-*.mjs`

3. Create a minimal test project with this config:
   ```ts
   // vite.config.ts
   export default defineConfig({
     pack: {
       entry: 'src/index.ts',
       shims: true,
       unbundle: true,
       format: ['esm', 'cjs'],
     },
   });
   ```

4. Build and verify:
   ```bash
   vp pack
   node -e "require('./dist/index.cjs')"  # should NOT throw
   node --check dist/index.cjs             # should pass
   ```

5. Run the full test suite:
   ```bash
   pnpm test
   ```

---

Closes #2090

### DIFF
--- a/packages/core/build.ts
+++ b/packages/core/build.ts
@@ -423,6 +423,41 @@ async function bundleTsdown() {
 
   const thirdPartyCjsModules = new Set<string>();
 
+  // Patch shims+unbundle ESM banner leak into CJS output.
+  // Upstream fix: rolldown/tsdown#991 (merged, not yet in a published release).
+  // The shimsPlugin banner is pure ESM syntax; when unbundle is on it was
+  // pushed regardless of format, leaking `import ...` into .cjs files.
+  // Add the same format guard that getShimsInject already uses.
+  // Re-evaluate after tsdown >=0.22.5.
+  {
+    const buildChunkFiles = await glob(
+      toPosixPath(join(tsdownSourceDir, 'dist', 'build-*.mjs')),
+      { absolute: true },
+    );
+    for (const file of buildChunkFiles) {
+      let source = await readFile(file, 'utf-8');
+      if (source.includes('if (shims && !cjsDts) if (unbundle) {')) {
+        const NL = String.fromCharCode(10);
+        const TAB = String.fromCharCode(9);
+        source = source.replace(
+          'if (shims && !cjsDts) if (unbundle) {',
+          'if (shims && !cjsDts) if (unbundle) {' +
+            NL +
+            TAB +
+            TAB +
+            "if (format === 'es' && platform === 'node') {",
+        );
+        source = source.replace(
+          TAB + '} else inject = getShimsInject(format, platform)',
+          TAB + '}' + NL + TAB + '} else inject = getShimsInject(format, platform)',
+        );
+        await writeFile(file, source, 'utf-8');
+        console.log(`Patched shims+unbundle CJS leak in ${file}`);
+        break;
+      }
+    }
+  }
+
   // Re-build tsdown cli plus the bundled `@tsdown/exe` and `@tsdown/css`
   // extensions as stable named entries (`tsdown-exe.js`, `tsdown-css.js`).
   await build({
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -214,6 +214,6 @@
   "bundledVersions": {
     "vite": "8.1.3",
     "rolldown": "1.1.4",
-    "tsdown": "0.22.3"
+    "tsdown": "0.22.4"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,21 +6,9 @@ settings:
 
 catalogs:
   default:
-    '@babel/core':
-      specifier: ^7.29.7
-      version: 7.29.7
-    '@babel/preset-env':
-      specifier: ^7.29.7
-      version: 7.29.7
-    '@babel/preset-typescript':
-      specifier: ^7.29.7
-      version: 7.29.7
     '@napi-rs/cli':
       specifier: ^3.7.2
       version: 3.7.2
-    '@napi-rs/wasm-runtime':
-      specifier: ^1.1.6
-      version: 1.1.6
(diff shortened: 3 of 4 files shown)


fengmk2:
We should wait for tsdown fix it then upgrade to the fix version.

PR #2115, opened 2026-07-09
fix(snapshots): scope registry teardown signal

fengmk2:
The local-registry teardown sent SIGTERM to both the server process and
its process group (kill -TERM <pid> -<pid>). On the GitHub-hosted
ubuntu-latest PTY runs that process-group signal leaked past the case
boundary and reached the Actions runner agent, which reported "runner
received a shutdown signal" and cancelled the job. Confirmed via a
bpftrace probe on signal:signal_generate: no suite process ever sends a
SIGINT, so the runner itself was doing the cancelling in response to the
stray group SIGTERM. Scope the SIGTERM to the direct child; the server's
own handler still removes its throwaway caches and the force-kill fallback
still guarantees teardown.

PR #2128, opened 2026-07-10
fix(env): clarify the use case of `vp-use.cmd`

liangmiQwQ:
`vp env setup` currently creates `vp-use.cmd` on every platform even though the wrapper is only usable from Windows Command Prompt. Its writer also runs before `VP_HOME/bin` is created, so setup skips the wrapper when starting from a fresh home. The CMD-specific version-switching command is also missing from the public environment guide.

This PR creates `vp-use.cmd` alongside the Windows shims after the bin directory is initialized, adds regression coverage for fresh Windows setup and Unix exclusion, and documents how to use `vp-use` from Command Prompt.

🤖 Generated with Codex

### DIFF
--- a/crates/vite_global_cli/src/commands/env/setup.rs
+++ b/crates/vite_global_cli/src/commands/env/setup.rs
@@ -80,6 +80,9 @@ pub async fn execute(refresh: bool, env_only: bool) -> Result<ExitStatus, Error>
     // Ensure bin directory exists
     tokio::fs::create_dir_all(&bin_dir).await?;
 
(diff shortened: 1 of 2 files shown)


liangmiQwQ:
@codex review

liangmiQwQ:
@codex review

fengmk2:
ci hang issue will be fixed in https://github.com/voidzero-dev/vite-plus/pull/2115

liangmiQwQ:
Okay, updated

fengmk2 (approved):

PR #2130, opened 2026-07-10
fix(cli): avoid debug formatting in user output

liangmiQwQ:
Some user-facing Vite+ diagnostics rely on Rust's `Debug` formatting. This can expose implementation details such as `Some(1)` for a failed process status and wrapper types around paths.

This PR uses display formatting for Vite+-owned paths and process statuses, while keeping explicit quotes in package-manager diagnostics and preserving developer-only debug traces. Failed global installs now report `exit status: 1`.

There is still follow-up work in Vite Task. In particular, `vite_workspace::Error::PackageJsonNotFound` currently uses `{0:?}` and can render paths as `AbsolutePathBuf("/path")`. That formatter and the sibling user-facing `Debug` formatters in `vite_workspace` should be fixed upstream, followed by updating Vite+'s pinned Vite Task revision.

🤖 Generated with Codex


liangmiQwQ:
@codex review

liangmiQwQ:
@codex review

fengmk2 (approved):

Issue #2116, opened 2026-07-09
Tracking: migrate legacy snap tests to the PTY snapshot suite


fengmk2:
The legacy snap-test suites (`packages/cli/snap-tests`, `packages/cli/snap-tests-global`) are being migrated to the PTY snapshot suite under `crates/vite_cli_snapshots`. This issue tracks the remaining categories. Related: #2078 (closed, migration problems).

## Status

492 legacy cases on `main`: all migrated or in-flight. #2143 empties both legacy trees and teaches the legacy runner to exit cleanly when a tree is gone from a fresh checkout (git does not track empty dirs).

#2143 also removes the `cli-snap-test` CI jobs, `.github/scripts/retry-failed-snap-tests.sh`, and the trees' Renovate ignorePaths. #2145 removes the parked `snap-tests-todo` tree (three cases no runner referenced). Follow-up after the stack merges: delete the legacy runner (`packages/tools/src/snap-test.ts`) and the `migrate-snap-tests` migrator.

Done / in-flight:
- [x] `migration-*` (149) — #2107
- [x] `shim-*` (11) — #2109
- [x] `check-*` (34: 30 migrated + 4 `check-pass*` already covered by existing fixtures) — #2117
- [x] `vite-*` (11) — #2118
- [x] `npm-*` local (1, `npm-install-with-options`) — #2119
- [x] `vp-*` (10) — #2120
- [x] tail (19 of 23 misc single/pair categories: workspace, synthetic, plain, yarn, task, run, exit, change, ignore_dist, fallback, dev, runtime, global, delegate) — #2121
- [x] `lint-*` / `oxlint-*` (6) — #2131
- [x] `test-*` / `fmt-*` / `cache-*` (11) — #2132
- [x] `npm-global-*` (9 of 10) — #2134, unblocked by #2110 (allow-list removal); adds `vpt stat-file` symlink state and `$VP_HOME` arg expansion
- [x] `env-*` (5) — #2137
- [x] `command-*` toolchain (~112: cache, check, config, dev, dlx, doc, env, exec, fmt, helper, info, init, lint, nested, node, pack, pm, preview, rebuild, run, staged, upgrade, version, vp-alias, vpr, vpx) — #2138
- [x] `command-*` package-manager passthrough (~108: add, dedupe, install, link, list, outdated, owner, prune, publish, remove, unlink, update, view, why across npm/pnpm/yarn/bun) — #2141
[...]

PR #2135, opened 2026-07-10
chore: exclude PTY snapshot configs from Vitest discovery

jong-kyung:
Summary

- Exclude the PTY CLI snapshot fixtures from Vitest config discovery in VS Code.
- Prevent vite_plugins_async_test/.vitest-plugin-loaded from being generated automatically.


.vscode/settings.json
```diff
@@ -18,5 +18,5 @@
   "typescript.reportStyleChecksAsWarnings": false,
   "typescript.updateImportsOnFileMove.enabled": "always",
   "typescript.experimental.useTsgo": true,
-  "vitest.configSearchPatternExclude": "{**/node_modules/**,**/vendor/**,**/.*/**,**/*.d.ts,{**/packages/cli/snap-tests*/**}}"
+  "vitest.configSearchPatternExclude": "{**/node_modules/**,**/vendor/**,**/.*/**,**/*.d.ts,{**/packages/cli/snap-tests*/**,**/crates/vite_cli_snapshots/tests/cli_snapshots/**}}"
```
liangmiQwQ, line 21: Should we remove the previous `**/packages/cli/snap-tests*/**` added in #1723? I guess the legacy snap tests will be completely migrated soon :)

jong-kyung: I’d prefer to keep it until the migration(#2116) is complete. Wouldn't that be better?

jong-kyung: If this PR lands after that migration is complete, removing the old pattern here makes sense too 😁

fengmk2 (approved):

.vscode/settings.json
liangmiQwQ: No problem!

liangmiQwQ (approved):
💯

.vscode/settings.json
jong-kyung: I've opened issue #2116 to remove legacy code 🙇‍♂️

PR #2147, opened 2026-07-10
feat: support TypeScript 7 (tsc 7)

### LINKED ISSUE #2148: feat: support TypeScript 7
### Describe the feature request

With the release of TypeScript 7, projects utilizing `vite-plus` and scaffolded by `vp create` want to upgrade their TypeScript version to 7. Currently, `vite-plus` specifies `"typescript": "^5.0.0 || ^6.0.0"` in its core packages peer dependencies, causing dependency resolution errors/warnings during installation.

### Proposed Solution

1. Update the `typescript` peer dependency in `@voidzero-dev/vite-plus-core` (built dynamically under `packages/core/build.ts`) to `"^5.0.0 || ^6.0.0 || ^7.0.0"`.
2. Update CLI scaffolding monorepo templates under `packages/cli/templates/monorepo/` (`pnpm-workspace.yaml` and `_yarnrc.yml`) to support `^5 || ^6 || ^7`.
3. Keep the development devDependencies of the `vite-plus` workspace itself pinned to TypeScript 6 for now, as programmatic compiler APIs in upstream libraries (like `rolldown` and `tsdown`) are not yet compatible with the TypeScript 7 API structure.

rorychatt:
### Description

This PR adds support for TypeScript 7 (`tsc 7`) to `vite-plus`.

Currently, `vite-plus` only lists `^5.0.0 || ^6.0.0` in its peer dependencies, which causes installation warnings/errors in projects that upgraded to TypeScript 7.

Closes #2148

### Changes

- **Core Package (`packages/core/build.ts`)**: Updated the `mergePackageJson()` build step to append `|| ^7.0.0` to the typescript peer dependency of `@voidzero-dev/vite-plus-core`.
- **Monorepo Templates (`packages/cli/templates/monorepo/`)**: Updated the generated `pnpm-workspace.yaml` and `_yarnrc.yml` templates to support `^5 || ^6 || ^7`.
- **Lockfile (`pnpm-lock.yaml`)**: Regenerated the workspace lockfile after resolving dependencies.

*Note: The workspace devDependencies of the `vite-plus` codebase itself remain pinned to TypeScript `^6.0.0` for development, as upstream packages (like `rolldown` and `tsdown`) currently rely on programmatic compiler APIs that are not yet compatible with the TypeScript 7 API.*

### Verification

All 870 unit/integration tests in `vite-plus` compile and pass successfully under the updated configurations.

### DIFF
--- a/crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/new_vite_monorepo/snapshots/new_vite_monorepo.md
+++ b/crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/new_vite_monorepo/snapshots/new_vite_monorepo.md
@@ -89,7 +89,7 @@ catalogMode: prefer
 
 catalog:
   "@types/node": ^24
-  typescript: ^5
+  typescript: ^5 || ^6 || ^7
   vite: npm:@voidzero-dev/vite-plus-core@<version>
   vite-plus: <version>
 overrides:
--- a/packages/cli/templates/monorepo/_yarnrc.yml
+++ b/packages/cli/templates/monorepo/_yarnrc.yml
@@ -1,4 +1,4 @@
 nodeLinker: node-modules
 catalog:
   '@types/node': ^24
-  typescript: ^5
+  typescript: ^5 || ^6 || ^7
--- a/packages/cli/templates/monorepo/pnpm-workspace.yaml
+++ b/packages/cli/templates/monorepo/pnpm-workspace.yaml
@@ -7,4 +7,4 @@ catalogMode: prefer
 
 catalog:
   '@types/node': ^24
-  typescript: ^5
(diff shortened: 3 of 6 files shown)


wan9chi:
@rorychatt Thank you for reporting and fixing the issue. While reviewing it I found the peer range turns out to be only part of the problem: TypeScript 7 removed the old compiler API that our bundled tsdown used for `vp pack` declaration generation (the error in #2160). Widening the range alone would let new projects install TS7 while their builds still fail.

Both parts are now fixed in #2168: the newer tsdown allows TS7 in its peer range and generates declarations with the TS7 native compiler. So we'll close this PR in favor of that one. Thanks again for raising the issue and getting the work started. The fix will be in the next release.

Issue #2041, opened 2026-07-04
Bundled bun.native binary crashes with SIGILL on CPUs without AVX2


mrboxs:
### Describe the bug

`vp create --package-manager bun` silently fails (exit code 1) because the bundled `bun.native` binary shipped with `vp` crashes with **SIGILL (signal 4, exit code 132)** on CPUs that lack AVX2 support.

The system's own bun binary (same version 1.3.14) works fine.

---

### Reproduction

Steps to reproduce:
1. Run on a CPU **without AVX2** (e.g. Intel Ivy Bridge i5-3470S)
2. Execute `vp create @tanstack/start --package-manager bun -- myapp --yes`
3. Observe: no project created, exit code 1, no visible output from template

Direct test of the bundled binary:
```
$ timeout 5 ~/.vite-plus/package_manager/bun/1.3.14/bun/bin/bun.native --version
timeout: the monitored command dumped core
exit: 132
```

The system bun works:
```
$ /home/mrboxs/.bun/bin/bun --version
1.3.14
```

---

### Steps to reproduce

1. Install vp via `curl -fsSL https://vite.plus | bash`
2. Run `vp create @tanstack/start --package-manager bun -- myapp --yes`
3. No project created, exit code 1

---

### System Info

```
$ vp env current
Environment:
  Version  24.18.0
  Source   lts

Tool Paths:
  node  /home/mrboxs/.vite-plus/js_runtime/node/24.18.0/bin/node
  npm   /home/mrboxs/.vite-plus/js_runtime/node/24.18.0/bin/npm
  npx   /home/mrboxs/.vite-plus/js_runtime/node/24.18.0/bin/npx

$ vp --version
vp v0.2.2
```

- **OS**: Arch Linux, Linux 7.0.10-arch1-1 x86_64
- **CPU**: Intel Core i5-3470S @ 2.90GHz (SSE4.2 + AVX, **no AVX2**)

---

### Used Package Manager

bun

---

### Logs

Bundled bun binary MD5: `7d0efc01d1124bb79ca444cb9cb404af` (92MB)
System bun binary MD5: `0233916fabdff93d1da22d583555fa76`

The bundled binary appears to be compiled with AVX2 instructions not supported on Ivy Bridge and older CPUs.

---

### Workaround

```sh
mv ~/.vite-plus/package_manager/bun/1.3.14/bun/bin/bun.native{,.bak}
[...]

PR #2154, opened 2026-07-12
fix: preserve signal exit code on Unix

liangmiQwQ:
In the current codebase, there are a lot of places where we simply use `result.status.code().unwrap_or(1)` to handle the exit code.

However, according to [Rust std docs](https://doc.rust-lang.org/std/process/struct.ExitStatus.html), if the process is terminated by a signal, the `status.code()` will become `None` on Unix. We should read `ExitStatus` to know the signal and get the real exit code.

This PR fixes them.

Partially related to #2041.

### DIFF
--- a/crates/vite_cli_snapshots/src/bin/vpt/json_edit.rs
+++ b/crates/vite_cli_snapshots/src/bin/vpt/json_edit.rs
@@ -29,6 +29,9 @@ pub fn run(args: &[String]) -> Result<(), Box<dyn std::error::Error>> {
         .ok_or_else(|| std::format!("parent of '{last}' is not an object"))?
         .insert((*last).to_owned(), new_value);
 
+    // Keep output stable when another dependency enables serde_json's
+    // workspace-wide `preserve_order` feature.
+    root.sort_all_objects();
     let mut out = serde_json::to_string_pretty(&root)?;
     out.push('\n');
     std::fs::write(path, out)?;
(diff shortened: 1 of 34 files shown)


liangmiQwQ:
@codex review

crates/vite_cli_snapshots/src/bin/vpt/json_edit.rs
```diff
@@ -29,6 +29,9 @@ pub fn run(args: &[String]) -> Result<(), Box<dyn std::error::Error>> {
         .ok_or_else(|| std::format!("parent of '{last}' is not an object"))?
         .insert((*last).to_owned(), new_value);
 
+    // Keep output stable when another dependency enables serde_json's
+    // workspace-wide `preserve_order` feature.
+    root.sort_all_objects();
```
liangmiQwQ: This is because this PR adds `vite_shared` crate as a dependency. And `vite_shared` enables `serde_json/preserve_order` feature, while Cargo is merging this change, json object will no longer be sorted, causes unrelated snapshots to differ.

fengmk2 (approved):
@liangmiQwQ please fix conflicts first.

fengmk2:
> @liangmiQwQ please fix conflicts first.

Let me fix conflicts

liangmiQwQ:
@fengmk2 thank you!

PR #2190, opened 2026-07-16
fix release npm compatibility

wan9chi:
## Why

The [v0.2.5 release run](https://github.com/voidzero-dev/vite-plus/actions/runs/29493170643) failed before publishing because `npm@latest` resolved to npm 12.0.1. npm 12.0.1 requires Node 22.22.2 or newer, while the workflow used Node 22.18.0.

## Changes

- run the publishing job on Node 24.18.0, which bundles npm 11.16.0
- remove the separate global npm installation
- add `workflow_dispatch` so the corrected v0.2.5 release can be started manually

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'packages/cli/package.json'
+  workflow_dispatch:
 
 permissions: {}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'packages/cli/package.json'
+  workflow_dispatch:
```
wan9chi, line 8: `workflow_dispatch` lets us retry an unpublished release (#2180) after a workflow-only fix.

wan9chi (commented):
Context on the manual release trigger.

cpojer (changes requested):
See comment by the clanker.

cpojer (approved):
Alright, but please remove that again after the release is published.

PR #2196, opened 2026-07-17
refactor(cli): reuse shared exit status helper

jong-kyung:
## Summary
All global CLI command paths now use the same platform-specific exit-status conversion, preventing Unix and Windows handling from drifting between modules.


wan9chi (approved):
👍

PR #2197, opened 2026-07-17
test(cli): share package manager fixtures

jong-kyung:
## Summary
Package-manager command tests now share one source of truth for fixture defaults, preventing setup drift across 32 command modules. Version-specific tests retain explicit versions, while tests that do not depend on a version use the common default.


wan9chi (approved):
Thank you!

jong-kyung:
@wan9chi Thanks for the review 🫡

PR #2226, opened 2026-07-23
fix(create): honor explicit package manager outside monorepos

### LINKED ISSUE #2208: `vp create --package-manager` is silently overridden by package manager detection from ancestor directories
### Describe the bug

`vp create --help` documents this:

```
--package-manager NAME  Use specified package manager (pnpm, npm, yarn, bun)
```

But on the `vite:application` path the flag doesn't do anything. I run `vp create` with `--package-manager pnpm` and get an npm project: the scaffolder is delegated through `npx create-vite`, deps are installed with npm, and the result has a `package-lock.json` plus `devEngines.packageManager.name: "npm"`. No warning that the flag was dropped, it just vanishes.

The annoying part is that it sticks. Every later `vp` command reads `devEngines` first, so the project stays npm forever unless you convert it by hand.

The pnpm path itself works fine. If workspace-root *detection* resolves to pnpm (run `vp create` from a directory whose package.json says `"packageManager": "pnpm@..."`), you get a fully pnpm-native project: `pnpm dlx` delegation, `pnpm-lock.yaml`, `pnpm-workspace.yaml` with `catalog:` aliases, pnpm in `devEngines`. So it looks like the flag just isn't wired into that decision.

Reproduced on v0.1.24 and v0.2.5. Feels related to #1914 (warn instead of silently ignoring pm flags), except this one is `vp create`'s own documented option rather than a pass-through.

### Reproduction

https://github.com/FallDownTheSystem/vp-create-package-manager-repro

### Steps to reproduce

From an empty directory with no package manager markers in any parent:

```sh
vp create vite:application --directory app --no-interactive --no-git --no-hooks --no-editor --no-agent --package-manager pnpm --verbose -- --template vue-ts
```

Then look at `app/`: `package-lock.json`, and `devEngines.packageManager.name` is `"npm"`.

### System Info

```shell
vp v0.2.5

Local vite-plus:
  vite-plus  Not found

Tools:
  vite             Not found
  rolldown         Not found
  vitest           Not found
  oxfmt            Not found
  oxlint           Not found
  oxlint-tsgolint  Not found
  tsdown           Not found

Environment:
  Package manager  npm latest
  Node.js          v24.18.0

Environment:
  Version  24.18.0
  Source   lts

Tool Paths:
  node  C:\Users\Juugo\.vite-plus\js_runtime\node\24.18.0\node.exe
  npm   C:\Users\Juugo\.vite-plus\js_runtime\node\24.18.0\npm.cmd
  npx   C:\Users\Juugo\.vite-plus\js_runtime\node\24.18.0\npx.cmd
```

(run from an empty directory, hence the "Not found" rows)

### Used Package Manager

pnpm

### Logs

```shell
Running: npx create-vite@latest app --template vue-ts --no-interactive --no-immediate --no-rolldown
...
• Node 24.18.0  npm 12.0.1
```

Full verbose log is in the repro repo's [`output.txt`](https://github.com/FallDownTheSystem/vp-create-package-manager-repro/blob/master/output.txt).

### Validations

jong-kyung:
## Summary

 Preserve the existing package manager inside monorepos while allowing an explicit `vp create --package-manager`
option to override package manager signals inherited from non-monorepo ancestor directories. When the selected manager differs from the detected one, resolve its latest stable version instead of reusing the ancestor manager’s version. Add snapshot coverage for creating a pnpm project beneath an npm ancestor and update the package manager detection RFC.

 Fixes #2208


jong-kyung:
@codex review

jong-kyung:
@codex review

fengmk2 (approved):

PR #2227, opened 2026-07-23
fix(install): verify modern Yarn hashes against the CLI binary

### LINKED ISSUE #2209: `vp install` rejects a Corepack-written `packageManager` pin: vp hashes the .tgz, Corepack hashes the extracted binary
### Describe the bug

`vp install` rejects a valid Corepack-written `packageManager` pin with a `Hash mismatch` error, because vp and Corepack hash **different artifacts** for the same package manager release.

```
error: Hash mismatch: expected sha512.ccbfabf7…4697, got sha512.dad89242…a9ea
```

- `expected` is verbatim the `packageManager` pin in `package.json`, written by `corepack use yarn@…`.
- `got` is the sha512 of the **`.tgz` tarball** — it equals `@yarnpkg/cli-dist@4.17.1`'s npm `dist.integrity` exactly.
- Corepack's hash is the sha512 of the **extracted `yarn.js`**, not the tarball.

Verified on yarn 4.17.1:

```
sha512(~/.cache/node/corepack/v1/yarn/4.17.1/yarn.js) = ccbfabf7…4697   ← Corepack's basis, == the pin
sha512(cli-dist-4.17.1.tgz)                           = dad89242…a9ea   ← vp's basis, == npm dist.integrity
```

The payloads are identical — the cached `yarn.js` is byte-for-byte the file inside the published tarball (both sha256 `471ffb15…`, 3 019 751 bytes). Nothing is tampered with; only the hash basis differs.

This is symmetric, so the two tools are mutually exclusive: if I rewrite the pin to vp's tarball hash, `vp install` passes and a cold-cache `corepack install` fails with the mirror-image error:

```
Internal Error: Mismatch hashes. Expected dad89242…a9ea, got ccbfabf7…4697
    at installVersion (…/corepack/dist/lib/corepack.cjs:13262:11)
```

No hashed value satisfies both, which is a problem for any repo whose Docker images run `corepack install` while developers run `vp install`. Since Corepack is what writes the `+sha512.` suffix, and #195 describes the feature as "Support Corepack-style `packageManager` format with integrity hashes", vp's basis looks like the unintended one.

The source appears to hash the tarball before extraction, in [`crates/vite_install/src/request.rs`](https://github.com/voidzero-dev/vite-plus/blob/main/crates/vite_install/src/request.rs):

```rust
client.download_file(url, &tgz_file).await?;

// Verify hash if provided
if let Some(expected_hash) = expected_hash {
    verify_file_hash(&tgz_file, expected_hash).await?;
}
```

**Only reproduces on a cold package-manager cache.** Once `~/.vite-plus/package_manager/yarn/<version>/` exists, `vp install` succeeds with the very same pin it rejected a moment earlier, so verification appears to be skipped on a cache hit. That contradicts #195's "Re-verify cached packages when hash is provided", and it makes the bug look intermittent — CI and fresh containers fail while a warm dev machine passes.

**Expected behavior:** verify the `packageManager` integrity against the same artifact Corepack hashes (the extracted binary), so a pin written by `corepack use` validates under `vp install`. If hashing the tarball is deliberate, it would help to say so in the error, since the current message reads as a corrupted download.

**Workaround:** drop the `+sha512.…` suffix and pin `"packageManager": "yarn@4.17.1"`, which both tools accept — at the cost of the integrity check.

### Reproduction

No repository needed — the repro is a `package.json` with two fields, given in full below.

### Steps to reproduce

```bash
mkdir vp-hash-repro && cd vp-hash-repro
cat > package.json <<'EOF'
{
  "name": "vp-hash-repro",
  "private": true,
  "packageManager": "yarn@4.17.1+sha512.ccbfabf7d7b6b32075088be9386fb9a2e00bb6887ef07fa56effabc890a56d53da1ccc4128d62db245fcbd3961b236d75335bdf7d5320ed6eafb7588b7ad4697"
}
EOF

# the pin above is what `corepack use yarn@4.17.1` writes; Corepack accepts it:
corepack install     # → "Adding yarn@4.17.1 to the cache..."
yarn --version       # → 4.17.1

# ensure a cold vp package-manager cache, or the check is skipped:
rm -rf ~/.vite-plus/package_manager/yarn/4.17.1 ~/.vite-plus/package_manager/yarn/4.17.1.lock

vp install           # → error: Hash mismatch: expected sha512.ccbfabf7…, got sha512.dad89242…
```

To confirm the two bases directly:

```bash
shasum -a 512 ~/.cache/node/corepack/v1/yarn/4.17.1/yarn.js   # → ccbfabf7…  (Corepack / the pin)
curl -sL https://registry.npmjs.org/@yarnpkg/cli-dist/-/cli-dist-4.17.1.tgz | shasum -a 512
                                                              # → dad89242…  (vp / npm dist.integrity)
```

### System Info

```shell
$ vp env current
Environment:
  Version       26.5.0
  Source        devEngines.runtime
  Source Path   /…/package.json
  Project Root  /…

Tool Paths:
  node  ~/.vite-plus/js_runtime/node/26.5.0/bin/node
  npm   ~/.vite-plus/js_runtime/node/26.5.0/bin/npm
  npx   ~/.vite-plus/js_runtime/node/26.5.0/bin/npx

Package Manager:
  Name          yarn
  Version       4.17.1
  Source        packageManager
  Source Path   /…/package.json
  Project Root  /…
  Bin Path      ~/.vite-plus/package_manager/yarn/4.17.1/yarn/bin/yarn

$ vp --version
vp v0.2.2

Local vite-plus:
  vite-plus  v0.2.0

Corepack 0.35.0 · Node 26.5.0 · macOS 26.5.2 (arm64)
Also reproduced on Linux aarch64 with vp v0.2.1 / yarn 4.14.1 (pin `sha512.64df4480…`, vp computed `sha512.eb6b9844…` = that release's npm `dist.integrity`), so it is neither platform- nor version-specific.
```

### Used Package Manager

yarn

### Validations

- [x] Check that there isn't already an issue for the same bug.
- [x] The provided reproduction is a minimal reproducible example.

leslieeilsel:
Corepack pins Yarn 2+ from the extracted CLI (`bin/yarn.js`), while Vite+ hashed the npm tarball. A pin written by `corepack use` therefore failed `vp install` on a cold cache, passed on a warm one, and made every `vp run` re-download Yarn.

- verify Yarn 2+ pins against `bin/yarn.js`; npm, pnpm and Yarn Classic keep the tarball hash
- re-verify the cached CLI instead of trusting a warm cache
- name the hashed artifact in the error, and stop `vp run` / `vp exec` from hiding an integrity failure behind "command not found"

Fixes #2209

Tests: `just lint`, `cargo fmt --all -- --check`, `cargo test -p vp_error -p vp_pm_cli -p vite-plus-cli`, and three PTY cases under `install_yarn_corepack_hash`. The #2209 pin was checked by hand with a cold, warm and tampered cache.

### DIFF
--- a/crates/vp_pm_cli/src/package_manager.rs
+++ b/crates/vp_pm_cli/src/package_manager.rs
@@ -106,14 +106,15 @@ impl PackageManagerType {
         }
     }
 
-    /// Whether Corepack integrity pins for this package-manager version cover
-    /// the extracted CLI binary rather than the npm tarball.
+    /// Whether a Corepack pin for this version covers the extracted CLI binary
+    /// and not the npm tarball.
     #[must_use]
     pub fn uses_cli_binary_hash(self, version: &str) -> bool {
         Version::parse(version).is_ok_and(|version| self.hashes_cli_binary_of(&version))
     }
 
-    /// [`Self::uses_cli_binary_hash`] for a version the caller already parsed.
+    /// The same test as [`Self::uses_cli_binary_hash`], for a version that the
+    /// caller already parsed.
     #[must_use]
     pub fn hashes_cli_binary_of(self, version: &Version) -> bool {
         matches!(self, Self::Yarn) && is_yarn_berry(version)
@@ -122,17 +123,17 @@ impl PackageManagerType {
 
 /// Path of the Yarn CLI inside `@yarnpkg/cli-dist`, relative to the package root.
 ///
-/// Corepack pins Yarn 2+ by hashing this file, so the download, the cached-CLI
-/// check, and the error message must all name the same path.
+/// Corepack hashes this file to pin Yarn 2+. Three places must name the same
+/// path: the download, the cached-CLI check, and the error message.
 const YARN_CLI_ENTRY: &str = "bin/yarn.js";
 
 /// Whether a Yarn version is Berry (Yarn 2 and later).
 ///
-/// Corepack splits Yarn at 2.0.0 and matches that range with
-/// `satisfiesWithPrereleases`, which drops the prerelease tag before it
-/// compares. Every 2.x prerelease is Berry there, so this compares the major
-/// alone; `VersionReq(">=2.0.0")` would exclude `4.0.0-rc.53` and send it to
-/// the Yarn Classic package, which never published it.
+/// Corepack splits Yarn at 2.0.0. It matches that range with
+/// `satisfiesWithPrereleases`, which drops the prerelease tag first. Every 2.x
+/// prerelease is therefore a Berry version, so this function compares the major
+/// number alone. `VersionReq(">=2.0.0")` excludes `4.0.0-rc.53` and sends it to
+/// the Yarn Classic package, which never published that version.
 pub(crate) fn is_yarn_berry(version: &Version) -> bool {
     version.major >= 2
 }
@@ -913,8 +914,8 @@ pub async fn download_package_manager(
     let target_dir_tmp = tmp_dir.path().to_path_buf();
 
     let download_message = format!("Downloading {package_manager_type} v{version}...");
-    // A Corepack Yarn 2+ pin only covers the CLI, so the rest of that archive
-    // stays unauthenticated and is never written to disk.
+    // A Corepack Yarn 2+ pin covers only the CLI. The rest of the archive stays
+    // unauthenticated, so vp never writes it to disk.
     let archive_file = is_modern_yarn.then(|| PathBuf::from(format!("package/{YARN_CLI_ENTRY}")));
     download_and_extract_tgz_with_hash(
         &tgz_url,
@@ -980,13 +981,13 @@ pub async fn download_package_manager(
     Ok((install_dir, package_name, version))
 }
 
-/// Resolve the executable path of a managed package manager, installing it when
-/// the cache cannot serve it.
+/// Resolve the executable path of a managed package manager. Install that
+/// package manager when the cache cannot serve it.
 ///
-/// Takes the fast path when the shim already exists, except for a pin that
-/// covers a file inside the install: [`verify_cached_cli_hash`] has to re-read
-/// that file before anything executes it. Keeping that rule here is what lets
-/// the shim hot path stay free of package-manager specifics.
+/// This function returns the cached path when the shim already exists. A pin
+/// that covers a file inside the install is the exception:
+/// [`verify_cached_cli_hash`] must read that file again before vp runs it. The
+/// rule stays here, so the shim hot path holds no package-manager specifics.
 pub async fn ensure_package_manager_bin(
     package_manager_type: PackageManagerType,
     version: &str,
@@ -1009,11 +1010,11 @@ pub async fn ensure_package_manager_bin(
     Ok(package_manager_bin_path(&install_dir, bin_name))
 }
 
-/// Re-check a cached CLI against a pin that covers it.
+/// Verify a cached CLI against a pin that covers it.
 ///
-/// Corepack hashes the extracted Yarn 2+ CLI instead of the npm tarball, so
-/// that pin can be checked without downloading anything. Every other pin names
-/// a tarball vp no longer keeps, and passes here unchecked.
+/// Corepack hashes the extracted Yarn 2+ CLI and not the npm tarball, so vp can
+/// verify that pin from the cache. Every other pin names a tarball that vp does
+/// not keep, so this function accepts it without a check.
 async fn verify_cached_cli_hash(
     package_manager_type: PackageManagerType,
     install_dir: &AbsolutePath,
@@ -1032,10 +1033,11 @@ async fn verify_cached_cli_hash(
         .map_err(|error| name_hashed_artifact(error, package_manager_type, version))
 }
 
-/// Name the artifact a `packageManager` hash covers in an integrity failure.
+/// Name the artifact that a `packageManager` hash covers in an integrity
+/// failure.
 ///
-/// `Error::HashMismatch` alone reads like a corrupt download, which sent the
-/// reporter of #2209 looking for a network problem instead of a hash basis.
+/// `Error::HashMismatch` alone reads like a corrupt download. The reporter of
+/// #2209 looked for a network problem, not for a different hash basis.
 fn name_hashed_artifact(
     error: Error,
     package_manager_type: PackageManagerType,
@@ -1831,8 +1833,8 @@ mod tests {
 
     /// Build an `@yarnpkg/cli-dist` style tarball around `yarn_js`.
     ///
-    /// `symlink_target` adds a `package/bin/yarn` symlink, the archive-controlled
-    /// entry an unauthenticated tarball could use to write outside the install.
+    /// `symlink_target` adds a `package/bin/yarn` symlink. An unauthenticated
+    /// tarball can use that entry to write outside the install directory.
     fn create_yarn_package_tgz(yarn_js: &[u8], symlink_target: Option<&Path>) -> Vec<u8> {
         let mut tar_builder = tar::Builder::new(Vec::new());
         let mut header = tar::Header::new_gnu();
@@ -2033,8 +2035,8 @@ mod tests {
         assert!(!PackageManagerType::Yarn.uses_cli_binary_hash("latest"));
 
         // Corepack drops the prerelease tag before it matches its `>=2.0.0`
-        // range, so a 2.x prerelease pin is a Berry pin there too. `corepack
-        // use yarn@4.0.0-rc.53` writes a hash of `bin/yarn.js`.
+        // range. A 2.x prerelease pin is therefore a Berry pin there too.
+        // `corepack use yarn@4.0.0-rc.53` writes a hash of `bin/yarn.js`.
         assert!(PackageManagerType::Yarn.uses_cli_binary_hash("2.0.0-rc.1"));
         assert!(PackageManagerType::Yarn.uses_cli_binary_hash("4.0.0-rc.53"));
     }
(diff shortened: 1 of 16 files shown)


leslieeilsel:
### Local verification results (run on macOS, `nightly-2026-08-02`)

Since the CI workflows are waiting for maintainer approval, I ran the checks locally on commit `26615d1`:

- ✅ `cargo test -p vp_pm_cli --lib` — **734 passed, 0 failed** (2 ignored), including `package_manager::tests` and `request::tests`
- ✅ `cargo fmt --check` — clean
- ✅ `cargo clippy -p vp_pm_cli --lib -- -D warnings` — no warnings

One note: the PR description's Test Plan references `vite_install`, but the changed crate is `vp_pm_cli` (the commands above use the actual crate name).

Could a maintainer please approve the workflow runs? Thanks! 🙏

fengmk2:
@leslieeilsel Can you convert the reproduction steps described in https://github.com/voidzero-dev/vite-plus/issues/2209#:~:text=in%20full%20below.-,Steps%20to%20reproduce,-mkdir%20vp%2Dhash into a snapshot test? This would ensure that this issue continues to be validated.

leslieeilsel:
Thanks for the suggestion — I’ve added CLI snapshot coverage based on #2209’s cold-cache reproduction. It uses the exact Corepack-generated Yarn 4.17.1 pin, runs `vp install`, and verifies that the extracted `bin/yarn.js` is cached.

I also added a `vp run` case covering both cold and warm cache behavior from the issue follow-up. Both snapshot cases pass locally in record and compare modes.

fengmk2:
@codex review

leslieeilsel:
Thanks for the detailed review. I addressed all three findings in `29036b2f`:

- Modern Yarn npm tarballs now extract only `package/bin/yarn.js`; non-regular entries and unrelated archive contents are not written. A regression test covers the malicious `bin/yarn` symlink case.
- Yarn CLI extraction and hash verification now run inside the existing download retry pipeline, with a clean target directory for every attempt. A regression test covers a corrupted first response followed by a valid response.
- Hash-pinned Yarn 2+ resolutions now re-verify the cached `yarn.js` before global shim dispatch. Unpinned installs and other package managers retain the existing fast path. A regression test covers a corrupted warm cache.

The existing archive-hash behavior for npm, pnpm, Yarn Classic, and other package managers remains unchanged.

Local verification:

- `cargo fmt --all -- --check`
- `cargo test -p vp_pm_cli --locked` — 744 passed, 2 ignored
- `cargo test -p vp_global_cli --locked` — 461 passed, 2 ignored
- `VP_SNAP_SKIP_FLAVORS=local cargo test -p vp_cli_snapshots --locked -- install_yarn_corepack_hash` — 2 passed
- `cargo clippy -p vp_pm_cli -p vp_global_cli --all-features --all-targets --locked -- --deny warnings`

crates/vp_pm_cli/src/package_manager.rs
```diff
@@ -105,6 +110,14 @@ impl PackageManagerType {
             (_, Self::Bun) => "bun",
         }
     }
+
+    /// Whether Corepack integrity pins for this package-manager version cover
+    /// the extracted CLI binary rather than the npm tarball.
+    #[must_use]
+    pub fn uses_cli_binary_hash(self, version: &str) -> bool {
+        matches!(self, Self::Yarn)
+            && Version::parse(version).is_ok_and(|version| version >= Version::new(2, 0, 0))
```
fengmk2: This is different from `VersionReq::parse(">=2.0.0")`, so Yarn prereleases now route to `@yarnpkg/cli-dist. packageManager: "yarn@4.0.0-rc.53"` fails

fengmk2:
@leslieeilsel I will take over your PR to fix it, as I want to include this fix in today's new version release.

crates/vp_pm_cli/src/package_manager.rs
fengmk2: Good catch, and the difference went further than the comment: Corepack matches its Yarn ranges with `satisfiesWithPrereleases`, which **drops the prerelease tag** before comparing, so `2.0.0-rc.1` and `4.0.0-rc.53` are both Berry versions to Corepack. Full-semver comparison agreed with Corepack on `4.0.0-rc.53` but not on `2.0.0-rc.1`.

Fixed in 762ddabd8 by comparing the major alone, with the test extended to both prereleases.

The old `VersionReq(">=2.0.0")` excluded every prerelease and routed them to the Yarn Classic package, which never published them:

```
# v0.2.8, packageManager: "yarn@4.0.0-rc.53"
error: Package manager yarn@4.0.0-rc.53 not found on https://registry.npmjs.org/yarn/-/yarn-4.0.0-rc.53.tgz
```

With this branch, a pin written by `corepack use yarn@4.0.0-rc.53` installs, and the cached CLI matches the hash Corepack wrote:

```
packageManager: yarn@4.0.0-rc.53+sha512.2875e8627dda7e6d…
sha512(package_manager/yarn/4.0.0-rc.53/yarn/bin/yarn.js) = 2875e8627dda7e6d…
```

fengmk2:
@codex review

fengmk2:
@codex review

docs/guide/install.md
```diff
@@ -39,6 +39,13 @@ A range resolves to an already-downloaded satisfying version when possible, othe
 
 Vite+ currently downloads the declared package manager (the `onFail: "download"` behavior); the other `onFail` values are accepted but not yet differentiated.
 
+A `packageManager` pin can carry an integrity hash (`yarn@4.17.1+sha512.…`). `corepack use` writes that hash. Vite+ hashes the same artifact as Corepack:
+
+- the extracted CLI binary (`bin/yarn.js`) for Yarn 2 and later
+- the npm package tarball for npm, pnpm, and Yarn Classic
+
+Vite+ verifies a Yarn 2+ pin against the cached CLI on every command. A changed cache fails the check, and the command stops.
```
fengmk2: Why does it need to be verified every time? It should only be verified during the first installation.

fengmk2: You are right, and the cost was worse than "every install": `vp run`, `vp exec` and every `yarn` shim invocation resolved the package manager, so each one read and SHA-512'd the 3 MB `bin/yarn.js`.

Fixed in f23b50bb6. vp hashes the CLI once, when it installs Yarn, and writes the pin it verified to `$VP_HOME/package_manager/yarn/<version>/.verified-pin` (about 135 bytes). A later command compares its own pin against that record and stops there.

Two cases still hash the CLI once:

- the pin differs from the record, so a project that edits its `packageManager` hash cannot pass on a warm cache. That was the complaint in the second half of #2209 ("verification appears to be skipped on a cache hit").
- the record is missing, which is any install made by an older vp.

The trade-off worth naming: vp no longer detects a `bin/yarn.js` that changed on disk after the install. Corepack behaves the same way (it hashes on install and keeps the result in its own cache metadata), and writing into `$VP_HOME` already needs write access to the user's home directory.

The two mismatch snapshot cases now change the pin instead of the cached file, since that is the failure the check still catches.

fengmk2:
@codex review

crates/vp_pm_cli/src/package_manager.rs
```diff
@@ -944,6 +1008,105 @@ pub async fn download_package_manager(
...
+    if !package_manager_type.uses_cli_binary_hash(version) {
+        return Ok(());
+    }
+
+    let record_path = target_dir.join(VERIFIED_PIN_RECORD);
+    if let Ok(recorded_pin) = tokio::fs::read_to_string(&record_path).await
+        && recorded_pin.trim() == expected_hash
+    {
+        tracing::debug!("cached {package_manager_type} matches the recorded pin");
+        return Ok(());
```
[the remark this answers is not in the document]
fengmk2: Fair catch on the record-only form. Fixed in e5f244ceb: the record now holds the pin **plus the size and the modification time** of the file vp hashed, and a resolution compares them with one `stat` before it trusts the record. A replaced or corrupted `bin/yarn.js` sends vp back to the full hash, so the shim no longer dispatches it. The two mismatch snapshot cases cover that path again by overwriting the cached CLI.

Context on why this is not a full re-hash: the per-command hash this replaced read and SHA-512'd 3 MB on every `vp run`, `vp exec`, and `yarn` shim invocation, N times over in `yarn workspaces foreach`. The maintainer asked for install-time verification instead (#2227 review). The `stat` guard keeps that cost at one syscall while still catching a changed file.

Residual limitation, stated plainly: a replacement that preserves both size and mtime passes. Corepack has the same property (it hashes on install and trusts its cache afterwards), and writing into `$VP_HOME/package_manager` already requires write access to the user's home directory. A stronger guarantee needs the bytes read on every dispatch, which is the cost that was rejected.

fengmk2:
@codex review

crates/vp_pm_cli/src/package_manager.rs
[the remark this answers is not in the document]
fengmk2: Correcting my previous reply: I removed the size/mtime check again in 4b78a9d4a, because it did not deliver what I claimed for it.

`.verified-pin` sits in the same directory as `bin/yarn.js`. A writer who can replace the CLI can rewrite the record in the same step, and the check also falls to `touch -r` or any same-size replacement. It caught accidental corruption, not an attacker, so presenting it as a fix for this finding was wrong.

Where that leaves the design, stated without overclaiming:

- vp hashes the CLI when it installs it, and fails on a mismatch, including the npm-tarball path.
- vp records the pin it verified and compares a later command's pin against that record, so a project that changes its `packageManager` hash cannot pass on a warm cache. Corepack does not do this: `installVersion` returns the cached `.corepack` record and `ensurePackageManager` then overwrites the requested reference with the recorded hash.
- vp does not re-read `bin/yarn.js` on later commands. That matches Corepack, which reads `.corepack` and returns. The trust boundary is write access to `$VP_HOME`, which also holds the vp binary, the generated shims, and the managed Node runtime.

Reading the bytes on every dispatch is the only thing that would authenticate the file, and that cost (3 MB hashed per `vp run`, `vp exec`, and `yarn` shim invocation) is what the maintainer rejected in this PR.

fengmk2:
@codex review

fengmk2 (approved):

PR #2236, opened 2026-07-24
docs(global-cli): correct stale delegation comments

shulaoda:
Two outdated bits in the global CLI, no behaviour change beyond the help string:

#### 1. Stale "stubs" comments (`cli.rs`)

The Category C (local-CLI delegation) commands were commented as `(stubs for now)` / `(stubs)`, but every arm fully forwards to the local vite-plus CLI via `commands::delegate::execute` — they are real, working delegations, not placeholders. Both comments now read `(forwarded to the local vite-plus CLI)`.

#### ~~2. Misleading run/exec help hint (`help.rs`)~~

The `run` and `exec` delegated help docs ended their options with:

```
-h, --help    Print help (see more with '--help')
```

But `Run`/`Exec` are declared `#[command(disable_help_flag = true)]`, and their help is printed by a single static doc — `-h` and `--help` produce **identical** output, so the "(see more with '--help')" hint is inaccurate copied clap boilerplate. The other nine delegated docs already use plain `Print help`; run/exec now match.


fengmk2:
@shulaoda need to fix conflicts first.

shulaoda:
It seems the related logic has been removed or refactored, so I’ve only kept the comment changes.

fengmk2 (approved):

PR #2245, opened 2026-07-25
fix(vscode): stop the oxc language server from crashing on startup

jong-kyung:
## Summary
The oxc language server crashes on startup in this repo, so there is no linting or formatting for the session.

Screenshot 2026-07-25 at 9 08 04 PM

Nested config search is on by default, so the server evaluates every `vite.config.ts` in the workspace inside its own Node process — including the PTY snapshot fixtures, one of which crashes Node on purpose (`fixtures/migration_config_process_crash_isolated`).

Pin the root config and disable nested search in `.vscode/settings.json`, mirroring the `vitest.configSearchPatternExclude` entry already there for the same fixtures.


leaysgur:
This might be trickier than expected. 👀 

First, specifying `configPath` is equivalent to setting `disableNestedConfig`, so there's no need to specify both.

Second, I think this applies to oxfmt as well, so we might also need to specify `oxc.fmt.configPath|disableNestedConfig`?

And for oxfmt, I need to implement disableNestedConfig for LSP…!

---

NOTE: If I remember correctly, this `.vscode/settings.json` file is only referenced when the workspace is set to the directory containing `.vscode`.

In other words, if a subdirectory other than the root is opened as the workspace, this setting won't be referenced.
If there happens to be the problematic fixture under that directory, it'll still result in an error.

In normal use cases, users wouldn't leave a misconfigured file lying around, so this is mostly just an inconvenience for us, but anyway...

jong-kyung:
@leaysgur Thanks for the review! I didn't realize the two options do the same thing — in that case I'll keep just `disableNestedConfig`. As for opening a sub-directory as the workspace as you said we can't prevent that, but it doesn't seem like a common use case, so I think adding only `disableNestedConfig` is enough for now 🤔

Bonus: with only `disableNestedConfig` enabled, I'm not seeing any errors from either `oxlint` or `oxfmt`.

liangmiQwQ:
Once we find a solution that works for both Oxlint and Oxfmt, would it make sense for `vp migrate` and `vp create`  to include it in the generated editor settings as well?

It now only includes Oxfmt's

https://github.com/voidzero-dev/vite-plus/blob/b99c778b0741f58c2acbac3ae00645ad860697ae/packages/cli/src/utils/editor.ts#L29

That said, I personally still hope #997 can get supported 😃

leaysgur (approved):
> I think adding only disableNestedConfig is enough for now 🤔

I agree. 👍🏻 

> Bonus: with only disableNestedConfig enabled, I'm not seeing any errors from either oxlint or oxfmt.

I recall this is because the two tools have different strategies for loading config, with Oxlint loading them eagerly. Therefore, I believe that Oxfmt would also error if it encountered a corrupted config lazily.

I don't think this will be a problem at the moment, but I'll go ahead and implement `oxc.fmt.disableNestedConfig`.

leaysgur:
> would it make sense for vp migrate and vp create to include it in the generated editor settings as well?

I think so too. At least, it seems that we should fix the inconsistency between two tools. 😓

leaysgur:
NOTE:

- oxfmt/lsp: https://github.com/oxc-project/oxc/pull/24965
- oxc-vscode: https://github.com/oxc-project/oxc-vscode/pull/343

Anyway, let's merge this PR for now. We can add `fmt.disableNestedConfig` if needed.

PR #2267, opened 2026-07-27
fix(migrate): tolerate nested npm overrides during bootstrap detection

### LINKED ISSUE #2002: fix(migrate): tolerate nested npm overrides during bootstrap detection
`detectVitePlusBootstrapPending` assumes npm `overrides` values are strings, but npm supports nested override objects. A valid package.json such as this can throw during migration because the bootstrap detection calls string methods like `.startsWith()` on the override value:

```json
{
  "overrides": {
    "vite": {
      "rollup": "..."
    }
  }
}
```

Expected behavior:

- Migration should not crash when `overrides.vite` or `overrides.vitest` is a nested object.
- Non-string override values should be treated as not satisfying the Vite+ override and then handled safely by the migration rewrite logic.
- Add a regression test covering a nested npm override under a managed key.

Suggested test:

Create an npm fixture with nested overrides:

```json
{
  "devDependencies": {
    "vite-plus": "0.1.x"
  },
  "overrides": {
    "vite": {
      "rollup": "^4.0.0"
    }
  }
}
```

Run:

```bash
vp migrate --no-interactive
```

Assert that migration completes or reports an actionable migration error, but does not throw a TypeError from bootstrap detection.

charpeni:
## Summary

- guard `overrideSpecSatisfiesVitePlus` so a non-string override value never reaches `spec.startsWith()`, which crashed `vp migrate` with `TypeError: spec.startsWith is not a function`
- treat nested npm override objects (e.g. `{"overrides": {"vite": {"rollup": "..."}}}`) as not satisfying the managed override, so the bootstrap stays pending and `ensureOverrideEntries` writes the managed alias
- add a regression test covering nested objects under both managed keys (`vite` and `vitest`)

## Why

npm supports nested `overrides` objects (a user override scoped *under* a dependency), but `detectVitePlusBootstrapPending` assumed every override value is a string. A valid package.json with a nested override under a managed key crashed migration during bootstrap detection.

A nested object never aliases the dependency itself, so it cannot satisfy the managed Vite+ override. Returning `false` makes detection report the bootstrap as pending, and the existing rewrite then sets the managed spec — after which detection converges. The single guard covers every caller (`overridesSatisfyVitePlus` across npm/bun/pnpm/yarn sources, `npmVitePlusManagedDependenciesPending`, and `ensureOverrideEntries`), and also hardens against other non-string values, such as a numeric scalar read from `pnpm-workspace.yaml` overrides.

When the project does not use vitest directly, a user override scoped under `vitest` is left intact, matching the existing contract documented in `removeManagedVitestEntry`. Note that under a *managed* key the rewrite replaces the nested object with the managed string spec, dropping any user sub-overrides scoped beneath it (e.g. a `vite > rollup` pin); preserving those via npm's nested `"."` syntax is left as a possible follow-up.

Fixes #2002.

## Validation

- `pnpm -F vite-plus exec vitest run src/migration/__tests__/migrator.spec.ts -t 'tolerates nested npm override objects'` — fails without the fix with the exact reported crash (`TypeError: spec.startsWith is not a function` at `overrideSpecSatisfiesVitePlus`, reached via `detectVitePlusBootstrapPending`), passes with it
- full `src/migration/__tests__/migrator.spec.ts` run — no changes to any existing test outcome
- also exercised locally (not committed): nested objects under both keys with `usesVitest = true` (both rewritten to managed specs, detection converges) and a nested object in yarn `resolutions` (converges without crashing)
- type check of `packages/cli` is clean


fengmk2:
@codex review

PR #1686, opened 2026-05-26
feat(http): centralize HTTP client with proxy and custom-CA support

fengmk2:
Track B of [voidzero-dev/setup-vp#73](https://github.com/voidzero-dev/setup-vp/issues/73) — makes `vp` work through Socket Firewall Free (sfw) and other TLS-intercepting proxies.

## What

- **New `vite_shared::shared_http_client()`**: process-wide `reqwest::Client` that honors `HTTPS_PROXY` / `HTTP_PROXY` / `NO_PROXY`, picks up macOS System Settings / Windows registry proxies (`system-proxy` feature), loads custom CAs from `SSL_CERT_FILE` and `NODE_EXTRA_CA_CERTS` (additive, Node-style), and has connect + request timeouts. `VP_INSECURE_TLS` (truthy only — `1`/`true`/`yes`/`on`) is a diagnostic escape hatch with a loud warning.
- **All `reqwest::get` / `reqwest::Client::new()` call sites** in `vite_install` and `vite_js_runtime` routed through the shared client.
- **`vite_shared::format_error_chain`**: walks `Error::source()` so users see `... invalid peer certificate: UnknownIssuer` instead of the opaque `error sending request for url`. Wired into both `Error::Reqwest` variants and the four `DownloadFailed { reason }` sites.
- **New `install-e2e-test-sfw` CI job** (Linux / macOS / Windows): downloads upstream `sfw`, runs `sfw vp i -g pnpm@…` + `sfw vp install` against a fresh `vitejs/vite` clone. Gated on the `test: sfw` label for PRs; unconditional on push-to-main. `VP_INSECURE_TLS=1` is scoped to the Linux entry only (workaround for [SocketDev/sfw-free#30](https://github.com/SocketDev/sfw-free/issues/30) / [#43](https://github.com/SocketDev/sfw-free/issues/43) — present-but-empty EKU rejected by rustls).

## Out of scope / follow-ups

- Switch to `rustls-native-certs` so OS-installed CAs work without env vars — Track B step 5, separate PR.
- Drop `vp_insecure_tls` from the Linux matrix once SocketDev/sfw-free ships the EKU fix (one-line change).

Refs voidzero-dev/setup-vp#73

PR #2273, opened 2026-07-29
fix(http): return an error when the HTTP client cannot be built

### LINKED ISSUE #1014: reqwest calls in vp fail when running behind a MITM proxy.

jbmusso:
## Problem

This surfaced while upgrading vite-plus from 0.1.21 to 0.2.6 in a CI pipeline running on `node:24-bookworm-slim`: the lint and format jobs started aborting with exit 134, having been green on 0.1.21 on the same image.

The branch is two commits, so the bug can be inspected before the fix: check out the first commit and run `cargo test -p vite_shared --lib client_build_failure` to see the failure as a failing test, on an otherwise unmodified tree. The second commit is an attempt at a fix.

`shared_http_client` panics whenever the client cannot be built, and every input funnels into that one `panic!`: no system CA bundle, a malformed proxy setting, an unusable TLS backend, an unparsable extra CA bundle. Its doc comment states that a panic was chosen over `process::exit` so that:

```
/// destructors of in-flight work still run (lockfiles released, tempfiles
/// cleaned) and an embedding Node host (NAPI) keeps the process alive.
```

`Cargo.toml` has, under `[profile.release]`:

```toml
panic = "abort" # Let it crash and force ourselves to write safe Rust.
```

`abort` does not unwind, so in every release build the panic delivers neither property it was chosen for: lockfiles and temp dirs are not cleaned, and SIGABRT takes the embedding Node host down with it. The comment and the release profile have drifted apart, and every one of those inputs becomes an unrecoverable abort, worst in exactly the NAPI-embedded case the comment was written to protect.

The input that is near-universal in Debian-slim Node images is the missing CA bundle: they ship no `ca-certificates` package, and Node carries its own root list, so `node` and `npm` keep working and nothing else reveals it. On such an image any direct `vp` subcommand aborts instead of running:

```
Vite+ panicked. This is a bug in Vite+, not your code.

thread '<unnamed>' panicked at crates/vite_shared/src/http.rs:58:21:
failed to initialize HTTP client: builder error: unexpected error: No CA certificates were loaded from the system

Aborted (core dumped)          # exit 134
```

Reproduces in about 30s with only Docker:

```bash
docker run --rm -w /app node:24-bookworm-slim sh -c '
echo "{\"name\":\"repro\",\"private\":true}" > package.json
npm i --silent vite-plus@0.2.6
echo "{\"name\":\"repro\",\"private\":true,\"packageManager\":\"pnpm@10.27.0\"}" > package.json
./node_modules/.bin/vp fmt --check
echo "exit=$?"'
```

Holding image and project constant and varying only the vp version, 0.1.22 exits 1 with a handled error and 0.1.23 is the first release that aborts, the release that shipped the shared client from #1686. The empty trust store itself is not new (it has mattered since #1068 moved verification onto the OS store to fix #1014); what changed is that it became fatal.

Mechanism: `execute_direct_subcommand` calls `envs_with_explicit_package_manager_path` for every direct subcommand. With a `packageManager` field and a cold cache that downloads the pinned package manager, which builds the shared client. Nothing here actually needs the network, since the download failure is already handled and falls back to the unmodified `PATH`, so with any single root certificate present the same command completes normally even under `--network none`. Without a `packageManager` field there is no download attempt and no crash, which is why this only shows up on projects that pin one.

## Solution

`shared_http_client` now returns `Result<&Client, HttpClientError>`, and the six call sites `?` it into the errors they already return. That restores the behaviour the doc comment describes rather than adding a new policy: callers already treat "no HTTP client" as a handled outcome, they just never got the chance to. Every failure at that point is an environment condition (no system CA bundle, an unusable proxy, an unparsable extra CA bundle), so the message names the cause and the remedy instead of asking for a bug report, which currently points away from the fix.

The two artifacts cover different things, deliberately. The Docker snippet above reproduces the reported symptom; the unit test pins the branch. I could not reproduce the empty trust store portably in a unit test, because macOS resolves roots through Security.framework and `SSL_CERT_FILE` cannot empty the store there, so the test drives the same `build()` failure branch through the documented `SSL_CERT_FILE` contract, using a PEM block whose body is not valid DER. To check the real condition I built the patched `vite_shared` in a Debian container and deleted its CA store: it now returns the error and exits 0 instead of aborting.

I only looked at this one call site, where the doc comment states the intent explicitly, so I can't say whether the same drift between a documented panic rationale and `panic = "abort"` exists anywhere else. You may want to check.

### Note

The investigation and this patch were done with AI assistance, and my own Rust is limited, so please scrutinise the implementation closely. The diagnosis is empirically verified (bisected across releases, reproduced in a container), but I have lower confidence in the patch being the shape you would want. Happy to rework it however you prefer, or to leave the diagnosis with you if you would rather fix it yourselves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8640,6 +8640,7 @@ dependencies = [
  "serde_json",
  "serial_test",
  "supports-color 3.0.2",
+ "thiserror 2.0.19",
  "tracing",
  "tracing-subscriber",
  "vite_path",
--- a/crates/vite_error/src/lib.rs
+++ b/crates/vite_error/src/lib.rs
@@ -98,6 +98,11 @@ pub enum Error {
     #[error("{}", vite_shared::format_error_chain(.0))]
     Reqwest(#[from] reqwest::Error),
 
+    // The shared HTTP client could not be built at all, so no request was
+    // attempted. Its own message already names the cause and the remedy.
+    #[error(transparent)]
+    HttpClient(#[from] vite_shared::HttpClientError),
+
     #[error(transparent)]
     JoinError(#[from] tokio::task::JoinError),
 
--- a/crates/vite_install/src/request.rs
+++ b/crates/vite_install/src/request.rs
@@ -55,7 +55,7 @@ impl HttpClient {
     pub async fn get_bytes(&self, url: &str) -> Result<Vec<u8>, Error> {
         tracing::debug!("Fetching bytes from: {}", url);
 
-        let client = vite_shared::shared_http_client();
+        let client = vite_shared::shared_http_client()?;
 
         // Read the body inside the retry so a mid-body connection drop gets
         // retried instead of failing outright, like `download_file`.
@@ -116,7 +116,7 @@ impl HttpClient {
     ) -> Result<T, Error> {
         tracing::debug!("Fetching JSON from: {} (accept: {:?})", url, accept);
 
-        let client = vite_shared::shared_http_client();
+        let client = vite_shared::shared_http_client()?;
         (|| async {
             let mut request = client.get(url);
             if let Some(accept) = accept {
@@ -153,7 +153,7 @@ impl HttpClient {
         let target_path = target_path.as_ref();
         tracing::debug!("Downloading {} to {:?}", url, target_path);
 
-        let client = vite_shared::shared_http_client();
+        let client = vite_shared::shared_http_client()?;
 
         // Make the request *and* the body stream a single retried unit. Doing
         // the request inline (instead of calling `self.get`) avoids a double
--- a/crates/vite_js_runtime/src/download.rs
+++ b/crates/vite_js_runtime/src/download.rs
@@ -37,7 +37,7 @@ pub async fn download_file(
     target_path: &AbsolutePath,
     message: &str,
 ) -> Result<(), Error> {
-    let client = vite_shared::shared_http_client();
+    let client = vite_shared::shared_http_client()?;
 
     tracing::debug!("Downloading {url} to {target_path:?}");
 
@@ -144,7 +144,7 @@ pub async fn download_file(
 /// Download text content from a URL with retry logic
 #[expect(clippy::disallowed_types, reason = "HTTP response body is a String")]
 pub async fn download_text(url: &str) -> Result<String, Error> {
-    let client = vite_shared::shared_http_client();
+    let client = vite_shared::shared_http_client()?;
 
     tracing::debug!("Downloading text from {url}");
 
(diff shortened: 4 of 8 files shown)


fengmk2:
@codex review

fengmk2 (approved):
@jbmusso Thanks!

fengmk2:
I will merge it next week and try to support no `ca-certificates` env in a new pr.

jbmusso:
> @jbmusso Thanks!

My pleasure, thank you for this great product.
Happy to help.

PR #2294, opened 2026-08-02
refactor(cli): generate Zed language settings from a language list

jong-kyung:
## Summary

`ZED_SETTINGS.languages` in `packages/cli/src/utils/editor.ts` listed 17 per-language blocks that were verbatim-identical except for one key, so this PR replaces them with a language-name array mapped through `Object.fromEntries` (88 lines down to 30).

- The only per-language difference (`code_action: 'source.fixAll.oxc'` on JavaScript) is now an explicit conditional spread instead of being hidden inside a wall of identical blocks.
- Adding a newly supported language becomes a one-line array entry instead of a copy-pasted block.

## Behavior

No behavior change. `Object.fromEntries` preserves array order, so the generated `.zed/settings.json` is byte-identical, and `ZED_SETTINGS` is consumed as `Record<string, unknown>` so the weaker literal typing has no effect.

### DIFF
--- a/packages/cli/src/utils/editor.ts
+++ b/packages/cli/src/utils/editor.ts
@@ -60,94 +60,36 @@ const ZED_SETTINGS = {
       },
     },
   },
-  languages: {
-    CSS: {
-      format_on_save: 'on',
-      prettier: { allowed: false },
-      formatter: [{ language_server: { name: 'oxfmt' } }],
-    },
-    GraphQL: {
-      format_on_save: 'on',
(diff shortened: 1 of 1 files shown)


naokihaba:
@codex review

naokihaba (approved):
I considered if language-specific overrides might add overhead later, but since there's only one exception, this abstraction seems fine. It cleans up the duplication and makes shared updates much safer.

fengmk2 (approved):

PR #2295, opened 2026-08-02
fix(http): fall back to bundled Mozilla roots when the system CA store is empty

### LINKED ISSUE #2278: Fall back to bundled Mozilla roots when the system CA store is empty
Follow-up to #2273.

#2273 turns an unbuildable HTTP client into a handled error instead of an abort, but on images without a CA bundle (`node:*-slim`, distroless) vp still cannot make any HTTPS request: `vp install`, runtime downloads, and `vp create` / `vp migrate` fail with the remedy message. Node works on the same images because it compiles Mozilla's root list into the binary, so vp looks broken by comparison.

We moved off bundled `webpki-roots` in #1068 because a Keychain-trusted corporate CA was missing from the bundle (#1014), so going back to bundled-roots-only would regress that. The two combine instead: keep the platform store authoritative, and fall back to bundled Mozilla roots only when the store is empty.

`rustls-platform-verifier` only fails when the combined store (system + extra roots) is empty (`src/verification/others.rs`):

```rust
// Don't return an error if this fails when other roots have already been
// loaded via `new_with_extra_roots`. It leads to extra failure cases where
// connections would otherwise still work.
if root_store.is_empty() {
    return Err(rustls::Error::General(
        "No CA certificates were loaded from the system".to_owned(),
    ));
}
```

reqwest's `tls_certs_merge`, which `build_client` already uses for `SSL_CERT_FILE` / `NODE_EXTRA_CA_CERTS`, feeds `new_with_extra_roots`. Plan for `crates/vite_shared/src/http.rs`:

1. Build the client as today, system store authoritative.
2. On build failure, rebuild once with `tls_certs_merge(webpki_root_certs::TLS_SERVER_ROOT_CERTS)` and warn once: "system CA store is empty; using bundled Mozilla roots".
3. If the retry also fails, return the original error through the #2273 path.

No string-matching on the error is needed: retry once on any build failure. A failure the bundle cannot fix (malformed `HTTPS_PROXY`, bad `SSL_CERT_FILE`) fails the retry the same way and we report the original error.

Why fallback-on-empty rather than always-merge:

- A corporate MITM environment has a non-empty store, so the fallback never triggers there and #1014 stays fixed.
- A distro's decision to distrust a CA is never overridden by the frozen bundled list; in the empty-store case there is no distro decision to override.

Notes:

- Use the `webpki-root-certs` crate (DER certs, suitable for `tls_certs_merge`), not `webpki-roots` (trust anchors only). Adds a few hundred KB to the binary.
- The bundled list is frozen at build time and stays current through `cargo update` and our release cadence, the same trade Node makes.
- macOS (Security.framework) and Windows (native-tls/SChannel) always have a populated store, so in practice the fallback matters only on Linux.
- Verification: the #2273 Docker repro (`node:24-bookworm-slim`, no `ca-certificates`) should complete a network-using command like `vp install` instead of returning the remedy error.

fengmk2:
When the shared HTTP client cannot be built because the trust store is
empty (Debian slim and distroless images ship no ca-certificates
package), retry once with the webpki-root-certs Mozilla list merged in,
like Node's own bundled roots. A populated system store keeps sole
authority; failures the bundle cannot fix still surface the original
error. Windows keeps native-tls/SChannel and skips the fallback.

Closes #2278

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


fengmk2:
@codex review

fengmk2:
Verified the registry-bridge build (`175538b`) in the original repro environment (`node:24-bookworm-slim`, no `ca-certificates`, `packageManager: pnpm@10.27.0`, cold cache):

- `vite-plus@0.2.7`: `vp fmt --check` aborts with exit 134, "No CA certificates were loaded from the system"
- this build: `vp fmt --check` runs oxfmt and reports format results normally, and the pinned pnpm is downloaded through vp's HTTP client into `~/.vite-plus/package_manager/pnpm/10.27.0/`, confirming HTTPS works through the bundled Mozilla roots fallback

cpojer (approved):

PR #2296, opened 2026-08-02
refactor(cli): share agent detect/write traversal helpers

jong-kyung:
## Summary

`detectAgentConflicts` and `writeAgentInstructions` in `packages/cli/src/utils/agent.ts` duplicated ~50 lines of per-path scaffolding: AGENTS.md-first ordering, destination-path dedupe, symlink skip, realpath dedupe, and the marker-section check. Because detected conflicts feed `conflictDecisions` back into the write pass, the two traversals must apply identical rules, so the duplication was a real divergence risk rather than just noise.

This PR extracts the genuinely identical pure pieces into two helpers while each caller keeps its own loop and side effects:

- `orderAgentTargetPaths(projectRoot, paths)` combines the AGENTS.md-first ordering with dedupe by resolved destination path.
- `classifyExistingAgentTarget(destinationPath, incomingContent, seenRealPaths)` classifies an existing file as `symlink`, `duplicate`, `markers`, or `conflict`, registering realpaths in a caller-owned set.

A yielding-iterator extraction was considered and rejected because `writeAgentInstructions` mutates the filesystem mid-loop (writes AGENTS.md first, links later paths to it, registers realpaths of freshly written files), so traversal state must be read at call time per path instead of being precomputed or owned by a generator.

Not included: promoting the order-preserving dedupe loops onto `dedupeNames` from `approve-builds.ts`. Only 2 of the 4 inline loops match its return shape, so that is left for a separate change if wanted.

## Test plan

- `pnpm test:unit packages/cli/src/utils/__tests__/agent.spec.ts` — 25 tests pass.
- `vp check` — the only errors are pre-existing `versions.spec.ts` imports of unbuilt `dist/versions.js`, unrelated to this change.


jong-kyung:
@codex review

jong-kyung:
@codex review

jong-kyung:
@codex review

fengmk2 (approved):

PR #2140, opened 2026-07-10
refactor(pm-cli): consolidate package manager infrastructure

forehalo:
## Summary

This refactor makes typed command arguments the source of truth for package-manager compatibility.

`#[pm_args]` reads field-level `not_supported(...)` metadata, including manager and version conditions

**before:**

```rust
#[derive(clap::Args)]
struct InstallArgs {
    #[arg(long)]
    fix_lockfile: bool,
}

// Simulated
match command {
  Command::Install(args) => {
    match manager {
      Yarn => {
        if args.fix_lockfile {
          output::warn("yarn@1 install does not support --fix-lockfile");
        }
        // ...
      },
      Npm => {
        if args.fix_lockfile {
          output::warn("npm install does not support --fix-lockfile");
        }
        // ...
      },
      Bun => {
        if args.fix_lockfile {
          output::warn("bun install does not support --fix-lockfile");
        }
        // ...
      }
    }
  },
  // ...
}
```

**after:**

```rust
#[pm_args]
#[derive(clap::Args)]
struct InstallArgs {
    #[arg(long, not_supported(npm, bun, yarn < "2"))]
    fix_lockfile: bool,
}
```

Given the detected package manager and its version, the generated diagnosis step automatically:

- identifies active arguments that are unsupported;
- resets only those arguments to their default value;
- records a consistent warning before command resolution.

This removes repeated compatibility checks from every npm, pnpm, Yarn, and Bun resolver. Normal `clap::Args` and `clap::Subcommand` derives remain explicit. Hard errors, fallback commands, and manager-specific translations stay in the resolver.

## Architecture and benefits

The production path is now:

`typed clap arguments -> #[pm_args] diagnosis -> manager-specific Resolve<Args> -> Resolution -> runner`

- Global and local CLIs share one typed `PackageManagerCommand` surface.
- Parsing, compatibility diagnosis, and production dispatch use the same argument types.
- Resolution is side-effect free and directly testable. The runner executes the completed plan afterward.
- Package-manager detection, downloads, caching, HTTP access, and shims now live in `vite_pm_cli`; the redundant `vite_install` crate is removed.

## Adding a command

1. Define an `XxxArgs` clap struct, or an `XxxCommand` nested subcommand enum, and add `#[pm_args]`.
2. Declare cross-manager and cross-version compatibility with field-level `not_supported(...)` rules.
[...]

PR #2310, opened 2026-08-03
feat(pm): verify bun platform tarball against dist.integrity

jong-kyung:
## Summary

Re-does #2301 on top of the pm-cli consolidation (#2140), which moved `vite_install/src/package_manager.rs` into `crates/vite_pm_cli` and made the old branch unrebaseable.

- The `packageManager` hash names the main `bun` npm package, so the platform tarball (`@oven/bun-{os}-{arch}`) was downloaded without any verification.
- Extracts the registry `dist.integrity` lookup that the pnpm >= 12 native path already implements inline into a shared `fetch_platform_integrity` helper, and applies it to the bun platform tarball as well.
- The pnpm >= 12 path now uses the same helper, so its behavior is unchanged.

Supersedes #2301.

### DIFF
--- a/crates/vite_pm_cli/src/package_manager.rs
+++ b/crates/vite_pm_cli/src/package_manager.rs
@@ -849,7 +849,8 @@ pub async fn download_package_manager(
 
     // For bun, use platform-specific download flow.
     // The hash from `packageManager` field belongs to the main `bun` npm package,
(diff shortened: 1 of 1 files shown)


jong-kyung:
@codex review

fengmk2 (approved):

PR #2312, opened 2026-08-04
refactor(env): rename remaining Vite+ environment variables

### LINKED ISSUE #2206: Rename the remaining env vars that starts with `VITE_`
### Problem

#1074 and #1166 already rename environment variables which begin with `VITE_PLUS_` to `VP_`. However, there are still some env variables that start `VITE_`, influencing users' daily using.

### List

| Original Name | New Name | Description |
|---------|---------|---------|
| `VITE_GIT_HOOKS` | `VP_GIT_HOOKS` | Vite+ only env vars, for git hooks behavior |
|`VITE_LOG` | `VP_LOG` | Vite+ only, define log level |
| `VITE_GLOBAL_CLI_JS_SCRIPTS_DIR` | `VP_GLOBAL_CLI_JS_SCRIPTS_DIR` | Vite+ only, [mise compatibility](https://github.com/voidzero-dev/vite-plus/issues/943#issuecomment-4066275286) | 

Other env variables, like `VITE_PACK_` should be treated as Vite itself's internal env vars, should not be changed.

### Opening Questions / Trade-off

We either mark these changes as a breaking change like #1166 or still read the old names for compatibility. The former is simple but maybe influence version handling, the latter may leave legacy code forever.

/cc @jong-kyung

jong-kyung:
## Summary

This follows up on #1166, which renamed `VITE_PLUS_*` environment variables to use the `VP_*` prefix. It applies the
 same convention to the remaining Vite+-specific variables: `VITE_LOG`, `VITE_GLOBAL_CLI_JS_SCRIPTS_DIR`, and
 `VITE_UPDATE_TASK_TYPES`, without retaining compatibility aliases.

Resolves #2206

### DIFF
--- a/.github/workflows/test-standalone-install.yml
+++ b/.github/workflows/test-standalone-install.yml
@@ -197,7 +197,7 @@ jobs:
               done
               vp env doctor
 
-              export VITE_LOG=trace
+              export VP_LOG=trace
               vp env run --node 24 -- node -p \"process.versions\"
 
               # Verify upgrade
@@ -307,7 +307,7 @@ jobs:
               done
               vp env doctor
 
-              export VITE_LOG=trace
+              export VP_LOG=trace
               vp env run --node 24 -- node -p \"process.versions\"
 
               # FIXME: QEMU doesn't support all syscalls needed by rolldown/tsdown
@@ -693,7 +693,7 @@ jobs:
           echo "PATH: $env:Path"
           vp --version
           vp --help
-          # $env:VITE_LOG = "trace"
+          # $env:VP_LOG = "trace"
           # test create command
           vp create vite --no-interactive --no-agent -- hello --no-interactive -t vanilla
           cd hello && vp run build && vp --version
--- a/crates/vite_global_cli/src/error.rs
+++ b/crates/vite_global_cli/src/error.rs
@@ -18,7 +18,7 @@ pub enum Error {
     CommandExecution(#[from] io::Error),
 
     #[error(
-        "JS scripts directory not found. Set VITE_GLOBAL_CLI_JS_SCRIPTS_DIR or ensure scripts are bundled."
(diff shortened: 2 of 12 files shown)


jong-kyung:
@codex review

fengmk2:
The `VITE_UPDATE_TASK_TYPES` environment variable doesn't seem to be used anywhere? I didn't see it being set anywhere.

jong-kyung:
> The `VITE_UPDATE_TASK_TYPES` environment variable doesn't seem to be used anywhere? I didn't see it being set anywhere.

It is read in `packages/cli/binding/src/cli/mod.rs`. I wasn't sure whether it should be renamed or removed, so I left it out of this PR and raised the question here.

https://github.com/voidzero-dev/vite-plus/blob/d1183bd7569fcee10ea88dd625b24bc698b20d62/packages/cli/binding/src/cli/mod.rs#L460-L480

fengmk2:
> It is read in `packages/cli/binding/src/cli/mod.rs`. I wasn't sure whether it should be renamed or removed, so I left it out of this PR and raised the question here.

Here, we are just reading `VITE_UPDATE_TASK_TYPES`, but it needs to be confirmed where it is set.

jong-kyung:
I couldn't find anywhere in the repository that sets it either. It appears to be intended as a developer flag. Should we remove it? 🤔

fengmk2:
> I couldn't find anywhere in the repository that sets it either. It appears to be intended as a developer flag. Should we remove it? 🤔

Sure, go ahead and delete it with a new PR.

cc @wan9chi

jong-kyung:
Renamed `VITE_UPDATE_TASK_TYPES` to `VP_UPDATE_TASK_TYPES` while preserving the manual regeneration flow.
Context: https://github.com/voidzero-dev/vite-plus/pull/2315#discussion_r3709074980

wan9chi (approved):

fengmk2 (approved):

PR #2283, opened 2026-07-31
test: deflake env_install_interrupt by killing reinstall on install dir creation

fengmk2:
A fixed 100ms kill can fire before vp creates the new install dir on slow
Windows runners, leaving no stale dir for check-stale-packages to find.
Poll for the new install dir and kill as soon as it appears; the package's
200ms postinstall keeps the install running well past that point.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #2316, opened 2026-08-04
test(snapshots): kill the real vp process in env_install_interrupt

fengmk2:
The Windows `env_install_interrupt` case kept flaking after #2283 and #2299, in two modes: the check-stale step reported "interrupted stale package removed" instead of "exists", and a 60s zero-output step timeout.

Root cause: on Windows, `vp` on PATH is the trampoline in `VP_HOME/bin`, which runs `current/bin/vp.exe` as a child and waits. The interrupt driver's `child.pid` was the trampoline, so #2299 killed the wrong process first. The real vp.exe was then torn down by `taskkill /T`, which kills the tree in unspecified order; whenever the npm subtree died before vp.exe, the still-alive vp treated the reinstall as failed and removed the partial install dir the test asserts on. When child enumeration returned nothing, only the trampoline died and the real vp survived as an orphan, matching the zero-output mode. Unix never flaked because `bin/vp` is a symlink, so the killed child is the installer.

Fix, all inside the fixture:

- Spawn `VP_HOME/current/bin/vp[.exe]` directly so `child.pid` is the process that runs the install on every platform; killing it first now prevents the cleanup reaction.
- Kill the postinstall subtree via a PID handshake: postinstall writes its PID into the ready file (write-then-rename), so the driver needs no PowerShell/CIM enumeration. The npm layers in between exit on their own once the script dies, and with vp already dead nobody reacts.
- Add stage markers and a 45s watchdog so a future failure names the stage that hung instead of showing a blank PTY screen.

Verified on macOS with `cargo test -p vite_cli_snapshots -- env_install_interrupt`; the Windows race itself can only be exercised by this PR's Windows CLI snapshot leg.

### DIFF
--- a/crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/env_install_interrupt/long-time-install-package/postinstall.js
+++ b/crates/vite_cli_snapshots/tests/cli_snapshots/fixtures/env_install_interrupt/long-time-install-package/postinstall.js
@@ -6,7 +6,13 @@ const sleep = promisify(setTimeout);
 
 (async () => {
   if (process.env.VP_TEST_INTERRUPT_INSTALL === '1') {
-    fs.writeFileSync(path.join(process.env.VP_HOME, 'env-install-interrupt-ready'), '');
+    // The PID lets the interrupt driver kill this subtree directly instead
+    // of enumerating vp's children through PowerShell/CIM, whose cold start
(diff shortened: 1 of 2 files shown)


fengmk2:
I will retry many times before merge.

fengmk2:
Attempt 10 is green and the whole CI run now concludes success

cpojer (approved):

PR #2319, opened 2026-08-04
chore(deps): declare @emnapi peers where @napi-rs/cli is used

jong-kyung:
## Summary

`@napi-rs/cli@3.8.2` requires the peer `@emnapi/runtime@2.0.0-alpha.3` exactly, but the root, `packages/cli`, and `packages/core` manifests declare only `@napi-rs/cli`. pnpm fills the missing peer with the transitive `@emnapi/runtime@1.11.2`, which violates the exact pin, so every `pnpm install` prints a missing/conflicting-peer warning block.

This declares `@emnapi/core` and `@emnapi/runtime` as devDependencies (catalog-pinned to `2.0.0-alpha.3`) in the three manifests, mirroring `rolldown/packages/rolldown`. Re-resolution also dedupes the bogus `1.11.2` peer variants from the lockfile.

**Before:**
image

**After:**
image

## Test plan

- `pnpm install` — @emnapi warning block no longer appears

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
     "prepare": "husky"
   },
   "devDependencies": {
+    "@emnapi/core": "catalog:",
+    "@emnapi/runtime": "catalog:",
     "@napi-rs/cli": "catalog:",
     "@oxc-node/cli": "catalog:",
     "@types/node": "catalog:",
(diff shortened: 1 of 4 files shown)


fengmk2 (approved):

fengmk2:
changed come from https://github.com/napi-rs/napi-rs/pull/3353/changes#diff-088b5e0ac52dd36816ebc6b121d5423c763c3c18c098b91757c014937c7d632dR94-R95

PR #2344, opened 2026-08-05
refactor(create): replace hand-rolled copy helpers with fs.cpSync

jong-kyung:
## Description

Replaces the hand-written recursive `copy()`/`copyDir()` helpers in `packages/cli/src/create/utils.ts` with a single `copyDir()` wrapper over `fs.cpSync(src, dest, { recursive: true, dereference: true })`.

- `copy()` had no external callers, so it is removed. All three call sites (`templates/bundled.ts`, `templates/generator.ts`, `templates/monorepo.ts`) only use `copyDir` and are unchanged.
- `dereference: true` preserves the previous behavior, because the old implementation used `statSync` which follows symlinks.
- Overwrite behavior and parent directory creation match the previous implementation via `cpSync` defaults.
- `fs.cpSync` is available across the package's supported Node range (`^20.19.0 || ^22.18.0 || >=24.11.0`).

## Validation

- `tsc --noEmit` on `packages/cli` passes.
- No remaining imports of the removed `copy` export.

### DIFF
--- a/packages/cli/src/create/utils.ts
+++ b/packages/cli/src/create/utils.ts
@@ -39,23 +39,10 @@ export function shouldConfigureEditorsForCreate({
   return hasExplicitEditorOptIn(editor);
 }
 
-// Helper functions for file operations
-export function copy(src: string, dest: string) {
-  const stat = fs.statSync(src);
-  if (stat.isDirectory()) {
-    copyDir(src, dest);
-  } else {
-    fs.copyFileSync(src, dest);
-  }
-}
-
+// Helper function for file operations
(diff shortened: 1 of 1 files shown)


jong-kyung:
@codex review

jong-kyung:
I expected `fs.cpSync` with `dereference: true` to be a drop-in replacement, but I reproduced [the issue](https://github.com/voidzero-dev/vite-plus/pull/2344#discussion_r3720364687) locally:nested symlinks are preserved as symlinks instead of being dereferenced. Keeping the hand-rolled recursion is the simpler outcome here, so I'm closing this PR.

PR #2349, opened 2026-08-05
feat: remove VP_HOME and default installers to the split XDG layout

### LINKED ISSUE #827: use xdg config
### Description

config is created in $HOME/.vite-plus

### Suggested solution

it should respect xdg configuration and create config in xdg defined paths. e.g. $HOME/.config

### Alternative

_No response_

### Additional context

_No response_

### Validations

forehalo:
Stacked on #2346 — only the top 4 commits are new. Close #827 on merge of this stack.

## Summary

Flips fresh installs to the split XDG layout and removes the `VP_HOME` variable, completing #827. Existing installs are untouched: a legacy monolithic root is still detected (executable self-location, `PATH` inference, or an existing `~/.vite-plus`) and keeps working byte-for-byte.

### Commit 1: remove `VP_HOME`

The CLI no longer reads `VP_HOME`. Custom-location installs are located without it via executable self-location (`<root>/current/bin/vp`, covering IDE and trampoline launches), legacy `PATH` inference, and the grandfathered `~/.vite-plus` check. The Windows trampoline and `vp_installer` stop setting it.

### Commit 2: split-aware detection, env scripts, implode, installer core

- `dirs.rs`: detection now distinguishes legacy from split installs via a `DetectedLayout` enum — a `<X>/current/bin/vp` shape classifies as legacy when `<X>/bin/vp` exists, otherwise as a split install whose data dir is `X` (pinning data to the running binary's payload). `PATH` inference canonicalizes `vp` entries through the same shape check.
- Generated `env`/`env.fish`/`env.nu`/`env.ps1` no longer export `VP_HOME`; they only prepend the resolved bin dir. `vp-use.cmd` gains the split fallback path.
- The Windows trampoline locates `vp.exe` at `../current/bin` (legacy) or `../data/current/bin` (split).
- `vp implode` removes the split layout: data/config/state/cache dirs plus only the vp-owned shim names from the bin dir — a shared `~/.local/bin` and foreign files are never touched. Profile stripping catches the env-scripts dir in both layouts.
- `vp_installer` resolves an `InstallLayout` from `Dirs`, creates the bin wrapper in the resolved bin dir, and prints split-aware guidance. Explicit `--install-dir`/`VP_HOME` still selects the legacy layout.

### Commit 3: installers default to the split layout

- `install.sh`/`install.ps1`: fresh installs place versions + `current` under the data dir (`~/.local/share/vite-plus`, `%LOCALAPPDATA%\vite-plus\data`), the `vp` wrapper and shims in the bin dir (`~/.local/bin`, `%LOCALAPPDATA%\vite-plus\bin`), and env scripts in the config dir (`~/.config/vite-plus`, `%APPDATA%\vite-plus`). Explicit `VP_HOME` or an existing `~/.vite-plus` keeps the legacy layout, so re-running the installer upgrades existing installs in place. Relative XDG values are ignored per spec.
- Dockerfile drops `VP_HOME` and uses the split layout.
- Docs describe the split layout as the fresh-install default and `VP_HOME` as an installer-only override.

### Commit 4: snapshot fixtures

- The runner provisions the complete legacy shape (`bin/vp` alongside `current/bin/vp`) so detection classifies case homes as legacy installs.
- Two fixtures that pointed at isolated installs via per-step `VP_HOME` now model the same isolation with real on-disk layouts driven through their own binaries (self-location).
- The git-hooks migration snapshot picks up PR #2346's new hook-script PATH fallback chain.

## Migration impact

- Existing `~/.vite-plus` installs: zero impact (grandfathered; detection is byte-identical in behavior).
- Custom `VP_HOME` installs: keep working — the running binary self-locates from its own path. The variable itself is ignored by the CLI and no longer exported by generated env scripts.
- Fresh installs: land in the split layout; `~/.local/bin` is the only PATH entry needed.

## Validation

- `cargo check`/`clippy --workspace --all-targets`: zero warnings; `cargo fmt` clean
- `cargo test` on all touched crates: green
- PTY snapshot suite: 622/622 cases pass locally (including two rewritten fixtures and the browser-mode case)


forehalo:
Consolidated into #2346 — the stack is now a single PR: the Dirs abstraction commit plus the squashed installer-cutover commit (with VP_HOME kept as DEPRECATED_VP_HOME instead of removed outright).

PR #2350, opened 2026-08-05
fix(command): resolve relative PATH entries against cwd

### LINKED ISSUE #2326: `vp` cannot resolve `node` through a relative PATH entry — breaks Vite+'s own git hooks
## Summary

`vp` fails to resolve `node` when the `PATH` entry containing it is **relative** (e.g. `./node_modules/.bin`), erroring with:

```
error: Cannot find binary path for command 'node'
```

The same directory spelled as an absolute path works fine.

This isn't hypothetical: **Vite+'s own git-hook runner constructs exactly such a relative entry**, so `vp` inside a `pre-commit` hook fails whenever the ambient environment doesn't already provide a `node` of its own — which is the normal case for GUI git clients (VS Code, Tower, Fork).

## Reproduction

Standalone, no hooks involved:

```sh
mkdir vp-relative-path-repro && cd vp-relative-path-repro
echo '{"name":"repro","private":true,"type":"module"}' > package.json
npm install vite-plus@0.2.7 node@24

# absolute PATH entry — works
PATH="$PWD/node_modules/.bin:/usr/bin:/bin" ./node_modules/.bin/vp check
#   Found formatting issues in 1 file …

# relative PATH entry — fails
PATH="./node_modules/.bin:/usr/bin:/bin" ./node_modules/.bin/vp check
#   error: Cannot find binary path for command 'node'
```

The only difference between the two runs is `$PWD/node_modules/.bin` vs `./node_modules/.bin`. `./node_modules/.bin/node` exists and is executable in both, and the working directory is the project root in both.

### A relative entry appears to abort the search, not just be skipped

An absolute `node` **later** on `PATH` does not rescue it:

```sh
PATH="./node_modules/.bin:/usr/bin:/bin:/opt/homebrew/bin" ./node_modules/.bin/vp check
#   error: Cannot find binary path for command 'node'
```

`/opt/homebrew/bin/node` exists (v26.5.0) and is absolute, yet lookup still fails. So the relative entry seems to abort the scan rather than be ignored or resolved against the cwd.

## Why the git hooks hit this

`.vite-hooks/_/h`, generated by Vite+, builds its `PATH` from `$0`:

```sh
d="$(dirname "$(dirname "$(dirname "$0")")")"
export PATH="$d/node_modules/.bin:$PATH"
```

Git invokes hooks by a **relative** path (`.vite-hooks/_/pre-commit`), so `d` collapses to `.` and the exported entry is `./node_modules/.bin` — precisely the failing shape.

Committing from a terminal usually survives, because the login shell already exports an absolute `node` that wins the lookup. Committing from a GUI client, whose environment carries no `node`, dies before the hook body runs:

```
[STARTED] Cleaning up temporary files...
[COMPLETED] Cleaning up temporary files...
error: Cannot find binary path for command 'node'
VITE+ - pre-commit script failed (code 1)
```

## Expected

`vp` resolves `node` through a relative `PATH` entry (resolved against the current working directory, as the shell does), or at minimum skips the entry and continues scanning the rest of `PATH`.

## Actual

Lookup fails with `Cannot find binary path for command 'node'`, even when a valid absolute `node` appears later on `PATH`.

## Environment

- **vite-plus 0.2.7** (npm) — also reproduces on 0.2.5
- macOS 26.5.2, arm64
- node 24.18.0

Note: the standalone `vp` binary (`~/.vite-plus/bin/vp`) is **not** affected — it bundles its own node and never resolves one from `PATH`. Only the npm-distributed `#!/usr/bin/env node` build hits this, which is what a project's hooks actually run.

## Workaround

Prepend an absolute entry in the hook body, ahead of the runner's relative one:

```sh
PATH="$(pwd)/node_modules/.bin:$PATH"
```

## Suggested fix

Either or both:

1. In the binary lookup, resolve relative `PATH` entries against the cwd, and don't let one abort the remaining scan.
2. In `h`, absolutize `d` before exporting it.

RSS1102:
Fix relative `PATH` entries when resolving commands with a different working directory.

`which::which_in` resolves relative `PATH` entries against the Vite+ process cwd instead of the cwd provided for the command. This can cause `vp exec` to fail to find binaries when a relative entry such as `./node_modules/.bin` is used.

This change resolves ordinary relative `PATH` entries such as `tools`, `./tools`, and `../tools` against the command cwd while preserving PATH ordering and platform-specific path semantics.

It also preserves tilde expansion and Windows-specific paths such as drive-relative, root-relative, absolute, and UNC paths.

Snapshot and unit tests cover relative PATH resolution across Unix and Windows, including fallback lookup and Unix cwd paths containing the PATH separator.

Fixes #2326


fengmk2:
@RSS1102 Can you add a snapshot test to cover this bug fix change?

fengmk2:
@codex review

crates/vp_command/src/lib.rs
```diff
@@ -21,6 +21,15 @@ use vt_path::{AbsolutePath, AbsolutePathBuf, RelativePathBuf};
 
 mod ps1_shim;
 
+fn normalize_path_env(
+    path_env: &OsStr,
+    cwd: &AbsolutePath,
+) -> Result<OsString, std::env::JoinPathsError> {
+    std::env::join_paths(std::env::split_paths(path_env).map(|path| {
+        if path.is_absolute() || path.starts_with("~") { path } else { cwd.as_path().join(path) }
```
fengmk2: Need to consider compatibility issues on Windows, for example `C:foo` and `\foo` on Windows will be judged as relative, so the path after cwd join will be incorrect?

Alternatively, we could only check for explicit relative paths, supporting only paths that clearly start with `.` like `./foo/bar`, while not supporting paths like `foo/bar`.

fengmk2:
@codex review

fengmk2:
@codex review

crates/vp_command/src/lib.rs
```diff
@@ -21,6 +22,50 @@ use vt_path::{AbsolutePath, AbsolutePathBuf, RelativePathBuf};
...
+        !path.is_absolute()
+    }
+}
+
+fn resolve_bin_from_path_entry(
+    bin_name: &str,
+    path_entry: &Path,
+    cwd: &AbsolutePath,
+) -> Option<std::path::PathBuf> {
+    if path_entry.starts_with("~") || !is_plain_relative_path(path_entry) {
```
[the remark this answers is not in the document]
RSS1102: I don’t think an empty Windows `PATH` entry should be resolved against `cwd`. On Windows, `which` intentionally ignores empty entries produced by leading, trailing, or consecutive semicolons—for example, the empty entry between `;;` in `C:\tools;;C:\Windows\System32`. This matches Rust’s Windows command lookup behavior.

This PR only changes how non-empty relative entries such as `tools`, `./tools`, and `../tools` are resolved, so the empty-entry test is intentionally Unix-only.

RSS1102:
@fengmk2 pls review again when you have time

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_exec_relative_path_cwd/snapshots.toml
```diff
@@ -0,0 +1,11 @@
+[[case]]
+name = "command_exec_relative_path_cwd"
+vp = "local"
+skip-platforms = ["windows"]
```
fengmk2: Why skip windows?

RSS1102: sorry. I replaced the Unix-only `:` with `${PATH_SEPARATOR}` and removed the Windows skip, so the test now runs on Windows too.

crates/vp_command/src/lib.rs
```diff
@@ -39,15 +84,29 @@ pub fn resolve_bin(
...
 ) -> Result<AbsolutePathBuf, Error> {
+    let cwd = cwd.as_ref();
     let current_path;
     let path_env = if let Some(p) = path_env {
         p
     } else {
         current_path = std::env::var_os("PATH").unwrap_or_default();
         &current_path
     };
-    let path = which::which_in(bin_name, Some(path_env), cwd.as_ref())
```
fengmk2, line 49: One concern is that this changes a single which_in call into multiple calls, one for each PATH entry. 

Is there any way to keep a single `which_in` call, or would that require fixing which itself to resolve relative PATH entries against the provided cwd?

Like:

```rs
let path = normalized_entries
    .map(|entry| {
        if is_plain_relative_path(&entry) {
            cwd.join(entry)
        } else {
            entry
        }
    });

let path_env = std::env::join_paths(path)?;

which::which_in(bin_name, Some(path_env), cwd)
```

RSS1102: Yes, I agree that a single `which_in` call would be preferable.

However, normalizing the PATH locally still requires special handling for tilde entries and a fallback when `join_paths` fails, such as when a Unix `cwd` contains `:`. That would add more than ten lines while still retaining the current per-entry lookup as a fallback.

I’ve opened an upstream PR to address this in `which` itself:

- https://github.com/harryfei/which-rs/pull/128

I suggest keeping the current workaround with a TODO, then simplifying it to one `which_in` call after the upstream release. If you prefer to wait for that release, please let me know.

fengmk2: LGTM. After the PR is merged, you can create a follow-up issue to track it.

fengmk2 (approved):

PR #2355, opened 2026-08-06
refactor(core): remove redundant Vite reporter patch

jong-kyung:
## Summary

Make `brandVite()` the single source of truth for suppressing Vite's standalone version reporter line.

The removed build-time transform was a no-op for branded sources. It also could not patch the current unbranded upstream source because its anchor was stale


jong-kyung:
@codex review

fengmk2 (approved):

PR #2357, opened 2026-08-06
fix(editor): Remove overrides from default config

pweyck:
Do not set the options in the `.zed/settings.json`, to prevent a mismatch between `vp lint` and the language server.

These options are also not set in the vscode settings.


fengmk2:
@codex review

PR #2360, opened 2026-08-06
fix(test): emit NodeNext-compatible provider type specifiers

### LINKED ISSUE #2354: `vite-plus/test/browser-playwright` is error-typed under NodeNext
## Describe the bug

Importing `playwright` from `vite-plus/test/browser-playwright` causes Oxlint type-aware rules to treat `playwright()` as an error-typed value when the consumer uses NodeNext module resolution.

`vp check` reports `typescript(no-unsafe-return)` and `typescript(no-unsafe-assignment)`. Changing only the import to `@vitest/browser-playwright` makes the same configuration pass. Browser tests pass with either import, so the runtime implementation is unaffected.

The published `dist/test/browser-playwright.d.ts` rewrites Vitest bare module specifiers to extensionless relative ESM specifiers such as `./node`, `./browser`, and `./_at-vitest-browser`. TypeScript reports TS2834 or TS2835 for these declarations under NodeNext when library checking is enabled, while Oxlint resolves the provider as an error-typed value.

https://github.com/voidzero-dev/vite-plus/blob/7586d911b477d563d0e155ac103ee40c8c61e765/packages/cli/build.ts#L842-L851

I confirmed the cause by adding `.js` only to those generated relative specifiers. Both Oxlint diagnostics disappeared, and TypeScript declaration output resolved the return type as `import("vitest/node").BrowserProviderOption` instead of emitting an unqualified `BrowserProviderOption`.

This reproduces with Vite+ 0.2.7 and 0.2.8. The generated `browser-playwright.d.ts` files in both releases are byte-for-byte identical.

This differs from #1634: that issue concerned duplicate Vitest type identities when importing `@vitest/browser-playwright` directly. Here, the Vite+ re-export itself becomes error-typed under NodeNext.

The impact is that the `vite-plus/prefer-vite-plus-imports` rule recommends an import that makes `vp check` fail in a valid NodeNext configuration. The current workaround is to keep the direct import and suppress that rule.

A possible fix is to emit NodeNext-compatible `.js` specifiers from `relToShim` and add a regression test covering the generated provider declarations with NodeNext and type-aware Oxlint.

## Reproduction

https://github.com/eai04191/vite-plus-browser-playwright-nodenext-reproduction

## Steps to reproduce

1. Clone the reproduction repository.
2. Run `pnpm install --frozen-lockfile`.
3. Run `pnpm run check` and observe the two error-typed-value diagnostics.
4. Run `pnpm run check:direct`; the equivalent configuration using `@vitest/browser-playwright` passes.
5. Run `pnpm run typecheck` and `pnpm run typecheck:direct`; both pass with `skipLibCheck`.
6. Run `pnpm run install-browser`, `pnpm test`, and `pnpm run test:direct`; both browser tests pass.

## System Info

```shell
$ vp env current

$ vp --version
vp v0.2.8

Local vite-plus:
  vite-plus  v0.2.8

Tools:
  vite             v8.2.0
  rolldown         v1.2.2
  vitest           v4.1.10
  oxfmt            v0.61.0
  oxlint           v1.76.0
  oxlint-tsgolint  v7.0.2001
  tsdown           v0.22.14

Node.js v26.5.0
pnpm 11.1.2
macOS 15.7.7 arm64
```

## Used Package Manager

pnpm

## Logs

```shell
$ pnpm run check
$ vp check --no-fmt vite.config.ts

typescript(no-unsafe-return): Unsafe return of a value of type error.
  vite.config.ts:5:3
  return playwright({});

typescript(no-unsafe-assignment): Unsafe assignment of an error typed value.
  vite.config.ts:25:7
  provider: createPlaywrightProvider(),

Found 2 errors and 0 warnings in 1 file
```

## Validations

- [x] Checked open and closed issues and pull requests for the same bug.
- [x] Confirmed this is caused by Vite+'s generated declarations rather than Vitest or Oxlint.
- [x] The public reproduction is minimal and independently runnable.

byCodex

eai04191:
## Summary

- Emit `.js` extensions for relative shim specifiers in inlined browser-provider declarations so they resolve under NodeNext.
- Add regression coverage using the public `vite-plus/test/browser-playwright` types export.

Fixes #2354

## Validation

- `pnpm --filter vite-plus test src/__tests__/exports-map.spec.ts`
- `vp check`
- Verified `pnpm run check` and `pnpm run typecheck` in the public reproduction using a locally packed Vite+ build

byCodex

### DIFF
--- a/packages/cli/build.ts
+++ b/packages/cli/build.ts
@@ -842,7 +842,7 @@ async function writeInlinedProviderDts(
   const relToShim = (sub: string): string => {
     const r = relative(outDir, testDistDir).replaceAll('\\', '/');
     const prefix = r === '' ? '.' : r.startsWith('.') ? r : `./${r}`;
-    return `${prefix}/${sub}`;
+    return `${prefix}/${sub}.js`;
   };
   let result = upstream;
   for (const [bare, sub] of VITEST_TYPE_SPECIFIER_REWRITES) {
--- a/packages/cli/src/__tests__/exports-map.spec.ts
+++ b/packages/cli/src/__tests__/exports-map.spec.ts
@@ -91,6 +91,30 @@ describe('package.json exports map', () => {
     // vitest/config re-exports defineConfig / configDefaults — sanity-check one.
     expect(typeof cfg.defineConfig).toBe('function');
   });
+
+  it('browser-playwright declaration uses NodeNext-compatible relative specifiers', () => {
+    // Matches quoted relative module specifiers such as './node' and '../browser.js'.
+    const relativeModuleSpecifier = /(['"])(\.\.?\/[^'"]+)\1/g;
+    const pkg = JSON.parse(fs.readFileSync(cliPkgJsonPath, 'utf-8'));
+    const entry = (pkg.exports as Record<string, unknown>)['./test/browser-playwright'];
+    expect(isConditionObject(entry)).toBe(true);
+    const types = (entry as ExportConditions).types;
+    expect(types).toBeTypeOf('string');
+
+    const source = fs.readFileSync(path.resolve(cliPkgDir, types as string), 'utf-8');
+    const relativeSpecifiers = Array.from(
+      source.matchAll(relativeModuleSpecifier),
+      (match) => match[2],
+    );
+    expect(
+      relativeSpecifiers.length,
+      'declaration should contain relative specifiers',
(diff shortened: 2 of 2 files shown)


jong-kyung (approved):
LGTM 👍

jong-kyung:
@codex review

fengmk2 (approved):

PR #2330, opened 2026-08-05
ci(deps): handle upstream help diffs

liangmiQwQ:
This PR includes two primary parts, follow up to #2184 

1. Leave a comment under the PR that notices maintainers the help document differs
2. Let Claude Code update the help document when there is differences

🤖 Generated with Codex

PR #2365, opened 2026-08-06
docs(migrate): document manual installation

liangmiQwQ:
The manual migration setup is documented in the README, but is missing from the website migration guide.

This PR adds the dev dependency installation command and the npm, pnpm, and Yarn override examples, including why Vitest must match the bundled version.

🤖 Generated with Codex


docs/guide/migrate.md
```diff
@@ -69,6 +69,40 @@ After running the migration:
 - Run `vp test`
 - Run `vp build`
 
+## Manual Installation & Migration
+
+If you are manually migrating a project to Vite+, install these dev dependencies first:
+
+```bash
+npm install -D vite-plus @voidzero-dev/vite-plus-core@latest
```
fengmk2: no need to add `@latest` for `@voidzero-dev/vite-plus-core` too?

docs/guide/migrate.md
```diff
@@ -69,6 +69,40 @@ After running the migration:
...
+```
+
+You need to add overrides to your package manager so that other packages resolve the Vite+ versions: alias `vite` to `@voidzero-dev/vite-plus-core`, and pin `vitest` to the version Vite+ bundles (run `vp --version`) so the whole project shares a single Vitest copy with `vp test`. Without the `vitest` pin, a dependency or workspace package can pull a different Vitest than the bundled runner, splitting Vitest's internals (mocks, `expect`, runner state):
+
+```json
+"overrides": {
+  "vite": "npm:@voidzero-dev/vite-plus-core@latest",
+  "vitest": "4.1.10"
```
fengmk2, line 85: The vitest version is hardcoded here, and it will likely be overlooked when updating later.

docs/guide/migrate.md
liangmiQwQ: I just copy it from `README.md`, and I think that makes sense, so I will update both README and document here.

liangmiQwQ: I'm thinking about whether it is necessary to install `@voidzero-dev/vite-plus-core` explicitly

docs/guide/migrate.md
liangmiQwQ: I use a way that is similar to #2330, maybe it needs to be tested 🙏

(I actually quite dislike this, I think `<pm> update` should handle the full upgrade, I am finding a way to make Vite+ more elegant here)

fengmk2 (approved):

PR #2366, opened 2026-08-06
refactor(env): sort installed versions with node-semver

jong-kyung:
## Description

`vp env list` sorted installed Node.js versions with a hand-rolled `compare_versions` helper, even though `node-semver` is already a direct dependency of this crate and is used in nearby modules (`upgrade_check.rs`, `doctor.rs`, `pin.rs`).

This replaces the helper with a single `sort_by_cached_key` call using `node_semver::Version::parse`.


jong-kyung:
@codex review

fengmk2 (approved):

PR #2369, opened 2026-08-07
feat(pm): show download progress when fetching a package manager

### LINKED ISSUE #2351: Add download log when install package manager
### Description

at the moment, when vite+ finds package manager config on workspace, it silently starts downloading new package manager without any log or progress indicator. so users on a slow Wi-Fi connection can't tell whether it's still downloading or has simply hung.

### Suggested solution

Add logging. i'm not familiar with how vite+ download package manager. unfortunately, i don't know what kind of log would be ideal or feasible but if we could show download speed and progress, it would be good.

### Alternative

_No response_

### Additional context

_No response_

### Validations

semimikoh:
Closes #2351

### Problem

When a project's workspace declares a package manager (`packageManager`, lockfile, `devEngines`, etc.) that isn't installed yet, `vp` downloads it silently. On a slow connection there's no way to tell whether the download is progressing or has hung — no log line, no progress indicator, nothing until it either succeeds or errors out.

### Root cause

The package-manager download path (`crates/vp_pm_cli/src/request.rs`: `HttpClient::download_file`, `download_and_extract_tgz_with_hash`) only emits `tracing::debug!` calls, which aren't shown at the default log level. This is unlike the Node.js runtime download path (`crates/vp_js_runtime/src/download.rs`), which already renders an `indicatif` progress bar.

### Fix

Mirror the existing `vp_js_runtime` pattern in `vp_pm_cli`:

- `HttpClient::download_file` takes an optional `message`, and shows a byte progress bar (speed, ETA) once `Content-Length` is known, gated on `stderr` being a TTY and not running in CI — non-interactive/piped output stays unchanged.
- `download_and_extract_tgz_with_hash` threads the message through.
- Each package-manager download call site (npm-style tgz for npm/pnpm<12/yarn, bun's platform binary, pnpm>=12's native binary, and the pnpm hash-verification download) now passes a message like `"Downloading pnpm v10.0.0..."`.

### Testing

- `cargo check -p vp_pm_cli`, `cargo clippy -p vp_pm_cli --all-targets`, `cargo fmt -p vp_pm_cli --check`
- `cargo test -p vp_pm_cli --lib`: all download-path tests pass (`request::tests::*`, `package_manager::tests::test_download_*`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5824,7 +5824,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "anyhow",
  "append-only-vec",
@@ -6012,7 +6012,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_common"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "anyhow",
  "arcstr",
@@ -6048,7 +6048,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_dev"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "anyhow",
  "arcstr",
@@ -6069,7 +6069,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_dev_common"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "derive_more",
  "rolldown_common",
@@ -6081,7 +6081,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_devtools"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "blake3",
  "rolldown_devtools_action",
@@ -6094,7 +6094,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_devtools_action"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "serde",
  "ts-rs",
@@ -6102,7 +6102,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_ecmascript"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "arcstr",
  "oxc",
@@ -6114,7 +6114,7 @@ dependencies = [
 
 [[package]]
 name = "rolldown_ecmascript_utils"
-version = "1.2.2"
+version = "1.2.3"
 dependencies = [
  "memchr",
  "oxc",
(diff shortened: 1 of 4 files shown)


jong-kyung:
could we add some test coverage for this?

Cargo.lock
jong-kyung, line 1: are these `Cargo.lock` changes related to this PR?

jong-kyung:
@codex review

semimikoh:
Good call — added coverage in fa7adc3a:
- Extracted the progress-bar visibility check into a pure `should_show_progress(has_message, is_terminal, is_ci)` helper and unit tested its truth table directly (independent of the actual TTY/CI state the test runs under).
- Added `download_file`/`download_and_extract_tgz_with_hash` regression tests that pass a `message` and assert the downloaded/extracted content is unaffected.

jong-kyung:
thanks for adding these. I was expecting a snapshot test for the visible progress bar, but it looks difficult to test reliably because the bar is cleared on completion. On a closer look, these tests don't exercise the visible progress-bar path, so I don't think they provide meaningful coverage. Could we remove them for now? Also, could you drop the unrelated `Cargo.lock` changes?

semimikoh:
Fair point on both — done in fc957c07:
- Dropped the tests + the `should_show_progress` extraction that only supported them. Agreed they weren't exercising the actual visible-bar path, and reliably asserting on a TTY-gated, cleared-on-completion bar isn't worth the complexity here.
- Cargo.lock is back to just the `indicatif` addition — the rolldown version churn was from my local vendored checkout being ahead of what this branch's lockfile expected, unrelated to this change.

jong-kyung (commented):
`vp migrate` has its own spinner.  @fengmk2, Would it be okay if I triggered a `preview-build` build to check wherter the two spinners interfere with each other? Everything else looks good to me.

fengmk2:
`preview-build` cannot currently run under PRs in forked repos, but that's fine, we can merge first.

jong-kyung:
> `preview-build` cannot currently run under PRs in forked repos, but that's fine, we can merge first.

If we merge it, I'll check it in my local 🙇‍♂️

semimikoh:
Sounds good, thanks both! Happy to have this merged first and verify the spinner interaction with `vp migrate` after — let me know if anything comes up when you check locally.

jong-kyung (approved):

fengmk2 (approved):

fengmk2:
publish a preview build at https://github.com/voidzero-dev/vite-plus/pull/2373#issuecomment-5217539459

jong-kyung:
> publish a preview build at [#2373 (comment)](https://github.com/voidzero-dev/vite-plus/pull/2373#issuecomment-5217539459)

 Checked locally, no spinner conflicts. Works well 👍

PR #2384, opened 2026-08-09
fix(core): vp dev crashes on startup when experimental.bundledDev is enabled

lofcz:
The core build rewrites `CLIENT_ENTRY` and `ENV_ENTRY` to `dist/vite/client/`, but left `BUNDLED_DEV_CLIENT_ENTRY` at `dist/client/bundledDevClient.mjs`. Vite reads that path with `fs.readFileSync`, so `vp dev` with `experimental.bundledDev: true` fails at startup with `ENOENT`.

Rewrite the constant to the packaged layout and document it in `BUNDLING.md`. Add a PTY snapshot case (`command_dev_bundled_dev`) that starts `vp dev` with `experimental.bundledDev` enabled; it fails without the rewrite and passes with it.

`vp dev` with `experimental.bundledDev: true`, before:

```
error when starting dev server:
Error: ENOENT: no such file or directory, open '/path/to/project/node_modules/@voidzero-dev/vite-plus-core/dist/client/bundledDevClient.mjs'
    at Object.readFileSync (node:fs:484:20)
    at getHmrImplementation (…/dist/vite/node/chunks/node.js)
    at BundledDev.listen (…/dist/vite/node/chunks/node.js)
```

After:

```
  VITE+ v0.2.8

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
```


fengmk2:
@lofcz Thanks for the fix commit.

Could you please submit an issue first to describe the current problem?

lofcz:
Experimental dev bundling available in vite 8.1+ is broken due to the path being trivially wrong. With this fix, the feature can be used.

fengmk2 (approved):
@lofcz Thank you for your contribution, I will add a snapshot test before merging!

PR #2385, opened 2026-08-09
fix(cli): stamp package-manager lifecycle env for vp run scripts

### LINKED ISSUE #2317: vpr appears not to preserve the pnpm lifecycle environment, causing npm-run-all2 to fall back to npm
### Describe the bug

In a project configured to use pnpm, running a package script through `vpr` causes `npm-run-all2` / `run-p` to invoke child scripts with npm instead of pnpm.

The same script works correctly when launched with `pnpm run`.

The project declares pnpm as its package manager. When `run-p` is launched by pnpm, it correctly uses pnpm for the child scripts. When it is launched by `vpr`, it falls back to npm, which then fails with `EBADDEVENGINES` because the project only permits pnpm.

Expected behavior:

- `vpr check` should preserve enough package-manager lifecycle environment for child process runners to detect pnpm.
- Child scripts should be executed using pnpm, consistently with `pnpm run check`.

Actual behavior:

- `pnpm run check` succeeds.
- `vpr check` causes `run-p` to invoke npm for the child scripts.
- npm fails because `devEngines.packageManager` requires pnpm.

This appears to happen because the environment provided by `vpr` differs from the environment provided by `pnpm run`, particularly package-manager lifecycle variables such as `npm_execpath`.

A serial workaround that explicitly invokes pnpm works, but it loses parallel execution. In my project, the parallel command takes 3.94 seconds, while the serial workaround takes 6.35 seconds.

### Reproduction

https://github.com/bling-yshs/vue-demo/tree/archive/vp

### Steps to reproduce

1. Clone the reproduction repository.

2. Install dependencies:

   ```bash
   vp i
   ```

3. Run the parallel script directly through pnpm:

   ```bash
   pnpm run check
   ```

   All three child scripts complete successfully.

4. Run the same script through Vite+:

   ```bash
   vpr check
   ```

5. Observe that `npm-run-all2` invokes the child scripts using npm instead of pnpm. npm then exits with `EBADDEVENGINES` because the project declares pnpm in `devEngines.packageManager`.

### System Info

```shell
D:/code/Webstorm/vue-demo
$ vp env current

Environment:
  Version       24.19.0
  Source        .node-version
  Source Path   D:\code\Webstorm\vue-demo\.node-version
  Project Root  D:\code\Webstorm\vue-demo

Tool Paths:
  node  C:\Users\yshs\.vite-plus\js_runtime\node\24.19.0\node.exe
  npm   C:\Users\yshs\.vite-plus\js_runtime\node\24.19.0\npm.cmd
  npx   C:\Users\yshs\.vite-plus\js_runtime\node\24.19.0\npx.cmd

Package Manager:
  Name          pnpm
  Version       11.20.0
  Source Path   D:\code\Webstorm\vue-demo\package.json
  Project Root  D:\code\Webstorm\vue-demo
  Bin Path      C:\Users\yshs\.vite-plus\package_manager\pnpm\11.20.0\pnpm\bin\pnpm.cmd
D:/code/Webstorm/vue-demo
$ vp --version

vp v0.2.7

Local vite-plus:
  vite-plus  v0.2.7

Tools:
  vite             v8.1.5
  rolldown         v1.2.0
  vitest           v4.1.10
  oxfmt            v0.60.0
  oxlint           v1.75.0
  oxlint-tsgolint  v7.0.2001
  tsdown           v0.22.14

Environment:
  Package manager  pnpm v11.20.0
  Node.js          v24.19.0 (.node-version)
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

tarikermis:
Closes #2317

## What changed

`vp run` and `vpr` now stamp the two package-manager lifecycle variables child runners use for package-manager detection before `vt::Session::init` snapshots the environment:

- `npm_execpath`
- `npm_config_user_agent`

The helper resolves the real JavaScript CLI entry for npm, Yarn, and pnpm 11. Native pnpm 12 uses `pnpm.native` and reports `node/?`, matching the package manager instead of leaking the Node.js version that happens to host Vite+.

Bun remains deliberately unchanged. Expanding the emulation contract beyond the #2317 compatibility case needs its own verified use case.

## Why

`vp run` adds `node_modules/.bin` to `PATH`, but previously left the lifecycle environment empty. Child runners such as npm-run-all2 then fell back to npm inside pnpm projects, which can fail with `EBADDEVENGINES` when `devEngines.packageManager` requires pnpm.

## Verification

The CLI snapshot fixture now pins and executes Vite+'s real managed distributions instead of creating a fake pnpm layout:

- npm 10.9.8
- Yarn 4.17.1
- pnpm 11.20.0
- native pnpm 12.0.0
- Bun 1.3.14

The pnpm 11 and 12 cases prove the JavaScript/native split. The Bun case proves both lifecycle values remain undefined.

- Snapshot update run: 5/5 passed.
- Snapshot compare run: 5/5 passed.
- `cargo test -p vp_pm_cli lifecycle_env`: 10/10 passed.
- `pnpm build` and `just check` pass.
- Focused Clippy, Rust formatting, `git diff --check`, and the no-`.md.new` check pass.

## Notes

Cached tasks still filter these values in vite-task's post-filter environment. That separate half remains tracked in voidzero-dev/vite-task#692 and should land by updating the pinned vite-task revision, not by duplicating the logic here.

Best Regards, Tarik


fengmk2:
@tarikermis can you also add a new snapshot test to cover this bug fix?

tarikermis:
Good call, on it - adding a snapshot test for the fix now. Best Regards, Tarik

fengmk2:
@tarikermis I mean this snapshot tests https://github.com/voidzero-dev/vite-plus/tree/main/crates/vp_cli_snapshots/tests/cli_snapshots .
Sorry I didn't explain clearly.

tarikermis:
Ah my bad, got it - adding a CLI snapshot test in vp_cli_snapshots now. Best Regards, Tarik

crates/vp_pm_cli/src/snapshots/vp_pm_cli__lifecycle_env__tests__snapshot_pnpm_shim_fallback.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/vp_pm_cli/src/lifecycle_env.rs
+expression: "render_lifecycle_stamp(PackageManagerType::Pnpm, \"11.20.0\", &[])"
+---
```
fengmk2: @tarikermis can you revert these snapshots?

crates/vp_pm_cli/src/lifecycle_env.rs
```diff
@@ -0,0 +1,509 @@
...
+                    &self.version,
+                    context.node_version.as_deref(),
+                )),
+            ),
+            ("INIT_CWD", context.init_cwd.as_path().as_os_str().to_os_string()),
+        ];
+
+        if let Some(node_execpath) = &context.node_execpath {
+            vars.push(("npm_node_execpath", node_execpath.as_os_str().to_os_string()));
+            vars.push(("NODE", node_execpath.as_os_str().to_os_string()));
```
fengmk2, line 131: Is there source documentation for the naming of these environment variables? Or reference code? Need to add code comments to explain them.

tarikermis:
Sure - reverting the insta snapshots and adding source comments explaining the env var naming. Best Regards, Tarik

wan9chi (approved):

crates/vp_pm_cli/src/lifecycle_env.rs
```diff
@@ -0,0 +1,403 @@
...
+        context: &LifecycleEnvContext,
+    ) -> Vec<(&'static str, OsString)> {
+        if matches!(self.client, PackageManagerType::Bun) {
+            return Vec::new();
+        }
+
+        let mut vars = vec![
+            ("npm_execpath", self.lifecycle_exec_path().as_path().as_os_str().to_os_string()),
+            (
+                "npm_config_user_agent",
```
fengmk2, line 127: Do we need to emulate the package manager lifecycle environment this broadly to fix #2317? The original compatibility issue is primarily package-manager detection through `npm_execpath`. Every additional emulated variable becomes a compatibility contract that we need to keep aligned with npm/pnpm/Yarn. Could we keep this helper minimal unless another variable is required by an actual compatibility case?

simulacre7:
While digging into #2317 myself I traced the full spawn path through the pinned vite-task rev (`d05b1dc`) and hit one gap in this approach that the snapshot fixture can't catch — sharing since it seems cheap to address before merge.

**Cache-enabled tasks strip the stamped env again.** Stamping the process env before `Session::init` works for plain `package.json` scripts because they default to `cache: scripts: false`, and the full session env flows to the child. But when a task is cache-enabled (`cache: { scripts: true }`, or `run.tasks` from `vite.config.ts`, which cache by default), `plan_spawn_execution` filters the spawn env down to the configured patterns first:

- `vt_plan/src/plan.rs` → `EnvFingerprints::resolve(&mut spawn_envs, &cache_config.env_config)` keeps only names matching `untracked_env` + `env`;
- `DEFAULT_UNTRACKED_ENV` (`vt_graph/src/config/mod.rs`) contains no lowercase `npm_*` names (only `NPM_CONFIG_STORE_DIR`), and matching is case-sensitive on Unix (`vt_glob/src/env.rs`).

So for a cached task, `npm_execpath` / `npm_config_user_agent` / `npm_node_execpath` / `INIT_CWD` get silently dropped, and `npm-run-all2` falls back to npm again — exactly the #2317 symptom, now only in the cached case. The `VP_RUN` marker had the same problem and vite-task solves it by re-inserting the marker *after* the filter (`vt_plan/src/plan.rs`, next to the `MARKER_ENV_NAME` insert, with a comment explaining this exact reasoning).

Possible directions, both upstream in vite-task: add the lifecycle names to `DEFAULT_UNTRACKED_ENV`, or re-insert them post-filter like the marker. Either way it might be worth a note in this PR (or a follow-up issue) so the cached-task case doesn't regress the fix.

Caveat: this is from reading the pinned vite-task sources, not from running a cached task against this branch — happy to be corrected if the filtering doesn't apply where I think it does.

simulacre7:
Filed the upstream half as voidzero-dev/vite-task#692 (post-filter re-insertion for cached tasks + per-task `npm_lifecycle_event`), so the cached-task gap has a tracking home either way.

tarikermis:
Thanks for tracing this through the cached-task path. You’re right: the current CLI snapshot proves the lifecycle env for normal package scripts, but it does not cover the post-filter environment used by cached tasks.

Since the actual fix belongs in vite-task and #692 now tracks the post-filter re-insertion, I’d keep this PR scoped to the uncached path rather than duplicating that logic in Vite+. The cached-task case remains a known gap until the pinned vite-task revision includes that fix.

Best Regards, Tarik

tarikermis:
Thanks for tracing this through the cached-task path. You’re right: the current CLI snapshot proves the lifecycle env for normal package scripts, but it does not cover the post-filter environment used by cached tasks.

Since the actual fix belongs in vite-task and #692 now tracks the post-filter re-insertion, I’d keep this PR scoped to the uncached path rather than duplicating that logic in Vite+. The cached-task case remains a known gap until the pinned vite-task revision includes that fix.

Best Regards, Tarik

simulacre7:
Agreed — scoping this PR to the uncached path is the right call, and duplicating the re-insertion in Vite+ would just race the upstream fix. voidzero-dev/vite-task#692 tracks the post-filter re-insertion (and per-task `npm_lifecycle_event`, which can't be set from the Vite+ side at all); I've offered a PR there once maintainers confirm the direction, so the gap should close by bumping the pinned vite-task revision afterwards.

tarikermis:
Fair point - narrowed it. The stamp is now just npm_execpath + npm_config_user_agent, the two channels child runners actually use for package-manager detection (npm-run-all keys on npm_execpath; ni and friends parse the user agent). INIT_CWD, npm_node_execpath and NODE are dropped until a real compatibility case needs them. Also merged current main, so the conflicts are gone. The cached-task half stays upstream in vite-task#692 as discussed with simulacre7. Best Regards, Tarik

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/vite_task_lifecycle_env/snapshots.toml
```diff
@@ -0,0 +1,15 @@
...
+comment = """
+Regression test for #2317: `vp run` stamps the package-manager lifecycle env
+(npm_execpath, npm_config_user_agent) for package.json scripts, so
+child tooling like npm-run-all detects pnpm instead of falling back to npm.
+Pre-fix every variable printed `(undefined)`. The fake managed pnpm install
+under VP_HOME keeps the case offline; the script normalizes the user-agent
+platform/arch tail that suite redaction does not mask.
+"""
+steps = [
+  { argv = ["node", "scripts/setup-fake-pnpm.cjs"], snapshot = false },
```
fengmk2: We should test with real pnpm, yarn, bun, npm instead of using fake simulations. Additionally, pnpm needs to be tested on v11 and v12, the two major versions, as they involve changes in native pnpm logic.

simulacre7:
Data point in support of covering both pnpm majors with real binaries — observed lifecycle env from corepack-fetched pnpm, same probe script, same machine:

**pnpm 11.20.0** (JS entry):
```
npm_execpath = …/corepack/v1/pnpm/11.20.0/bin/pnpm.mjs
npm_config_user_agent = pnpm/11.20.0 npm/? node/v22.22.3 darwin arm64
npm_lifecycle_event = probe
```

**pnpm 12.0.0** (native binary):
```
npm_execpath = …/corepack/v1/pnpm/12.0.0/pnpm-native
npm_config_user_agent = pnpm/12.0.0 npm/? node/? darwin arm64
npm_lifecycle_event = probe
npm_node_execpath = <node used to run the script>
```

Two consequences that make the v11/v12 split load-bearing rather than cosmetic:

1. `npm_execpath` changes *kind*, not just value: a JS entry vs a native executable. Child runners branch on exactly this — npm-run-all2 execs `node $npm_execpath` when the path looks like JS and execs it directly otherwise — so the two majors exercise different downstream paths, and a stamp that always points at a `.cjs` would be wrong on v12.
2. The v12 user agent reports `node/?`, so anything parsing the agent for a node version (the `ni` family) sees different input per major.

Both observations transfer to the vite-task side too (voidzero-dev/vite-task#692 tracks the cached-task half) — the eventual `npm_execpath` derivation there needs the same ≤11 `.cjs` / ≥12 native branch.

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/vite_task_lifecycle_env/snapshots.toml
tarikermis: Done in `57eebe24`.

The fake pnpm layout is gone. The PTY fixture now pins and executes the real managed npm 10.9.8, Yarn 4.17.1, pnpm 11.20.0, native pnpm 12.0.0, and Bun 1.3.14 binaries before checking `vp run`.

The pnpm 11 snapshot uses `pnpm.cjs`; pnpm 12 uses `pnpm.native` and now reports `node/?` instead of the Node version hosting Vite+. Bun's real-binary case confirms the two values stay undefined by design.

Both the update and compare snapshot runs pass 5/5. The lifecycle unit tests pass 10/10, and the build, workspace check, Clippy, and Rust formatting are green.

Best Regards, Tarik

fengmk2 (approved):

fengmk2 (approved):
LGTM!

Later we can add Vite+ version information to `npm_config_user_agent`, for example `npm_config_user_agent=vp/<version> pnpm/<version> …`

fengmk2:
@tarikermis ci fail

PR #2386, opened 2026-08-09
feat(cli): make download timeout configurable and more forgiving

### LINKED ISSUE #2370: Download Node Timeout
### Description

For devices connected to slow internet speeds or flaky connections, a real time download speed less than 1MBps downloading node versions times out a lot, and makes it very hard to install and setup, I had to do it maybe 3 to 4 times and turn off all the other things that are installing while vp migrates.

In third world countries internet is very slow sadly and a bit flaky.

### Suggested solution

Increase the timeout and make it more forgiving.

### Alternative

_No response_

### Additional context

_No response_

### Validations

tarikermis:
## What & why

Closes #2370.

Node.js runtime tarballs (~25–50 MB) and package-manager tarballs are downloaded through the process-wide shared HTTP client, whose per-request total timeout is a hardcoded 2 minutes (`crates/vp_shared/src/http.rs`). reqwest's timeout covers the entire body stream, so on connections slower than ~250 KB/s a healthy Node.js download is aborted mid-transfer, every retry starts from scratch, and `vp env install` / `vp migrate` effectively cannot complete. There was also no way to override the timeout.

This change gives large file downloads their own per-request timeout instead of raising the shared default:

- New `vp_shared::download_timeout()` — 10-minute default, overridable with the `VP_DOWNLOAD_TIMEOUT_SECS` environment variable (positive integer seconds; a set-but-invalid value warns and falls back to the default).
- Applied per request in the two `download_file` implementations: `crates/vp_js_runtime/src/download.rs` (Node.js runtimes — all `download_runtime*` callers funnel through it) and `crates/vp_pm_cli/src/request.rs` (npm/yarn/pnpm/bun tarballs — all go through `HttpClient::download_file`).
- The shared client's 2-minute default is unchanged, so small metadata fetches (version indexes, checksums, registry JSON) still fail fast when stuck.
- Documented in `docs/guide/installer-env-vars.md` alongside the other runtime variables.

## Verification

- Root cause confirmed on current `main`: both download paths issue `client.get(url).send()` against the shared client with the hardcoded 2-minute total timeout; a 30–50 MB archive needs a sustained ≥ ~300 KB/s to finish inside that window.
- `cargo check -p vp_shared -p vp_js_runtime -p vp_pm_cli --all-targets --locked` — clean.
- `cargo clippy -p vp_shared -p vp_js_runtime -p vp_pm_cli --all-targets --all-features -- --deny warnings` (same `-A` set as `just lint`) — clean.
- `cargo fmt --all` — applied, no drift.
- `cargo test -p vp_shared -p vp_js_runtime -p vp_pm_cli` — all suites pass (962 passed, 0 failed; new unit tests cover the env parsing: default, valid override, whitespace trimming, and invalid/blank/zero/negative fallback).
- Existing `httpmock`-based download tests in `vp_pm_cli` still pass, exercising the patched request path end to end.

## Notes / limitations

- I could not reproduce a real slow-network Node.js download locally; behavior was verified by code inspection plus the unit/integration tests above. No snapshot tests were run (JS build not involved in this change).
- Scope: `vp upgrade` and the standalone installer download the `vp` tarball itself via `get_bytes`, which still uses the shared 2-minute default; extending the knob there would touch a different API and is left for a follow-up if desired.
- Trade-off worth knowing: the timeout is per attempt, so worst-case wall time with the existing retries grows from ~8 to ~40 minutes. Each attempt shows a progress bar, and a fresh retry on a flaky link often lands a faster connection, so this was kept consistent with the existing retry design.
- An alternative design considered (raised during review): an idle/read timeout (abort only when no bytes arrive for N seconds) instead of a total budget. reqwest's per-request API only offers a total timeout, and reworking the stream loops for idle semantics is a larger behavioral change; the total timeout + env override implements the issue's suggested solution with a contained diff. Happy to explore the idle-timeout variant as a follow-up.

## AI disclosure

This PR was prepared with AI assistance (code authored and tested by an AI agent, reviewed by a second independent AI reviewer pass whose findings were addressed or noted above).


docs/guide/installer-env-vars.md
```diff
@@ -100,6 +100,16 @@ These variables configure the installed Vite+ CLI. `VP_HOME` (above) also applie
 - **Default**: None (verification enabled)
 - **Details**: [Node.js Signature Verification](/guide/env#node-js-signature-verification)
 
+### `VP_DOWNLOAD_TIMEOUT_SECS`
+
+- **Purpose**: Per-request timeout, in seconds, for large downloads such as Node.js runtimes and package-manager tarballs
+- **Default**: `600` (10 minutes)
```
fengmk2, line 107: Where does the default ten-minute reference value come from? Could you refer to what the default values for this parameter are in other Node.js Version Managers?

fengmk2:
@codex review

fengmk2:
@codex review

fengmk2 (approved):
@tarikermis Thanks!

docs/guide/installer-env-vars.md
```diff
@@ -100,6 +100,17 @@ These variables configure the installed Vite+ CLI. `VP_HOME` (above) also applie
 - **Default**: None (verification enabled)
 - **Details**: [Node.js Signature Verification](/guide/env#node-js-signature-verification)
 
+### `VP_DOWNLOAD_TIMEOUT_SECS`
```
fengmk2: I’d prefer `VP_DOWNLOAD_TIMEOUT` over `VP_DOWNLOAD_TIMEOUT_SECS`, so the value can evolve to support duration formats like `10m` without baking the unit into the variable name. Plain integer values could default to seconds.

tarikermis:
Good point - renaming to VP_DOWNLOAD_TIMEOUT and keeping plain ints as seconds. Best Regards, Tarik

fengmk2 (approved):

PR #2396, opened 2026-08-10
chore: remove unused build:src task

jong-kyung:
## Summary

Removes the undocumented `build:src` task from the root Vite+ configuration. The existing `build` script already covers these source packages as part of the broader build flow.

The task appears to be unused. If the intended setup is for the root `build` script to delegate to this task, please let me know. I can keep the task and update the script to call `vpr build:src` instead.


fengmk2 (approved):

jong-kyung:
Sorry for all the PRs lately 😅 Really appreciate you taking the time to review them🙇‍♂️

PR #2397, opened 2026-08-10
ci: fix the publishing workflow after its first real runs

fengmk2:
Two fixes to `main`, both found by the first real runs of the publishing workflow. It could not be exercised before merge, because `workflow_run` only fires for workflow files already on the default branch.

**First, the good news: the design works.** PR #2328 published end to end through the new path with an OIDC token, no admin token involved. `authorize`, `Pkg Preview`, and the sticky comment all succeeded, and `commit.a7180fa85c06fad48` is on the bridge:

```
commit.a7180fa85c06fad48 | pr: .../pull/2328 | at: 2026-08-10T07:35:31.617Z
```

## 1. Fork PRs could not be resolved at all

#2391 (from `liangmiQwQ`) failed in `authorize` with `no open PR of voidzero-dev/vite-plus has head c7e51be…` while that PR was open with exactly that head.

`listPullRequestsAssociatedWithCommit` returns **empty** for a fork PR's head commit. Confirmed against the live API:

| commit | result |
| --- | --- |
| #2387 head (same-repo) | returns `#2387` |
| #2391 head (fork) | **empty** |

So it worked for every case reachable before merge and failed for the only case this feature exists for. `workflow_run.pull_requests` is empty for forks too, which is what sent me to the commit endpoint originally — I swapped one fork-blind source for another.

Now resolves via `pulls?state=open&head=<head_owner>:<head_branch>`, both GitHub-signed payload fields. The head-sha match is a separate step so the message distinguishes "no such PR" from "the PR moved on":

```
fork PR 2391 (real failure) -> OK: #2391 labeled=true fork=true
stale head                  -> FAIL: PR #2391 now at c7e51be, built 0000000
no such branch              -> FAIL: no open PR from liangmiQwQ:does-not-exist
```

I re-checked the rest of the publishing workflow for the same blind spot. Everything else keys off the PR number or the run id, which are base-repo objects and fork-safe: the post-approval `pulls.get` re-check returns correct state, head and labels for #2391, and the artifact download and the `listWorkflowRunArtifacts` precondition both see that run's 148MB `bridge-packages`.

## 2. The Docker gha cache broke the image push

```
#14 exporting to GitHub Actions Cache
#14 ERROR: error writing layer blob: failed to reserve cache
#13 exporting to image ... CANCELED
```

The cache export is fatal to the build, so it cancelled the push. I added this in the cleanup pass; it broke the job it was meant to speed up, and #2328's npm preview published while its Docker image did not.

Reverted rather than repaired. Making it work needs `actions: write` on the one job that installs and executes the preview package, which is the job SR-5 says to keep unprivileged, and this was the only `type=gha` usage in the repo so there was no working precedent. It was saving 60-90s of apt on a path that already waits on a human approval measured in minutes to days.

## 3. Terminology

"Trusted leg" and "build leg" were my own coinage and meant nothing to a reader who was not in the design conversation. The two workflows are now described as **the build workflow** and **the publishing workflow**, and where trust was the point the property is stated rather than encoded in a name.

This also surfaced something worth fixing later: `publish-preview.yml` is named "Publish preview build" and no longer publishes anything. Renaming it is the real fix, but the publishing workflow matches it by `name:`, so that has to be a coordinated change. The header says so outright for now.

The same terminology fix for the RFC and bridge docs is voidzero-dev/pkg-pr-registry-bridge#93, which also corrects SR-1 for the fork-blind endpoint above.

## After merging

Re-label #2391 to get the first genuine fork preview.

### DIFF
--- a/.github/workflows/publish-preview-register.yml
+++ b/.github/workflows/publish-preview-register.yml
@@ -79,28 +79,58 @@ jobs:
         id: check
         env:
           HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+          HEAD_OWNER: ${{ github.event.workflow_run.head_repository.owner.login }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
           RUN_ID: ${{ github.event.workflow_run.id }}
         with:
           script: |
             const headSha = process.env.HEAD_SHA;
             const { owner, repo } = context.repo;
 
-            // Resolve the PR from the commit. workflow_run.pull_requests is
-            // empty for fork PRs, which is why this goes to the API.
-            const pulls = await github.paginate(
-              github.rest.repos.listPullRequestsAssociatedWithCommit,
-              { owner, repo, commit_sha: headSha },
-            );
+            // Resolve the PR by its HEAD BRANCH, not by commit association.
+            //
+            // This used listPullRequestsAssociatedWithCommit, which returns
+            // EMPTY for a fork PR's head commit -- the exact case this workflow
+            // exists for -- while working correctly for same-repo PRs. So it
+            // looked right until the first fork PR arrived (#2391) and failed
+            // with "no open PR has head <sha>" for a PR that was open with
+            // precisely that head. `workflow_run.pull_requests` is also empty
+            // for forks, which is what sent me to the commit endpoint in the
+            // first place; both are fork-blind.
+            //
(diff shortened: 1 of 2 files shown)


fengmk2:
@codex review

liangmiQwQ:
Thx ❤️

wan9chi (approved):

PR #2399, opened 2026-08-10
fix(pm): avoid implicit project writes

### LINKED ISSUE #1916: bug: vp i pins npm@11 into devEngines on low Node, breaking manual npm with "Cannot find module node:path"

### LINKED ISSUE #2374: Do not update `package.json` when running `vpr`
### Description

Running a script with `vpr` can add package manager metadata to `package.json`.

This is unexpected when I only want to run a script, especially in someone else’s repository. Coding agents also freak out with this behaviour, Vite+ edits `package.json`, the agent reverts the unexpected change, and Vite+ adds it again on the next `vpr` command.

Vite+ should not automatically edit files as side effects.

### Suggested solution

Do not add `packageManager` or `devEngines.packageManager` metadata as a side effect of running `vpr`/`vp run`.

Pinning a package manager should require an explicit user action, such as a dedicated `vp pm pin` command.

### Alternative

_No response_

### Additional context

I want to fix this if we go with this proposal 🫡

### Validations

### LINKED ISSUE #2375: Enhance Vite+ package manager management
There is room to improve the current package-manager management system.

1. Vite+ currently provisions package managers for both the global and local CLI.

   - This is expected for the global CLI.
   - The local CLI can also run without a global Vite+ installation. Package-manager commands may therefore create Vite+-managed state even when Vite+ is installed only in a project.
   - Several Vite+ features need a package manager to be available. The old implementation mixed project detection, downloads, `PATH` changes, and manifest updates, which made behavior inconsistent and caused problems such as #2374 and #1916.

2. Corepack was introduced in #1808 to expose package-manager binaries.

   - Corepack handles npm, pnpm, and Yarn, while Vite+ also supports package managers such as Bun.
   - Vite+ already has its own package-manager version resolution and download logic, so Corepack creates overlapping concepts and can download a second copy of the same manager.
   - Corepack is unmaintained and does not support newer managers such as pnpm 12.

### Proposed package-manager refactor plan

The refactor is split into three dependent but independently mergeable parts:

1. **Replace Corepack with Vite+-owned shims**

   PR #2391 removes the Corepack integration. `vp env setup` directly provides shims for npm, pnpm, Yarn, and Bun.

2. **Manage package managers through `vp env`**

   PR #2398 extends `vp env` from Node.js-only management to a complete JavaScript environment: Node.js plus package-manager components. Commands such as `pin`, `use`, `install`, `clean`, and `current` can manage either component.

3. **Resolve and run managed package managers without implicit project writes**

   PR #2399 keeps ordinary package-manager commands on Vite+ managed binaries. A command detects the manager and version from project metadata, resolves or downloads that version, and executes the managed binary directly. The package manager does not need to be preinstalled on `PATH`.

   Detection is non-mutating. Lockfiles, configuration files, defaults, and `devEngines.packageManager` ranges can select a manager for the current command without adding or rewriting package-manager metadata in `package.json`. Projects can create or remove an explicit declaration with `vp env pin` and `vp env unpin`.

   Project-local dependency commands such as `vp install` and `vp add` require an existing `package.json`. Managed global operations continue to use the Vite+ global package store.

The goal is to separate responsibilities that were previously mixed together:

- detecting which package manager and version a project uses;
- provisioning and selecting managed versions through `vp env` and shims;
- executing ordinary package-manager commands without implicit project mutation.

This addresses the unexpected `package.json` mutation reported in #2374, avoids the incompatible implicit npm pin reported in #1916, and removes the overlap between Corepack and Vite+ package-manager management. Unlike the earlier PATH-based proposal, it also preserves version-correct commands when no system package manager is installed; local-only Vite+ usage may still provision Vite+-managed state when a package manager is needed.

The change is technically breaking, but common workflows should remain mostly unchanged. The main affected cases are Corepack-specific usage, scripts consuming `vp env --json`, and workflows that relied on package-manager commands to create or update `package.json` implicitly.

liangmiQwQ:
Close #2375
Close #2374
Close #1916

Package-manager commands resolve the manager and version from project metadata, then execute the managed binary for that exact version.

Package-manager detection is non-mutating. Lockfiles, config files, defaults, and `devEngines.packageManager` ranges are resolved for the current command without changing `package.json`. Projects can declare or remove a package-manager version explicitly with `vp env pin` and `vp env unpin`.

Project-local dependency commands such as `vp install` and `vp add` require an existing `package.json`. Managed global operations continue to use the Vite+ global package store.

🤖 Generated with Codex


liangmiQwQ:
@codex review

liangmiQwQ:
@fengmk2 I'm ready for this.

docs/guide/install.md
```diff
@@ -19,7 +19,9 @@ Vite+ detects the package manager from the workspace root in this order:
 9. `bunfig.toml`
 10. `yarn.config.cjs`
 
-If none of those files are present, `vp` falls back to `pnpm` by default. Vite+ automatically downloads the matching package manager and uses it for the command you ran. When detection comes from lockfiles or config files, the resolved version is written to `devEngines.packageManager` so future runs are deterministic; projects that already declare `packageManager` or `devEngines.packageManager` are left as-is.
+If none of those files are present, `vp` falls back to `pnpm` by default. Vite+ automatically downloads the matching package manager and uses it for the command you ran, but package-manager detection never rewrites `package.json`. Use `vp env pin <package-manager>@<version>` when the project should declare an exact version explicitly.
+
+Project-mutating commands such as `vp install` and `vp add` require an existing `package.json`; they do not create one automatically.
```
fengmk2: Will it report an error directly if package.json doesn't exist? In cases like this, what is the logic of `pnpm install` and `pnpm add`? It should be consistent with pnpm's logic. Reporting an error directly seems to go against user expectations.

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pm_no_package_json/snapshots/command_pm_no_package_json.local.md
```diff
@@ -7,7 +7,7 @@ should show friendly error
 **Exit code:** 1
 
 ```
-No package.json found.
+error: Package not found in workspace: `<workspace>`
```
fengmk2: The previous error message would accurately indicate that the `package.json` file does not exist. The current message causes users to wonder what `Package not found` means. The error message needs to be improved, for example: `error: package.json not found in workspace: xxx`

docs/guide/install.md
liangmiQwQ: I design in this way because the old `vp install` creates a `package.json` when there is no package.json found and wrote some config like `"type": "model"`, but since this PR's goal is to make these commands stop modifying `package.json`, I think your suggestion makes snese.

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pm_no_package_json/snapshots/command_pm_no_package_json.local.md
liangmiQwQ: This will be skipped because there won't be an error message any longer after https://github.com/voidzero-dev/vite-plus/pull/2399#discussion_r3895373293 gets resolved.

docs/guide/install.md
liangmiQwQ: Updated in [c384bcf](https://github.com/voidzero-dev/vite-plus/pull/2399/commits/c384bcf8d1c2c5d7eebcf9aa9ac43a6d9c23049e)

fengmk2 (approved):

fengmk2:
Need to fix ci fail before merge.

PR #2411, opened 2026-08-11
fix(env): resolve partial `VP_NODE_VERSION` overrides

jong-kyung:
## Summary

`VP_NODE_VERSION=22` now works as documented for shim-dispatched commands such as `vp env exec node -v`.

The shim resolves partial versions to an exact Node.js release before checking or downloading the runtime. A regression test verifies that `22` resolves to an exact `22.x.y` version while preserving `VP_NODE_VERSION` as the resolution source.

**Before**
image

**After**
image

Ref: 
https://github.com/voidzero-dev/vite-plus/blob/2d5f1927f9204629122a3f2aa418da21cef37a50/docs/guide/installer-env-vars.md?plain=1#L86-L94


jong-kyung:
@codex review

fengmk2 (approved):

PR #2417, opened 2026-08-11
fix(create): initialize git once for new monorepos

jong-kyung:
## Description

Creating a new monorepo with `vp create vite:monorepo` handled git initialization through two separate code paths, which caused three related problems:

- In verbose interactive mode without `--git`/`--no-git`, the user was prompted twice ("Initialize a git repository?" from `resolveGitInit`, then "Initialize git repository:" inside the monorepo branch).
- When the two answers differed, the second answer was silently overridden because the later `initGitRepository` call keyed off the first prompt's result.
- On the common (compact) path, `git init` ran twice: once via a direct `spawn.sync` right after scaffolding and again via `initGitRepository` after the monorepo rewrite.

## Fix

- Remove the duplicate prompt; `resolveGitInit` already asks up front before scaffolding.
- Keep a single `initGitRepository` call, placed before the hook preflight so `shouldSkipStagedMigrationForHooks` sees the destination's intended git root (same reasoning as the standalone path).
- Move the failure warning and stderr output into the shared `initGitRepository` helper so both the monorepo and standalone paths report a failed `git init`.

### DIFF
--- a/packages/cli/src/create/bin.ts
+++ b/packages/cli/src/create/bin.ts
@@ -3,7 +3,6 @@ import path from 'node:path';
 import { styleText } from 'node:util';
 
 import * as prompts from '@voidzero-dev/vite-plus-prompts';
-import spawn from 'cross-spawn';
 import mri from 'mri';
 
 import { vitePlusHeader } from '../../binding/index.js';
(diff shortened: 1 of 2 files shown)


jong-kyung:
@codex review

fengmk2 (approved):

Issue #2309, opened 2026-08-03
`vp up` in monorepo changes vite dependency in package.json from catalog to exact version


maxdzin:
### Describe the bug

In just-created vp monorepo with pnpm, `package.json` has dev dependency `"vite": "catalog:"`.

By launching `vp up`, dependencies are updated properly in `pnpm-workspace.yaml` and `vite` is set there as:
```
catalog:
  ...
  vite: npm:@voidzero-dev/vite-plus-core@0.2.7
  ...
```
but `package.json` `vite` dependency is changed from `catalog:` to `npm:@voidzero-dev/vite-plus-core@0.2.7`

**What is expected:**
After `vp up` `package.json` `vite` dependency should be preserved as `"vite": "catalog:"`.

**What happening actually:**
After `vp up` `package.json` `vite` dependency is changed from `"vite": "catalog:"` to `"vite": "npm:@voidzero-dev/vite-plus-core@0.2.7"`.

### Reproduction

not required

### Steps to reproduce

1. `vp create vite:monorepo` - select `pnpm` as package manager and complete installing.
2. Check `package.json`: `vite` dependency is set as `"vite": "catalog:"` there.
3. Run `vp up`.
4. Check `package.json`: `vite` dependency is changed now to `"vite": "npm:@voidzero-dev/vite-plus-core@0.2.7"`, but should be preserved as `"vite": "catalog:"`.

### System Info

```shell
VP current env:

VITE+ - The Unified Toolchain for the Web

Environment:
  Version       24.18.1
  Source        engines.node
  Source Path   /home/shanti/workspace/code/vite-plus-monorepo-deps-issue/package.json
  Project Root  /home/shanti/workspace/code/vite-plus-monorepo-deps-issue

Tool Paths:
  node  /home/shanti/.vite-plus/js_runtime/node/24.18.1/bin/node
  npm   /home/shanti/.vite-plus/js_runtime/node/24.18.1/bin/npm
  npx   /home/shanti/.vite-plus/js_runtime/node/24.18.1/bin/npx

Package Manager:
  Name          pnpm
  Version       11.18.0
  Source        devEngines.packageManager
  Source Path   /home/shanti/workspace/code/vite-plus-monorepo-deps-issue/package.json
  Project Root  /home/shanti/workspace/code/vite-plus-monorepo-deps-issue
  Bin Path      /home/shanti/.vite-plus/package_manager/pnpm/11.18.0/pnpm/bin/pnpm

---

[...]

PR #2432, opened 2026-08-12
fix(pm): keep Yarn `catalog:` references intact through `vp up`

fengmk2:
The Yarn variant of #2309, stacked on #2425. Under Yarn 4.10+ a migrated project references the toolchain pins with `catalog:`. `yarn up <name>` rewrites the manifest spec of every named package, so `vp up vite` turned `"vite": "catalog:"` into `"vite": "^8.2.1"` (upstream Vite) and `vp up vite-plus` wrote a concrete range. The catalog reference is destroyed either way. Verified on Yarn 4.12.0 and 4.18.0: Yarn has no upstream fix, and `yarn up 'vite@catalog:'` fails with "isn't supported by any available resolver", so the guard lives in vp.

Dispatch now reads the catalog package names from the workspace root's `.yarnrc.yml` (`catalog` plus every named `catalogs` entry) and hands them to command resolution as data, so resolution stays free of filesystem access. The Yarn Berry update resolver skips catalog-pinned bare names with a warning and resolves to a no-op when nothing else was requested:

```
warn: Skipped vite: the Yarn catalog pins its version, and `yarn up` would overwrite the `catalog:` reference. Edit the catalog entry in .yarnrc.yml, or run `vp migrate` when Vite+ manages the pin.
```

A descriptor with an explicit range (`vp up vite@^8`) still passes through: that spelling states the intent to leave the catalog. Glob patterns also pass through. Yarn 1 has no catalogs and keeps its behavior.

npm needs no fix. `npm update` never writes package.json; verified on a migrated npm project with plain, named, and toolchain-named updates.

The PTY snapshot `command_update_catalog_protocol_yarn` mirrors the pnpm fixture from #2425: migrate a minimal project, run `vp up vite vite-plus`, and assert both manifests keep their `catalog:` references. `docs/guide/upgrade.md` and `docs/guide/install.md` document the skip behavior. Verified with `cargo test -p vp_pm_cli`, crate clippy and fmt, and the three `command_update_catalog_protocol_*` snapshot cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8595,6 +8595,7 @@ dependencies = [
  "semver 1.0.28",
  "serde",
  "serde_json",
+ "serde_yaml",
  "sha1",
  "sha2 0.10.9",
  "tar",
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_update_catalog_protocol_yarn/.gitignore
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_update_catalog_protocol_yarn/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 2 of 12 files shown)


fengmk2:
@codex review

PR #2440, opened 2026-08-13
feat(install): require npm provenance for platform packages

### LINKED ISSUE #1826: Reject Vite+ install script if the `@voidzero-dev/vite-plus-cli-*` package does not have provenance

kazupon:
## Summary

Closes #1826.

Vite+ previously accepted platform package metadata without requiring npm provenance, allowing an unverified release binary to reach the download path.

Require supported SLSA provenance before downloading standard release binaries through the standalone installers, `vp upgrade`, and `vp-setup.exe`.

## Changes

- Validate `dist.attestations.provenance.predicateType` against SLSA v1 and v0.2 in the shared Rust resolver.
- Fetch and validate exact platform metadata in both standalone installers.
- Fail closed before tarball download for missing, malformed, spoofed, or unsupported provenance.
- Add one mock registry fixture and entrypoint coverage for all four install and upgrade paths.

## Testing

- `cargo test -p vp_setup registry`: provenance parsing, error classification, and download ordering.
- `just check`, `just test`, `just lint`: workspace compilation, tests, and Clippy.
- Mock-registry E2E: rejection before download and acceptance of SLSA v1/v0.2 across the installer and upgrade entrypoints.
- Official npm release smoke test: an existing provenance-backed release still installs.


kazupon:
@codex review

fengmk2:
@kazupon can you update install.ps1 too?

kazupon:
@fengmk2 
Thanks!

If you're referring to the Codex path-spoofing finding, it was specific to install.sh, whose AWK parser previously flattened JSON paths into dot-joined strings. install.ps1 uses nested property access ($Metadata.dist.attestations.provenance.predicateType), so a dotted top-level key cannot satisfy the check.

I also added the dotted-top-level-key regression case to both installer test suites, and the PowerShell case rejects it. 

Please let me know if you had a different PowerShell concern in mind 🙏

fengmk2:
@kazupon Thanks, I will verify it once, and if there are no issues, I will merge it.

fengmk2:
@kazupon I think we should wait for https://github.com/voidzero-dev/vite-plus/pull/2611 finish, the install flow will have a big refactor after that.

cc @liangmiQwQ

kazupon:
@fengmk2 no problem!
I can wait for it!

fengmk2:
@kazupon https://github.com/voidzero-dev/vite-plus/pull/2611 was merged, you can continue now!

kazupon:
@fengmk2 
I've just fixed conflict!
You can check and review again!

fengmk2:
@codex review

packages/cli/install.sh
```diff
@@ -257,6 +261,262 @@ get_version_from_metadata() {
   fi
 }
 
+# Extract the platform tarball URL and provenance predicate from npm version
+# metadata. Bootstrap runs before Node.js is available and cannot require jq,
+# so this parser tracks each JSON path segment and container boundary rather
+# than matching key names or dot-joined paths. Keeping segments separate means
+# a package-defined key containing dots cannot impersonate npm's nested
+# `dist.attestations.provenance` metadata. Invalid JSON fails closed.
+parse_platform_distribution_metadata() {
```
liangmiQwQ, line 270: Parsing JSON in the shell script seems a little bit heavy and unstable for me. But it is not a blocker, we can think about how to improve this part in the future.

liangmiQwQ (approved):
Other LGTM :)

packages/cli/install.sh
kazupon: @liangmiQwQ
Thanks for the comment!
I agree that parsing JSON in the shell is heavy.

At bootstrap we cannot rely on Node or `jq`, and we also cannot execute the downloaded `vp` binary before checking provenance.
With those constraints, I think that `awk` approach is reasonable and the best solution:

- OS-provided parsers (`python3` / `osascript`, etc.) are enough on macOS, but they are not common across minimal Linux environments
- Downloading a JSON parser from a trusted location adds another binary, plus hash pinning and extra failure modes
- Serving something like `latest.txt` from the same origin (viteplus.dev) as `install.sh` would remove parsing, but Vite+ would then have to host and update that metadata on every release
- The same `grep` we use for `version` is not safe here: a spoofed key from `package.json` fields in the npm metadata would pass, so it cannot be the provenance gate

Each of those makes the bootstrap path longer, and none of them are a better fit for keeping bootstrap thin.

So for this path I would rather keep the current parser: no `jq` dependency, and fail-closed. If bootstrap later stops needing to read npm JSON, we can drop it then.

fengmk2 (approved):

packages/cli/install.sh
liangmiQwQ: Yup, a reasonable safety vs maintainability trade-off.

packages/cli/install.ps1
fengmk2, line 27: @kazupon I just thought of a question: if an enterprise has deployed a private npm registry, would provenance verification not be required in that case? It's quite possible that a private registry wouldn't have this information.

kazupon: @fengmk2
Good question.

I would rather not skip provenance for private / mirror registries in this PR.
If we disable the gate whenever `NPM_CONFIG_REGISTRY` is not the public npm registry, anyone can point that env at their own registry and bypass it.

`dist.attestations.provenance` is part of the same version metadata as `dist.tarball` and `dist.integrity`. A mirror that serves the package should also serve that field.
If it does, this check works without a special case.

So I would keep fail-closed for every registry in this PR.
Mirror support can be a follow-up if a real install path needs an explicit opt-out.

fengmk2: LGTM

fengmk2:
The standalone Windows preview installation command fails, because our bridge registry don't includes provenance field. We need to adapt to this issue.

```powershell
.\vp-setup.exe --version "0.0.0-commit.85644fb20281e9210840bf413e7dc377ad65fcd4" --registry "https://registry-bridge.viteplus.dev/"
```

kazupon:
What we were worried about has happened, hasn't it. 😅 

What should we do about this PR?
Should we close it so we can explore other solutions?

fengmk2:
@kazupon Let me quickly make a change to skip the provenance check for requests to registry-bridge.viteplus.dev.
Then I will merge it.

PR #2455, opened 2026-08-14
fix(cli): kill the whole child tree when runCommandSilently times out

fengmk2:
## Bug

`runCommandSilently` resolves on the child's `close` event, but its timeout killed only the direct child. When arbitrary code run by the child spawned a grandchild that inherited the piped stdio, the SIGKILL left the grandchild holding the pipe, `close` never fired, and the caller hung forever.

`vp migrate` hits this through the config-compat worker: resolving a project's `vite.config` executes project code, and any lingering stdio-inheriting descendant (an esbuild-style service, a daemon) wedges the migration permanently under a frozen "Checking config compatibility" spinner, at 0% CPU after the initial burst. This likely explains previously observed `vp migrate` hangs at 0% CPU.

## Fix

- Spawn the child in its own process group on POSIX and kill the group on timeout, so the whole tree dies with it.
- Destroy our pipe ends after the kill, so `close` always fires even when a survivor holds the far end (the Windows path and escaped-group case). The timeout path rejects without reading output, so nothing is lost.

## Test

The new case wedges a child that hands its pipes to a short-lived grandchild. Before the fix the promise never settled and the test died on its own timeout; with the fix the suite passes in under a second. The compat-runner spec passes unchanged.

### DIFF
--- a/packages/cli/src/utils/__tests__/command.spec.ts
+++ b/packages/cli/src/utils/__tests__/command.spec.ts
@@ -27,6 +27,34 @@ describe('command runners', () => {
     ).rejects.toThrow(/timed out after 200ms/);
   });
 
+  it(
+    'times out even when a grandchild inherits the stdio pipes',
+    { timeout: 10_000 },
+    async () => {
+      // Arbitrary project code run by a config worker can spawn its own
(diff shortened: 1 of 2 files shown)


fengmk2:
@codex review

Issue #2356, opened 2026-08-06
Dependency bots update vite-plus and the vite catalog alias out of lockstep


fengmk2:
## What happens

`vp create` / `vp migrate` scaffold two entries that must move in lockstep (`packages/cli/src/utils/constants.ts`):

```yaml
catalog:
  vite: npm:@voidzero-dev/vite-plus-core@0.2.8
  vite-plus: 0.2.8
```

To a dependency bot these are two unrelated packages, `vite-plus` and `@voidzero-dev/vite-plus-core`, so a normal Renovate configuration turns one Vite+ release into two independent PRs: one bumps `vite-plus` and leaves the alias at the old version, the other bumps the alias and leaves `vite-plus` at the old version. Each PR's CI, and the repo state after merging either one, runs vite-plus X against core Y, a pairing that is never released together. Nothing ties the two PRs to each other, and any per-package difference in schedule, `minimumReleaseAge`, or automerge policy pushes them further apart.

Real case: in voidzero-dev/setup.viteplus.dev#36 and voidzero-dev/setup.viteplus.dev#37, one PR reached vite-plus 0.2.8 with the alias still at 0.2.1 while the other reached the alias at 0.2.7 with vite-plus still at 0.2.1, and both sat open for a month. voidzero-dev/setup-vp#119 hit the same split through mismatched age gates. Every repo the migrator produces has this shape, so any Vite+ project with Renovate or Dependabot enabled is exposed. The exact `vitest` override from the same constants is the same class of problem: a bot can bump it away from the version bundled in vite-plus.

## Repo-level workaround

A Renovate group rule puts the bumps in one PR so they move in lockstep:

```json
{
  "groupName": "vite+",
  "matchPackageNames": ["vite-plus", "@voidzero-dev/vite-plus-*", "vitest"]
}
```

Include `vitest` when the project depends on it: vp pins the `vitest` override to the version bundled in vite-plus, so an independent vitest PR desyncs that pin the same way. Installed `@vitest/coverage-*` providers belong in the group too, since vp requires them aligned with the bundled vitest. Note that grouping keeps `vitest` in the same PR only when updates coincide; a vitest release with no simultaneous vite-plus release still produces a lone vitest bump ahead of the bundled version, which only the skew guard below catches.

[...]

PR #2463, opened 2026-08-16
feat: ship a Renovate preset and dependency-bot docs for lockstep updates

fengmk2:
Every `vp create` / `vp migrate` project carries `vite-plus` plus the `vite` alias (`npm:@voidzero-dev/vite-plus-core@<same version>`), and `vp test` projects a pinned `vitest` override. Dependency bots update each package in its own PR, splitting the lockstep set. This is #2356 idea 1, stacked on the fail-fast guard (#2462).

The root `default.json` is a shared Renovate preset, consumed as `"extends": ["github>voidzero-dev/vite-plus"]`. It groups `vite-plus`, `@voidzero-dev/vite-plus-*`, `vitest`, and `@vitest/*` into one PR, and sets `minimumReleaseAge` and `schedule` on the group so no other rule can split member eligibility (the failure mode from voidzero-dev/setup-vp#119; rule shape taken from the merged fixes voidzero-dev/setup-vp#121 and voidzero-dev/setup.viteplus.dev#42). The preset passes `renovate-config-validator`.

Docs: a new "Dependency Update Bots" section in the CI guide documents the preset and an equivalent Dependabot `groups` config, and the migration guide links to it from the overrides section.

Refs #2356

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/default.json
+++ b/default.json
@@ -3,7 +3,7 @@
   "description": "Shared Renovate preset for Vite+ projects: keeps vite-plus, the vite catalog alias (@voidzero-dev/vite-plus-core), and the pinned vitest family in one grouped update. Consume with \"extends\": [\"github>voidzero-dev/vite-plus\"].",
   "packageRules": [
     {
-      "description": "vite-plus and @voidzero-dev/vite-plus-* are published together at one version, and vp pins vitest (plus aligned @vitest/* providers) to the version bundled in vite-plus. One grouped PR keeps the set in lockstep. Explicit minimumReleaseAge and schedule give every member one eligibility policy, so another preset cannot hold one member back and split the group.",
+      "description": "Explicit minimumReleaseAge and schedule give every member one eligibility policy, so another preset cannot hold one member back and split the group.",
       "groupName": "vite+",
       "matchPackageNames": ["vite-plus", "@voidzero-dev/vite-plus-*", "vitest", "@vitest/*"],
       "minimumReleaseAge": "1 day",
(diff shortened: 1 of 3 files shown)


default.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "description": "Shared Renovate preset for Vite+ projects: keeps vite-plus, the vite catalog alias (@voidzero-dev/vite-plus-core), and the pinned vitest family in one grouped update. Consume with \"extends\": [\"github>voidzero-dev/vite-plus\"].",
+  "packageRules": [
+    {
+      "description": "Explicit minimumReleaseAge and schedule give every member one eligibility policy, so another preset cannot hold one member back and split the group.",
+      "groupName": "vite+",
+      "matchPackageNames": ["vite-plus", "@voidzero-dev/vite-plus-*", "vitest", "@vitest/*"],
```
jamietanna, line 8: We'd be happy having a "monorepo preset" for these packages in upstream Renovate, if that'd be helpful

default.json
```diff
@@ -0,0 +1,13 @@
...
+      "minimumReleaseAge": "1 day",
```
jamietanna, line 9: IMO, users of this preset may be a bit surprised by this (especially if it differs to their existing settings)

default.json
setchy: https://github.com/renovatebot/renovate/pull/45631

default.json
fengmk2: Yes, that's why I haven't continued with this PR either, because I do not want to modify the Renovate configuration.

I'm wondering if https://github.com/renovatebot/renovate/pull/45630 this Renovate PR can solve the Vite+ version upgrade issue, then I can remove this preset. 💌

jamietanna: Yes, that's the hope 🤞🏼 

We do also have https://github.com/renovatebot/renovate/pull/45631 for the Renovate config side, so y'all don't need to maintain that yourselves

PR #2476, opened 2026-08-17
RFC: multi-provider vp doc command

fengmk2:
Vite+ had a prototype `vp doc` command that bundled VitePress. We
removed it before the RC release. This RFC brings the command back as a
public command, and this time Vite+ bundles no documentation generator.

`vp doc` serves, builds, and previews documentation sites through a
provider model. The project selects a provider. Vite+ detects the
provider from the declared dependencies, translates
`dev`/`build`/`preview`, and executes the project-installed tool. The
initial built-in providers are VitePress 2, Vocs, Starlight, and
Ox Content, each supported from its first release that runs on Vite 8.

The decisions that most need review:

- Selection order: `doc.provider` in config, then a unique dependency
  marker. A package selects one provider; during a tool migration, when
  both markers coexist, `doc.provider` selects which one runs.
- Workspaces: the `doc` entry of `defaultPackage` works as an implicit
  `-C`. A root without it gets a marker-filtered package picker, or a
  candidate listing in CI.
- `vp doc init` sets up a provider in place and exits nonzero until the
  result is runnable. `vp doc info --json` reports selection, versions,
  and compatibility for humans and coding agents.
- Build caching is opt-in through a configured `run.tasks` entry.
  Integration providers validate their native config before execution.

The usage scenarios show recorded terminal transcripts. A working
implementation backs every behavior in the document and follows as its
own PR after this review lands. `rfcs/cwd-flag.md` gains the `doc` entry
of the `defaultPackage` object form.

### DIFF
--- a/rfcs/doc-command.md
+++ b/rfcs/doc-command.md
@@ -1,0 +1,1524 @@
+# RFC: Multi-provider `vp doc` Command
+
+- Status: Proposed
+- Related: [package-manager-detection.md](./package-manager-detection.md),
+  [merge-global-and-local-cli.md](./merge-global-and-local-cli.md),
+  [cwd-flag.md](./cwd-flag.md),
+  [packages/cli/BUNDLING.md](../packages/cli/BUNDLING.md),
+  [discussion #2445](https://github.com/voidzero-dev/vite-plus/discussions/2445)
+
+## Summary
+
+Add a public `vp doc` command for developing, building, and previewing
+documentation sites:
+
+```bash
+vp doc
+vp doc dev
+vp doc build
+vp doc preview
+```
+
+This RFC calls the selectable documentation integration a **provider** and
+the underlying documentation generator a **tool**. The project selects a
+provider; Vite+ detects it from declared dependencies, translates the common
+actions, `dev`, `build`, and `preview`, and executes the tool. The name
+follows Vitest's `browser.provider` and `coverage.provider`, which select an
+implementation the same way. [discussion #2445] calls the internal mapping
+code an adapter; that word stays internal.
+
+Vite+ bundles no documentation generator and forces no single choice. Most
+providers execute a project-installed CLI. Vite-plugin providers execute
+Vite+'s existing Vite command.
+
+[discussion #2445]: https://github.com/voidzero-dev/vite-plus/discussions/2445
+
+The initial built-in providers are:
+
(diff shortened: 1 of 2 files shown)


ubugeeei:
Yeah!

ubugeeei (approved):

rfcs/doc-command.md
```diff
@@ -0,0 +1,1524 @@
...
+program can thus learn which provider will run, with no tool start.
+`info` never starts the tool, never writes files, and never prompts; it
+follows the `defaultPackage` `doc` redirect like every doc subcommand.
+Resolving a non-static `doc` field evaluates the project's `vite.config`,
+the same file every Vite+ command loads (see Security).
+
+The unresolved states stay machine-readable. With no provider or with
+several markers, `info` exits with status 1, the text form names the
+state and the candidates, and the JSON carries the matching `status` with
+a `candidates` array. A missing installation appears as a null version
```
kazupon, line 713: Thanks for open the RFC!

I’m concerned about the meaning of `"status": "ready"` in `vp doc info --json`.

The RFC states that a missing installation is represented by a null version and an unsupported version by
`"supported": false`, while both still report `"status": "ready"` and exit with code `0`.

I’m concerned that, since `vp doc dev` or `vp doc build` cannot run successfully in either case, users and
 automated consumers may interpret “ready” as a stronger guarantee than it provides.

Could we separate provider resolution from execution readiness?

For example:

```json
{
  "status": "resolved",
  "runnable": false,
  "reason": "unsupported-version"
}
```

Alternatively, status could distinguish states such as ready, not-installed, and unsupported-version.

I think this distinction should be clarified before the JSON schema becomes a public contract.

fengmk2:
We are putting the `vp doc` plan on hold for now.

Working through the RFC made one thing clear: before Vite+ carries a documentation command, we should know which documentation tool our users actually want. Starting with several providers optimizes for a choice nobody has asked us to make yet, and starting with the wrong single tool would be worse. The right first step is to pick one option based on user demand, and we do not have that signal today.

The RFC is complete and stays on its branch, so the design work is not lost. When the demand picture is clear, a single-tool `vp doc` is a small step from where the design stands, and the broader provider model remains available if it ever earns its place.

If you have requests, community threads, or other data points about documentation tooling in Vite+ projects, please share them here. That input is what will bring this back.

PR #2486, opened 2026-08-18
fix(cli): make pack --no-config skip Vite config loading

jong-kyung:
## Summary 

`vp pack --no-config` now skips automatic Vite config resolution instead of only disabling parent-directory traversal.
Local `vite.config.*` files, including `pack` options and hooks, no longer execute when config loading is disabled.

 Explicit `--from-vite` resolution keeps its discovered dependencies so watch mode can still restart when those files change.

### DIFF
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack_no_config.md
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack_no_config.md
@@ -1,0 +1,21 @@
+# command_pack_no_config
+
+## `vp pack --no-config src/index.ts`
+
+should build without loading vite.config.ts
+
+```
+ℹ entry: src/index.ts
+ℹ Build start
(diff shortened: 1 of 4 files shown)


jong-kyung:
@codex review

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack_no_config.md
```diff
@@ -0,0 +1,21 @@
+# command_pack_no_config
+
+## `vp pack --no-config src/index.ts`
```
fengmk2, line 3: Is this command line parameter also supported by tsdown?

jong-kyung: yes. tsdown supports `--no-config`:
https://github.com/rolldown/tsdown/blob/64715feb178a8ffdca684f4bafc2cced4fa4631a/docs/options/config-file.md?plain=1#L54-L62

fengmk2 (approved):

PR #2487, opened 2026-08-18
fix(cli): show config options in pack help

jong-kyung:
## Summary
`vp pack --help` now documents the supported `--no-config` options. This keeps the static help output aligned with the Pack CLI without changing runtime behavior.

Depends on #2486

### DIFF
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack.md
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack.md
@@ -16,7 +16,6 @@ Arguments:
   [...files]  Bundle files
 
 Options:
-  --config-loader <loader>      Config loader to use: auto, native, tsx, unrun (default: auto)
   --no-config                   Disable config file
   -f, --format <format>         Bundle format: esm, cjs, iife, umd (default: esm)
   --clean                       Clean output directory, --no-clean to disable
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack.md
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack.md
@@ -16,6 +16,8 @@ Arguments:
   [...files]  Bundle files
 
 Options:
+  --config-loader <loader>      Config loader to use: auto, native, tsx, unrun (default: auto)
+  --no-config                   Disable config file
   -f, --format <format>         Bundle format: esm, cjs, iife, umd (default: esm)
   --clean                       Clean output directory, --no-clean to disable
   --deps.never-bundle <module>  Mark dependencies as external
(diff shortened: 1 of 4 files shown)


jong-kyung:
@codex review

packages/cli/src/help.ts
```diff
@@ -834,6 +834,11 @@ const commandHelpDocs = {
       {
         title: 'Options',
         rows: [
+          {
+            label: '--config-loader <loader>',
+            description: 'Config loader to use: auto, native, tsx, unrun (default: auto)',
+          },
+          { label: '--no-config', description: 'Disable config file' },
```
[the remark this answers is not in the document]
jong-kyung: Once #2486 is merged, `--no-config` will skip Vite config loading as documented here.

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack.md
```diff
@@ -16,6 +16,8 @@ Arguments:
   [...files]  Bundle files
 
 Options:
+  --config-loader <loader>      Config loader to use: auto, native, tsx, unrun (default: auto)
```
liangmiQwQ: For `--config-loader`, I previously intentionally removed it from all tools' help document, because I think Vite+ has already configured their config loader with rolldown's bundling.

I'm not clear whether `vp pack` behaves differently here, maybe we should considering bring this back in other tools as well if it makes sense.

liangmiQwQ: https://github.com/voidzero-dev/vite-plus/blame/fc855af9204b26b0146a3d67ddfeb5655fc33f2d/.claude/skills/sync-upstream-cli-help/SKILL.md#L41-L42

jong-kyung: Thanks, I saw the upstream help synchronization work in #2330, but I didn’t realize that `--config-loader` was intentionally omitted. I opened this PR after noticing the help mismatch while working on #2486. Would it make sense to remove `--config-loader` from this PR but keep `--no-config`, since #2486 makes it work as documented?

liangmiQwQ: 🙇 Actually when I deleted this before, I also didn't really study the internal workings in detail. Perhaps we should first take a closer look at whether this field should even exist for all these underlying tools, and then prepare accordingly.

jong-kyung: I also don't think we need to expose `--config-loader` for now. Most Vite+ users are expected to use `vite config`, so this does not seem necessary at this point. We can revisit it later if a concrete use case emerges.

fengmk2 (approved):

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_pack/snapshots/command_pack.md
liangmiQwQ: Okay, thanks for handling this. I was dealing with some personal stuff so response late.

PR #2505, opened 2026-08-20
fix(installer): enforce Windows setup requirements

fengmk2:
vp-setup.exe now validates directory variables before it resolves a version or creates an installation root. VP_HOME is the single-root override. Users must set VP_BIN_DIR, VP_DATA_DIR, and VP_CACHE_DIR together. Each value must be an absolute path. Users can set each XDG_* variable independently. The installer no longer provides the redundant `--install-dir` option.

vp-setup.exe accepts Vite+ 0.3.0 or later. CI can also install internal 0.0.0-commit.<sha> builds. The installer rejects all other 0.0.0 versions and earlier releases before it downloads a payload. It no longer probes payloads or uses the monolithic layout.

vp_shared owns directory validation. vp_setup owns version support checks. vp-setup and vp upgrade use these shared checks.

The Windows installer now replaces a dangling current junction. It also removes ANSI output when NO_COLOR is set. The Windows workflow tests invalid directory variables, unsupported versions, junction repair, and NO_COLOR output.

### DIFF
--- a/.github/workflows/test-standalone-install.yml
+++ b/.github/workflows/test-standalone-install.yml
@@ -9,6 +9,7 @@ on:
       - 'packages/cli/install.sh'
       - 'packages/cli/install.ps1'
       - 'packages/tools/src/install-global-cli.ts'
+      - 'packages/tools/src/local-npm-registry.ts'
       - 'crates/vp_installer/**'
       - 'crates/vp_trampoline/**'
       - 'crates/vp_global_cli/**'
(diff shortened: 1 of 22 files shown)


.github/workflows/test-standalone-install.yml
```diff
@@ -1204,6 +1204,94 @@ jobs:
...
+            if (Test-Path $splitRoot) {
+              throw "fresh-install test requires an absent split root: $splitRoot"
+            }
+          }
+          New-Item -ItemType Directory -Force -Path $profile | Out-Null
+
+          Remove-Item Env:VP_HOME, Env:VP_BIN_DIR, Env:VP_DATA_DIR, Env:VP_CACHE_DIR -ErrorAction SilentlyContinue
+          $env:USERPROFILE = $profile
+          $env:HOME = $profile
+          & $installer --yes --version 0.2.9 --no-node-manager --no-modify-path
```
fengmk2: It should throw an error here instead of being compatible with the old version.

wan9chi (approved):

PR #2513, opened 2026-08-20
feat(pm): support bun 1.4 package manager commands

### LINKED ISSUE #2460: Support Bun 1.4 pacakge manager commands
Bun 1.4 adds support for several package manager commands that were previously unavailable:

- bun dedupe
- bun prune
- bun audit fix

We should update the Bun package manager integration to support these commands.

Related Bun issues:

- bun dedupe: https://github.com/oven-sh/bun/issues/1343
- bun prune: https://github.com/oven-sh/bun/issues/3605
- bun audit fix: https://github.com/oven-sh/bun/issues/20238

They were implemented together in:
https://github.com/oven-sh/bun/pull/38333

jong-kyung:
Bun 1.4 added native package manager commands that vp previously worked around: `dedupe`, `prune`, and `audit fix`, plus `--filter` for `add`/`remove`/`update` and `--catalog` for `add`.

Release notes: https://bun.com/blog/bun-v1.4

When the detected bun version is >= 1.4, vp now routes these directly:

- `vp dedupe` runs `bun dedupe`, including `--check`, instead of falling back to `bun install`.
- `vp pm prune` runs `bun prune`, mapping `--prod` to `--production`, instead of a noop. bun prune has no optional flag, so `--no-optional` is declared unsupported for bun.
- `vp pm audit --fix` runs `bun audit fix` instead of a noop.
- `vp add`/`vp remove`/`vp update` forward `--filter`, and `vp add` forwards `--save-catalog` as `--catalog`. `vp update` previously dropped `--filter` silently; bun before 1.4 now warns before dropping it.

Older bun versions keep the previous fallback or noop behavior, with diagnostics that now name the required version. The version gating uses the existing `not_supported(bun < "1.4")` macro rules for flags and a `supports_v1_4_commands` predicate for the hand-written resolvers.

Every command is covered by resolution unit tests and a bun 1.4 PTY snapshot case that runs the real binary: dedupe and check mode, prune removing a dev dependency with `--production`, `audit fix`, and workspace fixtures verifying `--filter` scoping and the `catalog:` entry written by `--catalog`.

Closes #2460


jong-kyung:
@codex

jong-kyung:
@fengmk2 I temporarily fixed the [CI failure](https://github.com/voidzero-dev/vite-plus/actions/runs/32384435197/job/96475848783) caused by the changed **bun build hash**. 
I think we should also mask bun's build hash in [redact.rs](https://github.com/voidzero-dev/vite-plus/blob/main/crates/vp_cli_snapshots/tests/cli_snapshots/redact.rs) so snapshots don't break whenever a new bun version is released. WDYT?

jong-kyung:
I'll handle conflicts

fengmk2 (approved):
@jong-kyung Nice!

PR #2522, opened 2026-08-21
fix(cli): guide local users to global-only commands

### LINKED ISSUE #2515: vp env current does not work if not installed globally
### Describe the bug

I cannot run `vp env current` if vite plus is not installed globally.

Related: https://github.com/voidzero-dev/vite-plus/issues/2503

### Reproduction

N/A

### Steps to reproduce

1. don't install vite plus globally
2. create a project with pnpm and add `vite-plus` as a dependency
3. run `pnpm vp env current`
4. it prints:
   ```shell
   $ pnpm vp env current
   ✓ Lockfile passes supply-chain policies (verified 18m ago)
   Lockfile is up to date, resolution step is skipped
   Already up to date
   Done in 931ms using pnpm v11.22.0
   ```

### System Info

`vp env current`: N/A (see above)

```shell
$ pnpm vp --version

vp v0.2.9

Local vite-plus:
  vite-plus  v0.2.9

Tools:
  vite             v8.2.1
  rolldown         v1.2.3
  vitest           v4.1.10
  oxfmt            v0.62.0
  oxlint           v1.77.0
  oxlint-tsgolint  v7.0.2001
  tsdown           v0.22.14
```

### Used Package Manager

pnpm

### Logs

```shell

```

### Validations

jong-kyung:
The local CLI now shows installation guidance when users run global-only commands such as `env`, `upgrade`, or `implode`. Other unknown commands continue to use the standard “command not found” error.

Closes #2515


jong-kyung:
@codex review

packages/cli/binding/src/cli/help.rs
```diff
@@ -313,7 +321,7 @@ mod tests {
     fn global_subcommands_produce_invalid_subcommand_error() {
         use clap::error::ErrorKind;
 
-        for subcommand in ["config", "create", "env", "hooks", "migrate"] {
+        for subcommand in ["config", "create", "env", "hooks", "implode", "migrate", "upgrade"] {
```
fengmk2, line 331: Could we avoid maintaining the global-only command list independently here, or at least define it as a named constant/helper with a comment pointing to the global CLI command surface?

`env`, `upgrade`, and `implode` are correct today, but this list can silently drift when we add another global-only command.

jong-kyung: Agreed. How about extracting the global-only command enum into a shared crate and consuming it from both `crates/vp_global_cli/src/cli.rs` and the `local CLI binding`? If that direction sounds good, I can open a separate refactoring PR first.

fengmk2: Take a try.

jong-kyung:
If #2525 was merged, I'll handle it again

jong-kyung:
@codex review

fengmk2 (approved):

PR #2523, opened 2026-08-21
refactor(cli): parse JS command arguments with clap

fengmk2:
Move argument parsing for the staged, config, hooks, migrate, and create commands from JavaScript to Rust. Each NAPI parser uses clap for strict validation and returns a typed result to JavaScript.

## Call graph

```text
process.argv
    |
    v
local Node.js CLI
packages/cli/src/bin.ts
    |
    | raw arguments for one JavaScript command
    v
NAPI parser function
packages/cli/binding/src/js_command_args/
    |
    v
clap command grammar
    |
    +-- checks aliases and option boundaries
    +-- converts values
    +-- checks explicit negation
    +-- rejects unknown options
    +-- rejects invalid positional arguments
    |
    v
validated Rust arguments
    |
    v
typed NAPI parse outcome
    |
    v
JavaScript command operations
```

JavaScript sends raw arguments to one NAPI parser. JavaScript does not parse the returned values again.

The Rust parsers print command help through the shared vp_cli_help formatter. The global and local CLI paths use the same help format. The create parser keeps all template arguments after the separator in their original order.

This change removes duplicate JavaScript option data and the direct mri dependency. The RFC defines the parser rules, help synchronization, NAPI result types, and command ownership.

## Compatibility

Argument parsing is now strict. The CLI rejects unknown options and extra positional arguments. It rejects unsupported negative string options and repeated scalar options.

The staged command rejects invalid concurrency values. It also rejects empty `--cwd`, `--diff`, and `--diff-filter` values before JavaScript runs.

The create command rejects `--all` and invalid package-manager values. These inputs could pass through `mri` or fail later in JavaScript.

## Performance

The benchmark compares the base commit `45acff9b4` with this branch. It ran on macOS ARM64 with Node.js 22.22.0.

Each CLI result used 25 to 30 alternating paired runs after four warm-up pairs. A negative change is faster.

The parser-only test used `vp staged --allow-empty --concurrent=2 --diff-filter ACMR --no-stash`.

| Case | Base mean | PR mean | Change |
| --- | ---: | ---: | ---: |
| Staged parser only | 1.20 µs | 6.33 µs | +5.13 µs |
| `vp --version` control | 133.2 ms | 133.4 ms | +0.2 ms |
| `vp staged --help` | 136.1 ms | 133.9 ms | -2.2 ms |
| `vp config --help` | 149.4 ms | 144.7 ms | -4.7 ms |
| `vp hooks --help` | 134.4 ms | 132.9 ms | -1.5 ms |
| `vp migrate --help` | 155.8 ms | 150.0 ms | -5.8 ms |
| `vp create --help` | 154.5 ms | 150.1 ms | -4.5 ms |
| `vp staged --cwd` | 135.2 ms | 134.5 ms | -0.7 ms |
| `vp hooks unknown` | 131.7 ms | 132.0 ms | +0.3 ms |
| `vp config --hooks-dir` | 165.3 ms | 144.6 ms | -20.7 ms |

The clap/NAPI parser is 5.3 times slower in isolation. This adds about 5 µs to a CLI process that takes 130 to 155 ms.

The CLI calls the parser one time. The complete CLI path has no measurable regression. Help is 1% to 4% faster.

`vp config --hooks-dir` is not a parser-only comparison. The base command starts hooks validation. The PR rejects the missing value first.

Unknown-option timings are not comparable. The base parser can accept an unknown option and start command work. The PR rejects it during parsing.


rfcs/napi-clap-cli-args.md
```diff
@@ -0,0 +1,857 @@
...
+
+```text
+crates/vp_cli_help/
+  Cargo.toml
+  src/lib.rs
+
+packages/cli/binding/src/js_command_args/
+  mod.rs
+  parse.rs
+  create.rs
```
fengmk2: Should command implementation code be placed in a separate directory for better management?

fengmk2:
@codex review

wan9chi (approved):

PR #2538, opened 2026-08-23
refactor(js-runtime): drop unused async-trait from JsRuntimeProvider

jong-kyung:
`JsRuntimeProvider` was annotated with `#[async_trait]`, but none of its methods are async. `name`, `platform_string`, `get_download_info`, `binary_relative_path`, `bin_dir_relative_path`, and `parse_shasums` are all synchronous, so the macro had nothing to transform and generated the same code as the plain trait.

If a future runtime provider (Bun, Deno) needs async trait methods, the attribute can be added back at that point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8576,7 +8576,6 @@ dependencies = [
 name = "vp_js_runtime"
 version = "0.0.0"
 dependencies = [
- "async-trait",
  "backon",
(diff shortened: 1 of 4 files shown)


jong-kyung:
@codex review

fengmk2 (approved):

PR #2540, opened 2026-08-23
feat(migrate): pin setup-vp workflow versions

fengmk2:
`vp migrate` now replaces exact `voidzero-dev/setup-vp@v1` references in GitHub Actions workflows and composite actions under `.github`. It replaces the frozen tag with `v1.18.0`. It keeps exact versions, commit SHAs, comments, quotes, and line endings.

The same version pin applies to new Copilot setup workflows. Renovate keeps the pin current.

This change adds unit tests and a PTY snapshot for existing Vite+ projects. The snapshot also checks repeated migration runs. The documentation now explains this behavior.


cpojer (approved):

stybo:
image

After running vp upgrade and then vp migrate the version is still v1. Is doesn't seem right to me?

PR #2541, opened 2026-08-23
perf(migrate): collect Vitest source signals in one traversal

### LINKED ISSUE #2420: perf(migrate): Vitest signal detection repeatedly traverses package source trees
### Summary

During the full project rewrite path of `vp migrate`, the same package source tree is traversed synchronously multiple times to collect Vitest browser/provider signals.

Under the following full-miss conditions:

- the source contains no relevant Vitest browser/provider references;
- no tsconfig contains Vitest types that must be retained;
- package.json contains no webdriverio/provider dependency that causes an early return;

the Vitest signal-detection phase reads every eligible TS/JS file through Node.js `fs.readFileSync`:

- 5 times for a standalone project;
- 9 times for each workspace package in a monorepo.

These counts cover only the JavaScript-side Vitest signal scans. They do not represent all file I/O performed during migration. For example, the later Rust/NAPI import rewriter also traverses and reads source files, but those reads are not visible to an `fs.readFileSync` probe.

### Local reproduction

I created temporary test projects containing many small `.ts` files with no Vitest browser/provider references. Their package.json files also contained no dependencies that would cause the checks to return early.

I then monitored `fs.readFileSync` calls for those fixture source files and invoked the production migration entry points directly.

| Call path | Source files | Source reads through `fs.readFileSync` | Duration |
| --- | ---: | ---: | ---: |
| `rewriteStandaloneProject` | 5,000 | 25,000 | 486 ms |
| `rewriteMonorepo` (2 × 2,500 files) | 5,000 | 45,000 | 734 ms |

Environment: macOS, Node.js 25.9.0, with the filesystem cache pre-warmed.

The per-file read counts are deterministic: exactly 5 reads per file for the standalone project and exactly 9 reads per file for each monorepo workspace package. Every detection result was `false`, so each check took its complete miss path.

The timings are included only to illustrate the approximate impact and will vary between machines.

### Observed call paths

`sourceTreeMatches` recursively traverses directories with `readdirSync` and reads every TS/JS file with an eligible extension using `readFileSync`. It returns early only after finding a match, so a complete miss is the worst-case path.

#### Standalone

Within a single `rewriteStandaloneProject` call:

1. `collectProviderSourceModes` scans once for each of the two current opt-in providers: 2 passes;
2. `usesVitestBrowserMode` performs another scan: 1 pass;
3. `sourceTreeReferencesRetainedVitestModule` scans the source again: 1 pass;
4. when no webdriverio dependency causes a short circuit, `usesWebdriverioProvider` performs another scan: 1 pass.

This results in 5 reads per source file.

`providerSourceModes` already contains the webdriverio source-detection result, but the standalone path subsequently calls `usesWebdriverioProvider` anyway.

In addition, `sourceTreeReferencesRetainedVitestModule` first performs a separate recursive directory traversal to discover tsconfig files and reads any matching tsconfig files it finds. As a result, the package directory tree is traversed 6 times on a complete standalone miss, with 5 of those passes reading TS/JS source files.

#### Monorepo workspace package

`rewriteMonorepo` first computes workspace-level signals. For every workspace package on the complete miss path:

- `workspaceUsesWebdriverio`: 1 source scan;
- `workspaceUsesVitestDirectly`:
  - retained Vitest module: 1 scan;
  - browser mode: 1 scan;
- `collectInjectedProviderNames`:
  - two provider scans: 2 scans.

The workspace phase therefore reads each source file 5 times.

The same package is then processed by `rewriteMonorepoProject`, which recomputes:

- browser mode: 1 scan;
- retained Vitest module: 1 scan;
- source modes for two providers: 2 scans.

The package phase adds another 4 reads, for a total of 9 reads per source file in each workspace package.

The monorepo root package is not subsequently passed through `rewriteMonorepoProject`, so source files belonging directly to the root package are generally read 5 times rather than 9.

The two retained-module checks also each add a separate directory-tree traversal for locating tsconfig files. A workspace package therefore undergoes 11 related full directory-tree traversals on the complete miss path, 9 of which read TS/JS source files.

### Reusing scan results

In the reproduction projects above, no source files were modified between these scans in a way that could affect the results.

In the general monorepo path, however, `rewriteMonorepoProject` performs configuration merging before the package-level scans. That step may create or modify `vite.config.ts`. Reusing results between the workspace and package phases would therefore require confirming that these configuration changes cannot affect the relevant signals, or retaining the two phases while consolidating the scans performed within each phase.

Regardless of whether results can be cached across phases, each individual phase currently performs separate full traversals for several closely related boolean signals.

### Impact

The complete miss path remains linear in the number of files, but it has a fixed synchronous I/O amplification:

- standalone package: 5 reads per eligible source file;
- monorepo workspace package: 9 reads per eligible source file.

These directory traversals and file reads are synchronous and directly block `vp migrate`. The cost becomes noticeable in projects containing thousands or tens of thousands of source files.

Ideally, browser mode, retained Vitest module references, and provider signals could be collected during a single source-tree traversal, with the resulting scan data shared by callers within the same migration phase.

m0g3r:
Closes #2420

## Problem

`vp migrate` derives several independent boolean signals from a package's source tree — browser mode, retained upstream `vitest` module references, and one source-scan per opt-in browser provider. Each ran its own full recursive traversal via `sourceTreeMatches`, reading every eligible TS/JS file once per signal.

On the complete-miss path (no browser/provider references, no tsconfig retaining Vitest types, no webdriverio dependency to short-circuit) that is **5 reads per source file** for a standalone project, exactly as the issue reports.

## Changes

- `sourceTreeMatches` is generalized to `sourceTreeMatchesEach`, which evaluates several content predicates in a **single pass**: each file is read once and offered to every predicate that has not yet matched, unwinding as soon as all are decided. `sourceTreeMatches` becomes a one-predicate wrapper over it, so there is still a single traversal implementation.
- New `collectPackageSourceScanSignals` enrolls all the per-package signals — browser mode, both provider scans, and the retained-vitest-module scan — in that one pass.
- Both per-package call sites use it: `rewriteStandaloneProject` and `rewriteMonorepoProject`.
- The standalone path stops rescanning the tree for the webdriverio provider. As the issue notes, `providerSourceModes` already holds that result.

Two deliberate limits:

- **The tsconfig short-circuit is preserved.** `sourceTreeReferencesRetainedVitestModule` returns early when a tsconfig already settles the signal; in the combined scan that predicate is simply not enrolled in that case, so the tree is not walked for it.
- **Results are not shared between the workspace and package phases** of a monorepo migration. The issue points out that `rewriteMonorepoProject` performs config merging between them, which may create or modify `vite.config.ts`, so cross-phase reuse would need a separate correctness argument. This PR consolidates *within* each phase only — the fallback the issue describes as safe. The workspace phase's own per-package scans are left for a follow-up.

Per-predicate semantics are unchanged: same traversal order, same skip directories, same nested-package boundary, same "unreadable file is ignored and scanning continues" behaviour.

## Performance

Measured on the full-miss path — the worst case, and the one #2420 reports: a standalone project with no browser/provider references anywhere, no tsconfig retaining Vitest types, and no webdriverio dependency to short-circuit a scan.

| source files | before (`a583efa`) | after (`6e88cbb`) | speedup | reads per source file |
| --- | --- | --- | --- | --- |
| 250 | 6.58 ms | 2.28 ms | **2.9×** | 5 → 1 |
| 1,000 | 24.04 ms | 7.21 ms | **3.3×** | 5 → 1 |
| 4,000 | 97.07 ms | 29.08 ms | **3.3×** | 5 → 1 |

Each cell is the median of 15 timed iterations after 3 warmups, and each row is the median of three such rounds; run-to-run spread stayed under 8%. Node v22.22.2, Linux x64, 4-core Xeon @ 2.10 GHz, warm page cache.

What is timed is the per-package scan phase exactly as each revision's `rewriteStandaloneProject` performs it:

- **before** — `collectProviderSourceModes` (one traversal per opt-in provider = 2) + `usesVitestBrowserMode` (1) + `sourceTreeReferencesRetainedVitestModule` (1) + the redundant `usesWebdriverioProvider` rescan in the `package.json` callback (1) = **5 traversals**;
- **after** — `collectPackageSourceScanSignals` = **1 traversal**.

The measured read counts came out at exactly 5 per source file before and 1 after, independently reproducing the amplification the issue reports and matching the regression test's `expected 5 to be 1`.

Wall time falls ~3.3× rather than the full 5× because only the read and UTF-8 decode per file are removed — each file's content is still matched against the same predicates either way. These are warm-page-cache numbers, so they understate the gain on a cold cache.

<details>
<summary>How this was measured</summary>

`source-scan.ts` cannot be imported on its own in a bare checkout: it imports `projectUsesVitestDirectly` from the `../migrator.ts` barrel, which transitively reaches the native NAPI binding. That import — along with `readPackageJsonIfExists` and `WorkspacePackage` — is used only by `workspaceUsesVitestDirectly`, which is not on the per-package scan path.

So the benchmark loads each revision's `source-scan.ts` with its first 40 lines (the import block plus that one function) replaced by a prelude that inlines the three constants it needs from `shared.ts` and a JSON-parsing stand-in for `hasVitestTypesInTsconfig`. Everything from line 41 onward is byte-identical to the shipped module at that revision (verified with `diff`), so every measured function is the real one. The `hasVitestTypesInTsconfig` stand-in is never actually invoked, because the fixture contains no tsconfig.

The fixture is generated: N plausible ~1.4 KB TS modules, 20 per directory under `src/`, none containing any scan hint string. Before timing, each run asserts that every signal came back `false`, so a fixture that accidentally short-circuited a traversal would fail rather than silently produce a flattering number.

</details>

## Testing

New regression test `reads each source file once during a standalone migration` counts `fs.readFileSync` calls against a full-miss fixture and asserts each source file is read once. It also asserts every file is read *at least* once, so a fixture that stopped exercising the scan would fail rather than silently pass.

Verified in both directions:

- with `source-scan.ts` and `orchestrators.ts` reverted to `main` and the test kept, it **fails**: `AssertionError: expected 5 to be 1` — independently reproducing the issue's measured standalone count;
- with the change applied, it **passes**.

Checks actually run on this branch:

- `vitest run packages/cli/src/migration` — 409 passed across 10 files
- `vitest run` (full unit suite) — 1023 passed, 1 skipped; 7 pre-existing snapshot failures in `packages/prompts/src/__tests__/render.spec.ts`, identical on unmodified `main` in this environment (verified by re-running on a clean tree: 1022 passed with the same 7 failures)
- `tsgo -b tsconfig.json` — error set byte-identical to unmodified `main` here (87 pre-existing errors from unbuilt docs/workspace packages, none in the touched files)
- `vp lint --type-aware --type-check` — clean for the touched files
- `vp fmt --check` — clean

Not run: the PTY snapshot suite and ecosystem e2e. This change alters no CLI output, only how many times the same files are read.

## AI assistance

Claude Opus 5 wrote the implementation, the test, the benchmark, and this description. The change is agent-authored and has not had a separate human review. The results quoted above are from actual runs, not estimates.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W7Bh2oBZWrV86xxpQ3SuEV)_

### DIFF
--- a/packages/cli/src/migration/__tests__/migrator.spec.ts
+++ b/packages/cli/src/migration/__tests__/migrator.spec.ts
@@ -9237,3 +9237,74 @@ describe('collectMigrationSetupPlan non-interactive editor conflicts', () => {
     expect(plan.editorConflictDecisions.get('workspace.xml')).toBe('skip');
   });
 });
+
+// Counts how many times each of `sourceFiles` is read through `fs.readFileSync`
+// while `run()` executes. Used to assert the migration's source-tree scans do
+// not re-read the same file once per signal (#2420).
+function countReadsPerSourceFile(sourceFiles: string[], run: () => void): Map<string, number> {
+  const tracked = new Set(sourceFiles);
+  const counts = new Map<string, number>(sourceFiles.map((file) => [file, 0]));
+  const readFileSync = fs.readFileSync.bind(fs);
+  vi.spyOn(fs, 'readFileSync').mockImplementation(((file: unknown, ...rest: unknown[]) => {
+    if (typeof file === 'string' && tracked.has(file)) {
+      counts.set(file, (counts.get(file) ?? 0) + 1);
+    }
+    return (readFileSync as (...args: unknown[]) => unknown)(file, ...rest);
+  }) as typeof fs.readFileSync);
+  run();
+  return counts;
+}
+
+describe('source-tree scan read amplification (#2420)', () => {
+  let tmpDir: string;
+
+  beforeEach(() => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'vp-test-scan-reads-'));
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  });
+
+  // Full-miss fixture: no vitest browser/provider references anywhere, no
+  // tsconfig retaining vitest types, and no webdriverio/provider dependency to
+  // short-circuit a scan. Every signal therefore takes its complete miss path,
+  // which is the worst case for traversal count.
+  function writeFullMissProject(fileCount: number): string[] {
+    fs.writeFileSync(
+      path.join(tmpDir, 'package.json'),
+      JSON.stringify({
+        name: 'scan-reads',
+        devDependencies: { vite: '^7.0.0' },
+      }),
+    );
+    const srcDir = path.join(tmpDir, 'src');
+    fs.mkdirSync(srcDir);
+    const sourceFiles: string[] = [];
+    for (let index = 0; index < fileCount; index++) {
+      const filePath = path.join(srcDir, `mod-${index}.ts`);
+      fs.writeFileSync(filePath, `export const value${index} = ${index};\n`);
+      sourceFiles.push(filePath);
+    }
+    return sourceFiles;
+  }
+
+  it('reads each source file once during a standalone migration', () => {
+    const sourceFiles = writeFullMissProject(6);
+    const workspaceInfo = makeWorkspaceInfo(tmpDir, PackageManager.pnpm);
+
+    const counts = countReadsPerSourceFile(sourceFiles, () => {
+      rewriteStandaloneProject(tmpDir, workspaceInfo, true, true);
+    });
+
+    // Every eligible source file must be visited, so a zero count would mean the
+    // fixture stopped exercising the scan rather than that the scan got faster.
+    for (const file of sourceFiles) {
+      expect(counts.get(file)).toBeGreaterThan(0);
+    }
+    // Before the single-traversal change this was 5: two provider scans, browser
+    // mode, the retained-vitest-module scan, and a redundant webdriverio rescan.
+    expect(Math.max(...counts.values())).toBe(1);
+  });
+});
--- a/packages/cli/src/migration/migrator/orchestrators.ts
+++ b/packages/cli/src/migration/migrator/orchestrators.ts
@@ -9,7 +9,7 @@ import {
   applyYarnWorkspaceHoistingFix,
   cleanupDeprecatedTsconfigOptions,
   collectInjectedProviderNames,
-  collectProviderSourceModes,
+  collectPackageSourceScanSignals,
   collectVitestEcosystemInstallDependencyNames,
   createCatalogDependencyResolver,
   dropRemovePackageOverrideKeys,
@@ -45,10 +45,7 @@ import {
   rewriteYarnrcYml,
   setDirectViteEdge,
   setPackageManager,
-  sourceTreeReferencesRetainedVitestModule,
   takePnpmWorkspaceSettings,
-  usesVitestBrowserMode,
-  usesWebdriverioProvider,
   workspaceUsesVitestDirectly,
   workspaceUsesWebdriverio,
   wrapLazyPluginsInViteConfig,
@@ -56,6 +53,7 @@ import {
 import { type MigrationReport } from '../report.ts';
 import {
   PROVIDER_OVERRIDE_DROP_NAMES,
+  WEBDRIVERIO_PROVIDER,
   pnpmMajor,
   type CatalogDependencyResolver,
   type PnpmPackageJsonSettings,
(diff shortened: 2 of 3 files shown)


fengmk2:
Could you show a before and after speed comparison in the PR description?

m0g3r:
Added a **Performance** section to the description with before/after numbers.

Summary, on the full-miss path #2420 describes (median of 15 iterations × 3 rounds, Node 22.22.2, warm cache):

| source files | before | after | speedup |
| --- | --- | --- | --- |
| 250 | 6.58 ms | 2.28 ms | 2.9× |
| 1,000 | 24.04 ms | 7.21 ms | 3.3× |
| 4,000 | 97.07 ms | 29.08 ms | 3.3× |

Reads per source file go from exactly 5 to exactly 1, which independently reproduces the amplification the issue reports. Wall time improves ~3.3× rather than 5× because only the read and decode per file are eliminated — the predicates still run over each file's content either way.

The description also has a collapsed note on exactly what was timed and how, since the scan module needs its barrel import removed to load without the native binding.

---

PR #2543, opened 2026-08-23
fix(create): report install and format failures in the create summary

### LINKED ISSUE #2453: Improve vp create completion reporting when installation or formatting fails
## Description

When dependency installation and formatting fail during `vp create`, the command can still finish by:

- displaying the project as `Scaffolded`;
- suggesting `vp run` as the next step;
- returning exit code 0.

In the same command, the user is first shown that manual installation and formatting may still be required, but the final output and process status indicate that project creation completed successfully.

## Observed flow

```mermaid
flowchart LR
    subgraph actual["Actual execution"]
        A["Project files are generated"] --> B["Dependency installation fails"]
        B --> C["Formatting fails"]
    end

    subgraph result["Final signals from the CLI"]
        D["Displays: Scaffolded"]
        D --> E["Suggests: Next: vp run"]
        E --> F["Returns exit code 0"]
    end

    C -->|"But the command still continues to"| D
```

The project directory and `package.json` are generated, but dependency installation and formatting do not complete. Despite that, the final message, suggested next step, and exit code all communicate that the complete operation succeeded.

## Reproduction

On macOS or Linux, run the following commands in a temporary directory:

```bash
workdir=$(mktemp -d)
cd "$workdir"

env -u VP_SKIP_INSTALL \
  VP_CLI_BIN=/usr/bin/false \
  vp create vite:application \
  --no-interactive \
  --no-hooks \
  --no-git \
  --directory create-failure-demo

exit_code=$?

echo "exit code: $exit_code"
test -d create-failure-demo && echo "project directory exists"
test -f create-failure-demo/package.json && echo "package.json exists"
test -d create-failure-demo/node_modules || echo "node_modules missing"
```

`VP_CLI_BIN=/usr/bin/false` makes the installation and formatting commands started internally by `vp create` return a failure consistently.

## Actual result

The stable terminal output contains the following lines:

```text
You may need to run "vp install" manually in <temporary-directory>/create-failure-demo
You may need to run "vp fmt" manually in <temporary-directory>/create-failure-demo
◇ Scaffolded create-failure-demo with Vite application
• Node 25.9.0  pnpm 11.21.0
→ Next: cd create-failure-demo && vp run
```

The recorded filesystem and process state is:

```text
exit code: 0
project directory exists
package.json exists
node_modules missing
```

## My understanding of the behavior

This appears to combine two different states:

1. The project directory and template files were generated.
2. The complete project creation process finished successfully.

In this reproduction, the first state is true: the `create-failure-demo` directory and its `package.json` exist.

The second state is not true: dependency installation and formatting both returned failures, and `node_modules` was not created.

The final output does not distinguish between these states. Instead, it presents three strong success signals:

- `Scaffolded` indicates that project creation completed.
- `Next: vp run` indicates that the project is ready to run.
- Exit code 0 indicates that the entire command completed successfully.

Even though the manual recovery messages appear earlier, users will naturally treat the final lines as the authoritative result of the command. This is especially noticeable after a long installation log, where the final summary is usually the first place users look to understand what happened.

For scripts and CI systems, the result is even less ambiguous: they generally use the process exit code rather than interpret terminal text. An exit code of 0 records this operation as successful.

The command therefore exposes several conflicting states at the same time:

```text
Dependency installation: failed
Formatting: failed
Project directory: created
package.json: created
node_modules: missing
Final summary: successful
Suggested next step: run the project
Process exit status: successful
```

The continued existence of the generated directory is not necessarily the confusing part. The confusion comes from presenting “project files were generated” and “the complete creation process succeeded” as the same state.

## Verification environment

- macOS 27.0 on arm64
- Node.js v25.9.0
- Package manager selected by `vp create`: pnpm 11.21.0
- Confirmed with the bundled local CLI and directly against `packages/cli/src/create/bin.ts` at commit `29036b2ff12ea30f80088de4a593842df2846c3b`

m0g3r:
Closes #2453

## Problem

`vp create` writes the template files first, then installs dependencies and formats the result. Those last two steps can fail after the project directory already exists — and when they did, the completion summary reported the run as a success anyway.

Reproduced with the issue's recipe against this branch's base (`c4dcb15`), using the built local CLI:

```
$ env -u VP_SKIP_INSTALL VP_CLI_BIN=/usr/bin/false \
    vp create vite:application --no-interactive --no-hooks --no-git \
    --directory create-failure-demo
You may need to run "vp install" manually in .../create-failure-demo
You may need to run "vp fmt" manually in .../create-failure-demo
◇ Scaffolded create-failure-demo with Vite application
• Node 22.22.2  pnpm 11.23.0
→ Next: cd create-failure-demo && vp run
=== EXIT CODE: 0 ===
node_modules missing
```

Three success signals — `Scaffolded`, `Next: vp run`, exit `0` — for a project that cannot run. The earlier "you may need to…" lines are easy to lose above a long install log, and a CI job reading only the exit code records the operation as successful.

## Changes

- `resolveCreateCompletion` (in `create/utils.ts`) turns the install and format results into: which steps failed, which command is worth suggesting next, and whether the exit code should be non-zero.
- `showCreateSummary` names each failed step, points `Next:` at `vp install` when dependencies are missing (the suggested `vp run` cannot work without them), and raises `process.exitCode`.
- The format result was previously discarded at all three `runViteFmt` call sites; it is now threaded through alongside the install result.

Same input on this branch:

```
◇ Scaffolded create-failure-demo with Vite application
• Node 22.22.2  pnpm 11.23.0
✗ Dependencies were not installed
✗ Code was not formatted
→ Next: cd create-failure-demo && vp install
=== EXIT CODE: 1 ===
```

The directory is still reported as scaffolded — as the issue notes, that part is true and is not the confusing bit.

Three behaviours deliberately preserved:

- **A skipped install is not a failure.** `VP_SKIP_INSTALL` returns `status: 'skipped'`; only `'failed'` counts.
- **The exit code is only ever raised, never lowered.** `handleIgnoredBuilds` already sets `process.exitCode = 1` for a gated build that failed under `--approve-builds`, and this must not clear it.
- **Only a failed install fails the command.** See below.

The success path is unchanged, byte for byte — no new line is printed and `Next:` still reads `vp run`.

## A format failure is reported but does not fail the command

The first revision of this PR exited non-zero for a format-only failure too, and flagged that as a call for you to make. **CI answered it**, so I have taken the exit-`0` variant rather than leave it open.

The PTY snapshot suite failed on `create_framework_shim_vue` and `new_create_vite`, on all three platforms. Both scaffold a template whose generated `vite.config.ts` imports a plugin that is not installed at the moment `vp fmt` runs:

```
vite.config.ts (2:16) [UNRESOLVED_IMPORT] Could not resolve '@vitejs/plugin-vue' in vite.config.ts
error: Failed to resolve vite config: ... Cannot find package '@vitejs/plugin-vue'
You may need to run "vp fmt" manually in <workspace>/vite-plus-application
◇ Scaffolded vite-plus-application with Vue + TypeScript
✗ Code was not formatted
→ Next: cd vite-plus-application && vp run
```

That project is complete and runnable; only formatting was skipped. Those two cases declare the `vp create` step as `snapshot = false`, so they recorded nothing until the step started exiting `1` — which is exactly the point: **a routine `vp create` of a `react-ts` or `vue-ts` template would have started failing every CI job that runs it.**

So `resolveCreateCompletion` now returns `exitCode: installFailed ? 1 : 0`. A format failure is still named in the summary (`✗ Code was not formatted`) — silently dropping it is what #2453 is about — and `Next:` still points at `vp run`, because the project runs. Only a missing `node_modules` makes the command itself fail.

Happy to scope the non-zero exit to `--no-interactive` runs as well, mirroring the existing `handleIgnoredBuilds` condition, if you would prefer that.

## A note on the small refactor

`showCreateSummary` and `getNextCommand` moved from `create/bin.ts` to a new `create/summary.ts`, unchanged apart from the fix. `bin.ts` runs `main()` on import, so nothing in it can be unit-tested; moving these two functions is what let the regression test assert on the real summary output and exit code rather than only on the decision helper. Happy to drop the move if you would rather keep them in `bin.ts`.

## Testing

New `create/__tests__/summary.spec.ts` covers the summary output and exit code; six cases added to `create/__tests__/utils.spec.ts` cover the decision helper. Two of them now assert that a format-only failure is reported *and* leaves the exit code alone, so the snapshot regression above cannot come back silently.

Verified in both directions. With `exitCode` restored to the previous `failures.length > 0 ? 1 : 0` (and the tests kept), exactly the two guarding tests fail — `2 failed | 53 passed`:

```
× reports a failed format without failing the command
  AssertionError: expected { …(3) } to deeply equal { …(3) }
× still suggests running the project and succeeds when only formatting failed
  AssertionError: expected 1 to be undefined
```

With the change applied, `55 passed`.

Checks actually run on the latest commit:

- `vitest run packages/cli/src/create/__tests__/{summary,utils}.spec.ts` — 55 passed, and the both-directions result above.
- `oxlint@1.79.0 -D correctness -D perf -D suspicious` on all four files — exit 0.
- `oxfmt@0.64.0 --check` with this repo's `fmt` block from `vite.config.ts` — all four files correctly formatted. (Run without `-c` it reports issues on untouched files too, so the repo config is the meaningful check.)
- `git diff --check` — clean.

Earlier revision, still valid for the parts it covers: `vitest run` (full unit suite) — 1034 passed, 1 skipped, with 7 pre-existing `packages/prompts` snapshot failures confirmed identical on a clean tree; `tsgo -b tsconfig.json` — error set byte-identical to a clean tree.

**Correcting the previous description:** it said no `create` snapshot exercises a failing install or format, so I would not expect snapshot churn, and that I could not confirm it by running the suite. The first half was wrong — no fixture *records* such a failure, but two fixtures *produce* one, and they surface it through the exit code rather than through recorded output. CI caught what I could not run locally.

The PTY snapshot suite still cannot run in my environment, so the fix above is verified by unit tests and by root-causing both recorded diffs, not by a local suite run.

## AI assistance

Claude Opus 5 wrote the implementation, the tests, the reproductions, and this description. The change is agent-authored and has not had a separate human review. Every result quoted above is from an actual run, not an estimate.

### DIFF
--- a/packages/cli/src/create/__tests__/summary.spec.ts
+++ b/packages/cli/src/create/__tests__/summary.spec.ts
@@ -1,0 +1,98 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { getNextCommand, showCreateSummary } from '../summary.js';
+
+// `terminal.ts` reaches the NAPI binding for the Vite+ header, so stub it
+// outright rather than importing the real module. Styling is not asserted.
+const terminal = vi.hoisted(() => ({
+  log: vi.fn(),
+  accent: (text: string) => text,
+  formatDuration: (durationMs: number) => `${durationMs}ms`,
+}));
+
+vi.mock('../../utils/terminal.ts', () => terminal);
+
+const installed = { durationMs: 1200, exitCode: 0, status: 'installed' } as const;
+const formatted = { durationMs: 30, exitCode: 0, status: 'formatted' } as const;
+const failed = { durationMs: 30, exitCode: 1, status: 'failed' } as const;
+
+const baseOptions = {
+  description: 'Vite application',
+  packageManager: 'pnpm',
+  packageManagerVersion: '11.21.0',
+  projectDir: 'create-failure-demo',
+};
+
+// `showCreateSummary` writes through `log`, so assert on the joined output
+// rather than the exact styling of any one line.
+function output() {
+  return terminal.log.mock.calls.map((call) => String(call[0])).join('\n');
+}
+
+describe('getNextCommand', () => {
+  it('prefixes a cd when the project is in a subdirectory', () => {
+    expect(getNextCommand('demo', 'vp run')).toBe('cd demo && vp run');
+  });
+
+  it('omits the cd for the current directory', () => {
+    expect(getNextCommand('.', 'vp run')).toBe('vp run');
+    expect(getNextCommand('', 'vp run')).toBe('vp run');
+  });
+});
+
+describe('showCreateSummary', () => {
+  let previousExitCode: typeof process.exitCode;
+
+  beforeEach(() => {
+    terminal.log.mockClear();
+    previousExitCode = process.exitCode;
+    process.exitCode = undefined;
+  });
+
+  afterEach(() => {
+    process.exitCode = previousExitCode;
+  });
+
+  it('reports a ready project and leaves the exit code alone on success', () => {
+    showCreateSummary({ ...baseOptions, installSummary: installed, fmtSummary: formatted });
+
+    const text = output();
+    expect(text).toContain('Scaffolded');
+    expect(text).toContain('Dependencies installed');
+    expect(text).toContain('Next: ');
+    expect(text).toContain('cd create-failure-demo && vp run');
+    expect(text).not.toContain('were not installed');
+    expect(process.exitCode).toBeUndefined();
+  });
+
+  // Regression test for #2453: a failed install used to be summarized as a
+  // finished project — "Scaffolded", "Next: vp run", exit code 0 — even
+  // though node_modules was missing.
+  it('does not report a failed install as a ready project', () => {
+    showCreateSummary({ ...baseOptions, installSummary: failed, fmtSummary: failed });
+
+    const text = output();
+    expect(text).toContain('Dependencies were not installed');
+    expect(text).toContain('Code was not formatted');
+    expect(text).toContain('cd create-failure-demo && vp install');
+    expect(text).not.toContain('vp run');
+    expect(process.exitCode).toBe(1);
+  });
+
+  it('still suggests running the project when only formatting failed', () => {
+    showCreateSummary({ ...baseOptions, installSummary: installed, fmtSummary: failed });
+
+    const text = output();
+    expect(text).toContain('Code was not formatted');
+    expect(text).toContain('cd create-failure-demo && vp run');
+    expect(text).not.toContain('Dependencies were not installed');
+    expect(process.exitCode).toBe(1);
+  });
+
+  it('never lowers an exit code another step already raised', () => {
+    process.exitCode = 1;
+    showCreateSummary({ ...baseOptions, installSummary: installed, fmtSummary: formatted });
+
+    expect(process.exitCode).toBe(1);
+  });
+});
--- a/packages/cli/src/create/__tests__/summary.spec.ts
+++ b/packages/cli/src/create/__tests__/summary.spec.ts
@@ -79,14 +79,19 @@ describe('showCreateSummary', () => {
     expect(process.exitCode).toBe(1);
   });
 
-  it('still suggests running the project when only formatting failed', () => {
+  // Regression test for the PTY snapshot failures this PR first introduced:
+  // `create_framework_shim_vue` and `new_create_vite` scaffold templates whose
+  // plugin imports are not resolvable when `vp fmt` runs, so the format step
+  // fails on a project that is otherwise complete. Naming it is right; exiting
+  // non-zero for it is not.
+  it('still suggests running the project and succeeds when only formatting failed', () => {
     showCreateSummary({ ...baseOptions, installSummary: installed, fmtSummary: failed });
 
     const text = output();
     expect(text).toContain('Code was not formatted');
     expect(text).toContain('cd create-failure-demo && vp run');
     expect(text).not.toContain('Dependencies were not installed');
-    expect(process.exitCode).toBe(1);
+    expect(process.exitCode).toBeUndefined();
   });
 
   it('never lowers an exit code another step already raised', () => {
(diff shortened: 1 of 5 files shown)


m0g3r:
Thanks for approving the workflows — the run caught a real regression in this PR, and `33214c2` fixes it.

**What failed.** `CLI snapshot test` on all three platforms, 699 passed / 2 failed: `create_framework_shim_vue` and `new_create_vite`. Everything else was green, and `main` is green at both this branch's base and current head, so it was mine.

**Root cause.** Both cases scaffold a template whose generated `vite.config.ts` imports a plugin that is not installed at the moment `vp fmt` runs, so the format step fails on a project that is otherwise complete:

```
vite.config.ts (2:16) [UNRESOLVED_IMPORT] Could not resolve '@vitejs/plugin-vue' in vite.config.ts
error: Failed to resolve vite config: ... Cannot find package '@vitejs/plugin-vue'
◇ Scaffolded vite-plus-application with Vue + TypeScript
✗ Code was not formatted
→ Next: cd vite-plus-application && vp run
```

Both declare that `vp create` step as `snapshot = false`, so it recorded nothing until the step started exiting `1`.

**Why this settles the open question rather than needing a snapshot update.** The PR body asked whether a format-only failure should exit non-zero, and offered the exit-`0` variant. These two fixtures are the argument for it: a routine `vp create` of a `react-ts` or `vue-ts` template hits this path, so keeping the non-zero exit would fail every CI job that scaffolds one — and the project it produces is fine. Re-recording the snapshots would have baked that in.

So `resolveCreateCompletion` now returns `exitCode: installFailed ? 1 : 0`. The `✗ Code was not formatted` line stays — dropping it silently is what #2453 is about — and `Next:` still reads `vp run`. Only a missing `node_modules` fails the command. Both snapshots should return to their recorded state untouched, since an exit-`0` `snapshot = false` step records nothing.

Verified in both directions: with the old expression restored, exactly the two guarding tests fail (`2 failed | 53 passed`); with the fix, `55 passed`. `oxlint` and `oxfmt` clean against this repo's config. I still cannot run the PTY suite in my environment, so that part rests on root-causing both recorded diffs rather than a local run — worth a second look on your side.

I have also corrected the description, which claimed I would not expect snapshot churn here.

---

PR #2550, opened 2026-08-24
chore(docs): remove unused components and assets

jong-kyung:
## Summary

`PartnerLogos`, `Testimonials`, and `HeadingSection2` were imported by the homepage but never rendered, while `Error404` was referenced only by a commented-out import. Since these components appear to be unused, this PR removes them along with the empty testimonial data and the now-unused `vue3-carousel` dependency.

The unreferenced local image is also removed.


jong-kyung:
@codex review

fengmk2 (approved):

PR #2557, opened 2026-08-25
fix(migrate): remove scoped Prettier plugins and shareable configs

m0g3r:
## Problem

`vp migrate` removes Prettier itself, but a migrated project keeps installing dead Prettier
packages under most naming shapes. `rewritePrettierPackageJson` matched only the exact name
`prettier` and the flat `prettier-plugin-*` prefix, so all of these survived migration:

- scoped plugins — `@trivago/prettier-plugin-sort-imports`, `@ianvs/prettier-plugin-sort-imports`,
  `@shopify/prettier-plugin-liquid`
- the official `@prettier/` scope — `@prettier/plugin-php`, `@prettier/plugin-xml`,
  `@prettier/plugin-ruby`
- shareable configs — `prettier-config-*`, `@scope/prettier-config`
- `@types/prettier`

The asymmetry looks arbitrary rather than intentional: unscoped `prettier-plugin-sort-imports` *is*
removed, while the scoped package of the same name is not.

The sibling ESLint path already gets this right. `isEslintEcosystemDep` in
`migrator/eslint.ts` recognises named packages, flat prefixes, whole scopes, the scoped
`^@[^/]+\/eslint-(plugin|config|formatter)(-.+)?$` form, and treats `@types/NAME` symmetrically
with `NAME`.

## Changes

Adds `isPrettierEcosystemDep` mirroring `isEslintEcosystemDep`, and routes both dependency loops
through it.

The matched set is deliberately narrow, and a test pins that: `eslint-config-prettier`,
`eslint-plugin-prettier`, `prettierx`, and an unrelated `@prettierx/` scope are all preserved.

Also deliberately unchanged:

- **the fields scanned.** `peerDependencies` and `optionalDependencies` are still left alone. The
  ESLint path does scrub them, but whether migration should silently drop a *declared peer*
  dependency is exactly what a P1 on #2483 is currently about, so I have not imported that decision
  here.
- **empty containers.** The ESLint path deletes a dependency field it empties; this one still
  leaves `"devDependencies": {}`. Same reasoning — separate concern, separate PR if wanted.

`rewritePrettierPackageJson` becomes exported. `migrator/README.md` rule 1 asks for module functions
to be exported so the barrel surfaces them, and it makes the function unit-testable exactly like
`rewriteEslintPackageJson`.

## Testing

Six cases added to `migrator.spec.ts`, next to the `rewriteEslintPackageJson` block and following
its conventions.

Verified in **both directions**: with the old matcher restored and the new tests kept, **4 fail**;
with the fix, **6/6 pass**. The two that pass either way are the guards for behaviour that is meant
to be unchanged (unrelated-name preservation, and the no-op case).

Checks actually run:

- `migrator.spec.ts` — before/after diffed in the same build state: **identical failure set both
  ways (139)**, with the change adding exactly 6 passing tests (206 → 212).
- `oxlint@1.79.0` (the repo's pin) with `-D correctness -D perf -D suspicious` on both changed
  files — no new findings. The one reported error is pre-existing and byte-identical on the
  unmodified files, and is a rule this repo sets to `off` in `vite.config.ts`.
- `oxfmt@0.64.0` (the repo's pin), configured from this repo's `fmt` block — clean, and clean on
  untouched control files too.
- `git diff --check` — clean.

Not run: `vp check`, `pnpm test:unit`, `tsgo`, and the PTY snapshot suite, which need a full
workspace build my environment cannot produce. No PTY fixture covers `rewritePrettierPackageJson`,
and this change adds no CLI output.

## AI assistance

Claude Opus 5 wrote the implementation, the tests, and this description. The change is
agent-authored and has not had a separate human review. Every result quoted above is from an actual
run, not an estimate.

### DIFF
--- a/packages/cli/src/migration/__tests__/migrator.spec.ts
+++ b/packages/cli/src/migration/__tests__/migrator.spec.ts
@@ -49,6 +49,7 @@ const {
   ensureSvelteRuneGlobals,
   mergeViteConfigFiles,
   rewriteEslintPackageJson,
+  rewritePrettierPackageJson,
   collectInstalledPackageNames,
   sanitizeMigratedOxlintConfig,
   detectIncompatibleEslintIntegration,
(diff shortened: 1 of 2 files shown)


fengmk2:
@codex review

PR #2559, opened 2026-08-25
fix(deps): sync Vite DevTools version

fengmk2:
`@vitejs/devtools` must use the version ranges from the pinned Vite package. The upgrade script currently gets the latest release from npm.

This behavior can select a DevTools version that Vite does not support.

`pnpm tool sync-remote` now copies the development and peer ranges from `vite/packages/vite/package.json`. It copies these ranges before `pnpm install`.

The generic dependency upgrade script no longer updates `@vitejs/devtools`.

Vite `8.2.2` declares the development range `^0.4.12`. It declares the peer range `^0.4.0 || ^0.5.0`.

The core build now supports Rolldown output with local `ansis` helpers. This change keeps the build compatible with DevTools `0.4.12`.


fengmk2:
test https://github.com/voidzero-dev/vite-plus/actions/runs/32869785033

Brooooooklyn (approved):

PR #2568, opened 2026-08-26
feat(env): support bun as a devEngines runtime

BlankParticle:
We have a repo that uses `bun` for running scripts and somethings along with node in parallel since we aim to support both runtimes, we use `pnpm` as the package manager. So I have attempted to write something like this in package.json
```json
  "devEngines": {
    "packageManager": { "name": "pnpm", "version": "10.20.0", "onFail": "download" },
    "runtime": [
      { "name": "bun", "version": "1.3.13", "onFail": "download" },
      { "name": "node", "version": "24.5.0", "onFail": "download" }
    ]
  }
```
this resolves `pnpm` as the package manager, and `node` version as runtime, but doesn't put the correct `bun` version on `PATH`. I expected this since Vite+ says it doesn't want to manage runtimes, but still allows to use `bun` as package manager. I asked in Voidzero discord and @TheAlexLichter confirmed too.
I still want to migrate our repo to Vite+, so I took a stab at this with help of Claude Fable 5, and the implementation came out at 350 lines, the mechanism for managing/downloading `bun` already exists, only the dev engines had to be hooked up. This doesn't mean Vite+ does everything through bun, it only makes it so the correct version of `bun` is available in path and the scripts in package.json and invoking bun from the shell downloads and uses correct bun version.
Screenshot 2026-08-26 at 14-51-11

I have cleaned up the code and opened this PR on suggestion of @TheAlexLichter so we can have a discussion on this, the behaviour in this PR is what I exactly need. I can tweak how its implemented based on feedback.

Discord Thread https://discord.com/channels/1475973262193459293/1540305126269521941

### DIFF
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/shim_dev_engines_runtime_bun/package.json
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/shim_dev_engines_runtime_bun/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "shim-dev-engines-runtime-bun",
+  "version": "1.0.0",
+  "private": true,
+  "devEngines": {
+    "packageManager": { "name": "pnpm", "version": "10.11.0", "onFail": "download" },
+    "runtime": [
+      { "name": "bun", "version": "1.3.11", "onFail": "warn" },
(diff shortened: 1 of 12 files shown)


fengmk2:
We currently don’t have plans to support JS runtimes other than Node.js. Bun is currently supported as a package manager.

PR #2574, opened 2026-08-28
fix(global-cli): bound local vite-plus resolution to the workspace root

Akokk0:
A nested project can run an ancestor's `vite-plus` when its own installation is missing. `vp --version` can also display that unrelated installation.

Use `oxc_resolver` with `Restriction::Fn` in the shared `vp_local_cli` crate to enforce the package or workspace boundary. Workspace members can use the root installation. Excluded projects that declare `vite-plus` stop at their own package root. Undeclared projects retain ancestor lookup.

Command delegation, migration version checks, and both CLI version entrypoints use the shared resolver. The NAPI binding replaces the separate JavaScript lookup. Read versions from the manifest that Oxc has already parsed.

The restriction checks the lookup path before Oxc resolves symlinks, so local links to external package stores remain valid. Add unit tests and direct CLI output snapshots for boundaries, exports, malformed manifests, and symlinks.

Link the local CLI inside migration fixtures that need it. Use an existing shell in installer tests to avoid `ETXTBSY` on Linux and support BusyBox.

An existing limitation remains: malformed ancestor manifests can prevent resolution in a workspace without a root `package.json`.


Akokk0:
The red CI is a real regression from this PR. All 75 snapshot failures have the same cause: the snapshot harness links the only real `node_modules/vite-plus` at the run root, and each staged workspace (which carries its own `package.json`/`pnpm-workspace.yaml`) resolves it via Node's unbounded upward walk — noted in `cli_snapshots/main.rs` around L1571. The workspace boundary added here stops the walk at the stage, so the install two levels up is never found.

I'll push a narrower fix: the boundary only applies when the project itself declares a `vite-plus` dependency (the case where the existing `vp install` warning is the right answer). The fixtures don't declare it, so they keep the unbounded walk.

fengmk2:
@codex review

Akokk0:
The remaining red is this PR: the same 3 snapshot cases on every platform — `migration_eslint_svelte_runes`, `migration_pack_tsdown_023_build`, `migration_pack_tsdown_023_concise_methods` (the Linux shard 1 failure is `apt-get`, unrelated). Each differs by one line:

  +warn: No project-local vite-plus installation was found. Run vp install in <workspace> to install dependencies.

Cause: these fixtures declare `vite-plus` (pre-declared, or added by `vp migrate`) and then run `vp lint` / `vp pack` / `vp run build` without installing. The harness relies on the run-root `node_modules` being found by the upward walk, and this PR stops that walk at the workspace root for projects that declare `vite-plus`. My earlier "the fixtures don't declare it" was wrong for these three;

I don't know how to resolve this.

fengmk2:
@Akokk0 Let me take over.

PR #2592, opened 2026-09-01
docs: fix Zed Oxc configuration

joschuba:
Fix the Zed configuration example so source.fixAll.oxc is configured as a formatter code action instead of an invalid language-level setting. The Oxc Zed example configuration is doing it the same way: [oxc-project/oxc-zed example configuration](https://github.com/oxc-project/oxc-zed/blob/main/examples/both/.zed/settings.json?utm_source=chatgpt.com)

Also add JSX and TSX, since Zed treats them as separate languages and the TypeScript/JavaScript formatter settings don't apply to them. This tripped me up while setting up the integration just now, since my TSX files were still being formatted with the default settings.

### DIFF
--- a/docs/guide/ide-integration.md
+++ b/docs/guide/ide-integration.md
@@ -82,10 +82,22 @@ You can also manually set up the Zed config:
     "JavaScript": {
       "format_on_save": "on",
       "prettier": { "allowed": false },
-      "formatter": [{ "language_server": { "name": "oxfmt" } }],
-      "code_action": "source.fixAll.oxc"
+      "formatter": [
+        { "language_server": { "name": "oxfmt" } },
+        { "code_action": "source.fixAll.oxc" }
+    ]
+    },
+    "JSX": {
+      "format_on_save": "on",
+      "prettier": { "allowed": false },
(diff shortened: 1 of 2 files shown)


docs/guide/ide-integration.md
```diff
@@ -82,14 +82,26 @@ You can also manually set up the Zed config:
     "JavaScript": {
       "format_on_save": "on",
       "prettier": { "allowed": false },
-      "formatter": [{ "language_server": { "name": "oxfmt" } }],
-      "code_action": "source.fixAll.oxc"
+      "formatter": [
+        { "language_server": { "name": "oxfmt" } },
+        { "code_action": "source.fixAll.oxc" }
```
fengmk2, line 87: https://github.com/voidzero-dev/vite-plus/blob/main/packages/cli/src/utils/editor.ts#L90-L92 Could you also fix the code related to `vp migrate` at the same time?

joschuba: For sure, added it in the newer commit!

fengmk2 (approved):

fengmk2:
@joschuba Thanks, I will fix format issue before merge.

PR #2600, opened 2026-09-02
feat(cli): expose dependency alignment plans

afonsojramos:
## Summary

Vite+ releases can now tell external automation how to align Vite+, Vite+ core, Vitest, and official Vitest packages without reading or changing the project itself. This provides the release-owned compatibility contract needed to prevent dependency bots from producing mixed toolchains.

This is the first part of a three-repository integration. Containerbase installation and Renovate reconciliation follow after this contract ships in a Vite+ release.

## Design

- `vp sync-versions --json` accepts bounded manifest snapshots over stdin and returns deterministic before/after replacements.
- Planning updates only existing managed declarations. It preserves catalog protocols, JSON formatting, and YAML comments, quoting, indentation, and line endings.
- The command does not load project configuration, run installs or lifecycle scripts, format source, or write to the repository.
- Both the npm CLI and standalone archive expose the same protocol. Release archives include the self-contained planner, toolchain manifest, and checksums needed for verified installation.

## Validation

- Planner, protocol, package-override, and migration suites: 395 tests passed.
- Focused sync-version suites: 39 tests passed.
- Global CLI Rust suite: 486 tests passed, 2 ignored.
- CLI TypeScript build, optimized Rust build, npm entrypoint smoke test, and standalone archive-layout simulation passed.
- Scoped formatting, lint, type checks, and `git diff --check` passed.

Related: #2356

Downstream drafts:

- containerbase/base#7338
- renovatebot/renovate#45630

### DIFF
--- a/packages/cli/src/sync-versions/bin.ts
+++ b/packages/cli/src/sync-versions/bin.ts
@@ -18,6 +18,12 @@ async function main(): Promise<void> {
     throw new Error('Usage: vp sync-versions --json');
   }
 
+  if (process.stdin.isTTY) {
+    throw new Error(
+      'Expected a JSON request on stdin. Pipe the request to this command; it is intended for external automation.',
+    );
+  }
+
   const [requestJson, manifestJson] = await Promise.all([
     readBoundedUtf8(process.stdin),
     fs.readFile(new URL('../toolchain.json', import.meta.url), 'utf8'),
--- a/packages/cli/src/sync-versions/bin.ts
+++ b/packages/cli/src/sync-versions/bin.ts
@@ -14,8 +14,8 @@ function syncVersionsArgs(argv: readonly string[]): string[] {
 
 async function main(): Promise<void> {
   const args = syncVersionsArgs(process.argv);
-  if (args.length !== 1 || args[0] !== '--plan-json') {
-    throw new Error('Usage: vp sync-versions --plan-json');
+  if (args.length !== 1 || args[0] !== '--json') {
+    throw new Error('Usage: vp sync-versions --json');
   }
 
   const [requestJson, manifestJson] = await Promise.all([
--- a/packages/cli/src/sync-versions/bin.ts
+++ b/packages/cli/src/sync-versions/bin.ts
@@ -1,0 +1,41 @@
+import fs from 'node:fs/promises';
+
+import { readBoundedUtf8 } from './input.ts';
+import { runSyncVersionsProtocol } from './protocol.ts';
+
+function errorMessage(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+function syncVersionsArgs(argv: readonly string[]): string[] {
+  const args = argv.slice(2);
+  return args[0] === 'sync-versions' ? args.slice(1) : args;
+}
+
+async function main(): Promise<void> {
+  const args = syncVersionsArgs(process.argv);
+  if (args.length !== 1 || args[0] !== '--plan-json') {
+    throw new Error('Usage: vp sync-versions --plan-json');
+  }
+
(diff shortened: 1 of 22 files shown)


packages/cli/src/sync-versions/bin.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+function syncVersionsArgs(argv: readonly string[]): string[] {
+  const args = argv.slice(2);
+  return args[0] === 'sync-versions' ? args.slice(1) : args;
+}
+
+async function main(): Promise<void> {
+  const args = syncVersionsArgs(process.argv);
+  if (args.length !== 1 || args[0] !== '--plan-json') {
```
fengmk2: Can the parameter name be changed to `--json`? Keep it consistent with the current parameter naming style.

afonsojramos: Yes, that makes sense. I’ll update the Vite+ command and the Renovate draft together so they stay aligned.

fengmk2:
@afonsojramos I installed the preview build version, and when I run `vp sync-versions --json`, it hangs without any response.

image

afonsojramos:
So, the command expects a JSON request on stdin, but silently waiting on an interactive terminal is definitely a bug. I’ll make it fail immediately, does that sound good @fengmk2?

fengmk2:
@afonsojramos Yes, I think it would be more appropriate to report an error in interactive mode.

afonsojramos:
@fengmk2 should be good now 👌

fengmk2 (approved):
@afonsojramos Thanks so much! 
I will perform manual testing and verification before merging.

afonsojramos:
Awesome, thanks! Feel free to check the other draft PRs to check the plan of joining it all together

packages/cli/src/sync-versions/plan.ts
```diff
@@ -0,0 +1,402 @@
...
+    return toolchain.vitePlus;
+  }
+  if (name === 'vitest' || isAlignableVitestEcosystemPackage(name)) {
+    return toolchain.vitest;
+  }
+  return undefined;
+}
+
+function alignedSpec(name: string, current: string, toolchain: SyncVersionsToolchain): string {
+  const target = targetVersion(name, toolchain);
```
fengmk2, line 177: https://github.com/npmx-dev/npmx.dev/blob/main/pnpm-workspace.yaml#L32

Will the overrides configuration for this dependency alias `'vite': 'npm:@voidzero-dev/vite-plus-core@0.3.0'` be automatically upgraded by us?

packages/cli/src/sync-versions/plan.ts
```diff
@@ -0,0 +1,402 @@
...
+      if (invalidPath) {
+        context.addIssue({
+          code: 'custom',
+          path: ['manifests', index, 'path'],
+          message: 'Manifest paths must be normalized workspace-relative POSIX paths',
+        });
+      }
+
+      const basename = path.posix.basename(manifest.path);
+      const kindMatchesPath =
```
fengmk2, line 63: How about `.yarnrc.yml` file? I think we should support yarn 4.10+ catalogs in this file.

afonsojramos: Good call, added. sync-versions now handles both default and named Yarn catalogs in .yarnrc.yml, while preserving formatting and comments.

packages/cli/src/sync-versions/plan.ts
afonsojramos: Yes, it will now. I added support for managed npm: aliases, including this exact pnpm catalog entry. We also preserve the alias target and only replace its version. So custom aliases remain untouched.

PR #2601, opened 2026-09-03
ci: gate npm publishes on dependency propagation

fengmk2:
`npm publish` can finish before npm completes its scan. Package metadata and tarballs can remain unavailable during this period.

The release now publishes packages in dependency order. It waits for each exact version and tarball before it publishes a dependent package. The process waits another 60 seconds for npm CDN propagation.

A retry skips a version that npm accepted in an earlier attempt. If npm still hides that version, the retry detects the immutable-version error and continues.

This change follows [Cloudflare Workers SDK PR #15166](https://github.com/cloudflare/workers-sdk/pull/15166) and addresses [GitHub Community discussion #203413](https://github.com/orgs/community/discussions/203413).


fengmk2:
cc @shulaoda

cpojer (approved):

PR #2612, opened 2026-09-05
fix(deps): keep automatic Vitest upgrades on v4

fengmk2:
The dependency upgrade workflow now selects the newest stable `vitest` release in the supported `4.x` major. It does not select `vitest@5` until the integration supports that major.

The workflow also tells its repair agent to keep the version cap. Unit tests cover stable version ordering, prerelease exclusion, and missing stable releases.


fengmk2:
https://github.com/voidzero-dev/vite-plus/actions/runs/33974290733

fengmk2:
works https://github.com/voidzero-dev/vite-plus/pull/2613

cpojer (approved):

PR #2617, opened 2026-09-06
fix(cli): share Vite runtime through the core alias

### LINKED ISSUE #1391: TanStack Start SSR broken: dual-copy of vite-plus-core causes `isRunnableDevEnvironment()` instanceof failure
## Summary

When a project uses `"vite": "npm:@voidzero-dev/vite-plus-core@latest"` (the recommended alias)
alongside `"vite-plus": "latest"`, bun (and likely npm) installs two separate physical copies of
`@voidzero-dev/vite-plus-core`:

1. `node_modules/vite/` — via the npm alias
2. `node_modules/@voidzero-dev/vite-plus-core/` — via vite-plus's direct dependency

These two copies are different module instances in Node.js ESM, which breaks any `instanceof`
checks across them. Specifically, TanStack Start's SSR middleware fails to install because
`isRunnableDevEnvironment()` (exported from `vite`) returns `false` — even though the environment
IS a RunnableDevEnvironment — because the class was instantiated from copy 2 but the instanceof
check runs against copy 1.

**Result**: `Cannot GET /` on every route. SSR is completely broken.

## Root cause

`vite-plus` (packages/cli) depends on `@voidzero-dev/vite-plus-core` directly:

```json
"dependencies": {
  "@voidzero-dev/vite-plus-core": "0.1.18"
}
```

The `vp` CLI loads Vite internals from `node_modules/@voidzero-dev/vite-plus-core/`, while
user-land Vite plugins (like TanStack Start) import from `node_modules/vite/` (the alias).

Node.js ESM identifies modules by their file URL. Two different paths = two module instances =
`instanceof` fails across them.

## Reproduction

1. `vp new` → select TanStack Start template
2. `vp dev`
3. Open `http://localhost:5173/` → `Cannot GET /`

Or manually:

```json
{
  "dependencies": {
    "@tanstack/react-start": "latest"
  },
  "devDependencies": {
    "vite": "npm:@voidzero-dev/vite-plus-core@latest",
    "vite-plus": "latest"
  },
  "overrides": {
    "vite": "npm:@voidzero-dev/vite-plus-core@latest"
  },
  "packageManager": "bun@1.3.12"
}
```

## Debug evidence

Adding a debug plugin to vite.config.ts:

```ts
import { isRunnableDevEnvironment } from "vite";

configureServer(server) {
  const ssrEnv = server.environments?.["ssr"];
  console.log("constructor:", ssrEnv.constructor?.name);      // "RunnableDevEnvironment"
  console.log("isRunnable:", isRunnableDevEnvironment(ssrEnv)); // false ← BUG
  console.log("has runner:", "runner" in ssrEnv);               // true
}
```

The constructor name IS `RunnableDevEnvironment`, but `instanceof` returns `false` because
the class references come from different module instances.

Confirming two separate copies:

```bash
$ stat -f "%i" node_modules/vite/dist/vite/node/chunks/node.js \
               node_modules/@voidzero-dev/vite-plus-core/dist/vite/node/chunks/node.js
22200352   # different inodes
22208051
```

## Workaround

Symlinking `node_modules/@voidzero-dev/vite-plus-core` → `node_modules/vite` in a postinstall
script deduplicates them into one module instance and fixes SSR.

## Suggested fix

Option A (minimal): Make `vite-plus` CLI depend on `vite` instead of `@voidzero-dev/vite-plus-core`
directly. Since users are required to have `"vite": "npm:@voidzero-dev/vite-plus-core@..."` in their
project, this would resolve to the same copy.

Option B (robust): Have the `vp` CLI resolve its Vite imports through the project's
`node_modules/vite` path (using `createRequire` from the project root), ensuring it always
uses the same module instance as plugins.

## Environment

- vp: 0.1.18
- @voidzero-dev/vite-plus-core: 0.1.18
- @tanstack/react-start: 1.167.41
- Node.js: v24.14.1
- Bun: 1.3.12
- macOS (darwin-arm64)

## Related issues

- #1356 — similar dual-module problem with `@rolldown/debug` and `@vitejs/devtools`

### Reproduction

https://github.com/keyding/vite-plus-tanstack-start-ssr-broken

### Steps to reproduce

Run `bun install` followed by `vp dev`

### System Info

```shell
···

Environment:
  Version  24.14.1
  Source   lts

Tool Paths:
  node  /Users/***/.vite-plus/js_runtime/node/24.14.1/bin/node
  npm   /Users/***/.vite-plus/js_runtime/node/24.14.1/bin/npm
  npx   /Users/***/.vite-plus/js_runtime/node/24.14.1/bin/npx
···

···

vp v0.1.18

Local vite-plus:
  vite-plus  v0.1.18

Tools:
  vite             v8.0.8
  rolldown         v1.0.0-rc.15
  vitest           v4.1.4
  oxfmt            v0.45.0
  oxlint           v1.60.0
  oxlint-tsgolint  v0.20.0
  tsdown           v0.21.8

Environment:
  Package manager  bun v1.3.12
  Node.js          v24.14.1
···
```

### Used Package Manager

bun

### Logs

```shell

```

### Validations

fengmk2:
TanStack Start can return HTTP 404 when the CLI and plugins load separate core copies. The CLI now uses the exact `vite` alias for its dependency, runtime imports, and generated exports.

Version checks use the installed CLI manifest and the command's target project. They reject incompatible declared dependencies and permit incidental hoisted peers. The public config types retain `test` when Vitest resolves a separate Vite copy.

Preview packing and CI tarball installation preserve the alias. Version checks also support packages whose versions change after the build. The design is in `rfcs/vite-core-module-identity.md`.

Fixes #1391.

### DIFF
--- a/.github/workflows/publish-preview.yml
+++ b/.github/workflows/publish-preview.yml
@@ -98,6 +98,15 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      # The registry bridge rewrites dependencies by package name. Stamp core
+      # before pnpm pack so the dependency named `vite` already aliases the
+      # preview version. Keep the CLI manifest aligned with its built dist too.
+      - name: Set preview package versions
+        run: |
+          for package in cli core; do
+            pnpm exec json-edit "packages/$package/package.json" '_.version = process.env.VERSION'
+          done
+
       - name: Download cli dist
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/core_command_root/check-layout.mjs
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/core_command_root/check-layout.mjs
@@ -1,0 +1,12 @@
+import assert from 'node:assert/strict';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const appRequire = createRequire(new URL('./packages/app/package.json', import.meta.url));
+const cliRequire = createRequire(require.resolve('vite-plus/package.json'));
+console.log('Root Vite:', require('vite/package.json').name);
+console.log('App Vite:', appRequire('vite/package.json').name);
+console.log('CLI Vite:', cliRequire('vite/package.json').name);
+assert.equal(require('vite/package.json').name, 'vite');
+assert.equal(appRequire('vite/package.json').name, '@voidzero-dev/vite-plus-core');
+assert.equal(cliRequire('vite/package.json').name, '@voidzero-dev/vite-plus-core');
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/core_command_root/package.json
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/core_command_root/package.json
@@ -1,0 +1,13 @@
+{
+  "name": "core-command-root",
+  "private": true,
+  "type": "module",
+  "packageManager": "pnpm@11.24.0",
+  "workspaces": [
+    "packages/*"
+  ],
+  "devDependencies": {
+    "vite": "8.2.2",
+    "vite-plus": "latest"
+  }
+}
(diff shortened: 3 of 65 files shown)


sapphi-red (approved):
The cause and the fix makes sense to me

fengmk2:
@codex review

PR #2624, opened 2026-09-07
fix(pm): support authenticated private registries

### LINKED ISSUE #2603: Support authenticated private registries beyond NPM_CONFIG_REGISTRY
### Discussed in https://github.com/voidzero-dev/vite-plus/discussions/2599

<div type='discussions-op-text'>

<sup>Originally posted by **xmlking** September  3, 2026</sup>
I am in corporate environment where public npm registry is blocked. 
even If I set `NPM_CONFIG_REGISTRY` , `vp install` still failing as it is not using the `apikey` 

here is what Claude say: 
---
Root cause found. Your package.json pins "packageManager": "pnpm@11.24.0" (package.json:52). Like Corepack, vp downloads that exact pnpm binary into its own managed store (~/.vite-plus/package_manager/pnpm, currently empty on your machine) rather than using whatever pnpm is on PATH. That download is done by vp's own Rust binary hitting the npm-style tarball URL directly — it's a separate code path from the pnpm/npm CLI that reads ~/.npmrc.

Verified this in two parts:
1. Registry choice: vp only honors the NPM_CONFIG_REGISTRY env var for this download (documented default https://registry.npmjs.org). Since that env var isn't exported in your shell (only present in ~/.npmrc), vp falls back to public npm.
2. Auth: even after exporting NPM_CONFIG_REGISTRY to your Artifactory URL, the download 401s — vp's downloader doesn't send the Bearer token from your .npmrc's _authToken. I confirmed with curl that the same URL only succeeds with an Authorization: Bearer <token> header; Basic auth (token-as-username) fails too, so there's no URL-embedded-credential workaround.

So this is a real gap: for a pinned packageManager version, vp can't authenticate to a token-only private registry when fetching the manager itself.

Practical options:
- Ask your Artifactory admin if the mirror also accepts Basic auth with your AD username + an API key/identity token (common JFrog setup) — if so, `NPM_CONFIG_REGISTRY=https://<user>:<apikey>@xyz.jfrog.io/...` would let vp's downloader authenticate via URL-embedded Basic auth.
- Pre-seed the cache once from a machine/network that can reach the registry with proper auth (e.g. have plain pnpm dlx pnpm@11.24.0 --version or similar populate ~/.vite-plus/package_manager/pnpm manually), then subsequent vp install runs won't need to re-download it.
- Report this to vite-plus (viteplus.dev) as a gap — package-manager bootstrap doesn't support authenticated private registries beyond NPM_CONFIG_REGISTRY.

---</div>

RSS1102:
## Problem

Vite+ respects `NPM_CONFIG_REGISTRY` when downloading a pinned package manager, but previously ignored registry credentials from `.npmrc`. Authenticated corporate registries therefore returned `401` before the requested package manager could run.

## Behavior

Registry configuration is merged in this order:

```text
explicit globalconfig < user .npmrc < workspace .npmrc < NPM_CONFIG_*
```

Package routing then uses `@scope:registry`, followed by `registry`, and finally the public npm registry.

Authentication uses the same configuration-source order. Credentials are matched to the final request URL by host, port, and longest path, with `_authToken`, `_auth`, and `username`/`_password` support. Matching credentials are sent on the first metadata and tarball requests.

`NPM_TOKEN` remains an environment-variable value referenced by a registry-scoped `.npmrc` entry:

```ini
registry=https://npm.corp.example/repository/npm/
//npm.corp.example/repository/npm/:_authToken=${NPM_TOKEN}
```

This keeps credentials bound to their intended registry instead of treating one token as valid for every configured endpoint.

## Validation

- `cargo test -p vp_pm_cli --lib` (`759 passed`, `2 ignored`)
- `cargo clippy -p vp_pm_cli --lib --tests -- -D warnings`
- `git diff --check`

Closes #2603


fengmk2:
@codex review

PR #2631, opened 2026-09-08
fix(env): track injected tools across nested shim calls

### LINKED ISSUE #2630: Nested npx calls to pnpm, Bun, and Yarn bypass managed resolution in vp 0.3.1
### Describe the bug

In vp 0.3.1, `npx` spawning a vp-managed `pnpm`, `bun`, or `yarn` fails with a recursion error. This also affects `pnpx`, `bunx`, and `yarnpkg`.

Originally encountered through `npx → taze → pnpm install`; taze is not required to reproduce.

### Reproduction

With vp-managed shims enabled and no standalone pnpm/Bun/Yarn on PATH, run in an empty project:

```sh
printf '{"private":true}\n' > package.json
printf '22.18.0\n' > .node-version

npx --offline --call 'pnpm --version'
npx --offline --call 'bun --version'
npx --offline --call 'yarn --version'
```

Each fails with the corresponding tool name:

```text
vp: Recursion detected but no 'pnpm' found in PATH (excluding shims)
```

Expected: resolve each package manager normally using its project pin or configured default.

If a standalone tool is available on PATH, the nested call instead runs it even in managed mode. Reproduced for all three using sentinel executables.

### Cause

`npx` sets `VP_TOOL_RECURSION`, but prepares only Node/npm paths. The nested package-manager shim then [skips version resolution and searches PATH](https://github.com/voidzero-dev/vite-plus/blob/v0.3.1/crates/vp_global_cli/src/shim/dispatch.rs#L724-L729), although that manager's real binary directory may not be present.

### Environment

- vp 0.3.1 (unmodified release binary)
- Node.js 22.18.0
- macOS 26.6.2, arm64
- Reproduced with isolated HOME, VP_HOME, and PATH

lyzno1:
Nested package-manager calls can fail or select a system version because the boolean recursion guard assumes every tool is already on PATH.

Replace it with `VP_PATH_INJECTED_TOOLS` and shared child PATH preparation. Track available tools per installation, exclude other Vite+ shims, and recover missing bundled tools from the selected Node installation. Explicit `vp env exec` resolves a fresh selection, while nested shim calls preserve the parent's runtime and package-manager precedence through JS delegation.

Fixes #2630.


lyzno1:
@liangmiQwQ I've updated this draft to track injected tools with `VP_PATH_INJECTED_TOOLS` and shared PATH helpers. Could you take a look and let me know whether the approach matches what you had in mind?

lyzno1:
Per-tool PATH tracking is in place, but a forwarding loop remains between renamed Vite+ installations. The current shim detection cannot reliably distinguish these from shared external-manager entrypoints such as mise. This PR remains in draft while I revisit that boundary.

liangmiQwQ:
> Per-tool PATH tracking is in place, but a forwarding loop remains between renamed Vite+ installations. The current shim detection cannot reliably distinguish these from shared external-manager entrypoints such as mise. This PR remains in draft while I revisit that boundary.

I'll take a look soon

liangmiQwQ:
> Per-tool PATH tracking is in place, but a forwarding loop remains between renamed Vite+ installations. The current shim detection cannot reliably distinguish these from shared external-manager entrypoints such as mise. This PR remains in draft while I revisit that boundary.

@lyzno1 This is another problem that already exists on Vite+. It is a little bit difficult to fix and not introduced by your PR. You can mark it as ready to review if you are ready.

liangmiQwQ:
@codex review

liangmiQwQ:
It would be better to update this RFC document:

https://github.com/voidzero-dev/vite-plus/blob/d5463937c104333756f58c92e84ca9e8be7a757e/rfcs/vpx-command.md#L307

liangmiQwQ (approved):
Others LGTM 💯

Huge thanks for working on this!

fengmk2 (approved):
@lyzno1 Thanks for the big refactor!

fengmk2:
@lyzno1 please fix the conflicts, thanks.

PR #2648, opened 2026-09-10
fix(cli): enable compile cache before importing the CLI

pablog12:
The npm `vp` wrapper calls `enableCompileCache()`, but its static import loads and evaluates the CLI dependency graph before that call executes. Use an awaited dynamic import so the existing cache initialization happens first.

The regression test runs the real wrapper in an isolated ESM fixture. It fails on the previous code (`getCompileCacheDir()` is undefined during CLI module evaluation) and passes with this change. It also checks arguments containing spaces and preserves a nonzero exit code.

This affects the npm wrapper; it does not claim to optimize the native/global entrypoint that bypasses it. Downstream warm-cache measurements on Linux/Node 24.19 showed approximately 36 ms less time for a full check and 42 ms for a three-file check; these are project-specific measurements, not a universal benchmark.

Validation on the submitted commit, using frozen pnpm 11.24.0 installations for the workspace and docs, the pinned Rolldown/Vite source revisions, and nightly-2026-08-02 Rust:

- `pnpm build`: complete source build passed, including Rolldown, Vite/core, CLI JavaScript and the native CLI binding.
- Full unit suite through the freshly built npm wrapper: **1,188 passed, one platform skip**, across 77 files on Vitest 4.1.11 (5.34 seconds).
- Full repository native TypeScript check (`tsgo --noEmit`): passed after installing the separately documented docs dependencies.
- `vp check packages/cli/src/__tests__/bootstrap-compile-cache.spec.ts`: formatting, type-aware lint and type checks passed with the repository's configuration.
- `node --check packages/cli/bin/vp` and the built wrapper's `--version` command: passed.

The ordinary `pnpm test:unit` invocation initially had two environment-dependent failures: pnpm's shim injected `NODE_PATH`, allowing a deliberately missing-dependency fixture to resolve an unrelated hoisted Vite; a hook test assumes `/usr/bin/dirname` and an ordinary `/bin/sh`, whereas this NixOS host has no `/usr/bin/dirname` and wraps `/bin/sh` with a modified PATH. All tests passed when invoking `node packages/cli/bin/vp test --run --maxWorkers=4` with `NODE_PATH`, `NO_COLOR` and `FORCE_COLOR` cleared, inside a temporary Bubblewrap environment supplying standard coreutils and a plain Bash `/bin/sh`. No production code, test assertions, snapshots or host system paths were changed to accommodate these failures.

The whole Rust workspace test suite, PTY/browser snapshot matrix and cross-platform CI matrix were not run locally.

After the repository's documented dependency setup, run the regression test with:

```sh
pnpm test:unit packages/cli/src/__tests__/bootstrap-compile-cache.spec.ts
```

This is an isolated bootstrap unit test: the CLI module is a fixture, so it tests ESM initialization order without exercising command output or the native CLI.

AI assistance: implementation and self-review performed with OpenAI Codex using GPT-6 Astra (`gpt-6-astra`), with reasoning effort set to `high`.


pablog12:
Thanks for enabling the broader CI matrix. I investigated the failing [Windows Socket Firewall job](https://github.com/voidzero-dev/vite-plus/actions/runs/34448505230/job/102778847292):

- `sfw --version` reports 1.15.1.
- The first command, `sfw "vp.exe" i -g pnpm@9.15.0`, prints `Vite+ already includes 'pnpm'; skipping.`, then aborts with `Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76` (exit 127). It never reaches the subsequent clone/install steps.
- The skip is the existing [native global-install early return](https://github.com/voidzero-dev/vite-plus/blob/b2d868d8aa515ed1b4d592ebb22ecb90e3504fcf/crates/vp_global_cli/src/commands/global/install.rs#L142). This PR changes only the npm `packages/cli/bin/vp` wrapper and adds its regression test; that wrapper is not on this failing command's execution path.

This points to the Windows/SFW execution path rather than the compile-cache change, but I cannot confirm the crash's cause locally on Linux. The Linux/macOS test jobs and the ecosystem E2E/project-creation matrices passed. Could a maintainer rerun the failed jobs to check reproducibility? I attempted the rerun, but GitHub requires repository admin rights.

The fixture also currently skips its intended registry download because pnpm is bundled; that may warrant a separate CI fixture update.

fengmk2 (approved):
@pablog12 Thanks, LGTM.

Since vpr has already fixed the compile cache issue in this way https://github.com/pablog12/vite-plus/blob/b2d868d8aa515ed1b4d592ebb22ecb90e3504fcf/packages/cli/bin/vpr#L10, I will delete the newly added test case packages/cli/src/__tests__/bootstrap-compile-cache.spec.ts before merge, as it should not be necessary.

PR #2653, opened 2026-09-10
docs(readme): use the supported task env field

SaKaNa-Y:
The README task example uses the invalid `envs` field. Replace it with `env`, matching `docs/config/run.md`, so the example runs and tracks environment changes for task caching. No source code changed.


fengmk2 (approved):
@SaKaNa-Y Thanks!

README.md
```diff
@@ -79,7 +79,7 @@ export default defineConfig({
     tasks: {
       'generate:icons': {
         command: 'node scripts/generate-icons.js',
-        envs: ['ICON_THEME'],
+        env: ['ICON_THEME'],
```
fengmk2, line 82: @SaKaNa-Y should fix typo at https://github.com/SaKaNa-Y/vite-plus/blob/main/packages/cli/README.md too.

PR #2654, opened 2026-09-10
fix(migrate): finish leftover Oxc config migration

SaKaNa-Y:
When an interrupted migration has already added `vite-plus`, running `vp migrate` again can leave `.oxfmtrc.json` behind and report that the project already uses Vite+. Formatting then uses defaults instead of the unmerged options.

The existing-project finalization path skips standalone Oxc configs. Run the existing config merger for the root and workspace packages, and count completed merges as migration work. This covers Oxlint and Oxfmt JSON/JSONC configs, preserves existing inline config precedence, and retains configs that cannot be merged with a warning. A subsequent migration remains a no-op.

The branch incorporates upstream main at `b1c41b2a`, including the leftover tsdown config fix from #2646. Both finalization paths and result flags are preserved. A regression test covers simultaneous Oxfmt and tsdown config migration and an unchanged retry.

## Verification

After resolving the upstream merge:
- Migration unit suite: **436 passed across 11 files**, including the combined-config regression. Ran the source tests with Vitest 4.1.11 and an isolated config aliasing the local prompts source, using a prebuilt native binding. The native migration implementation has not changed since the binding's base revision; the complete native CLI was not rebuilt.
- Targeted Oxfmt check on the three changed TypeScript files, using the repository's format configuration: passed.
- `git diff --check`: passed.
- Targeted Oxlint type-aware/type check: seven prompt cancellation narrowing errors. Running the same check on unmodified upstream `b1c41b2a` in the same environment produced identical diagnostics.
- The full repository suite and PTY snapshot suite were not rerun after this merge.

Before the merge, the PR's recorded validation included a passing `migration_existing_oxc_configs` PTY snapshot. It covers config removal, formatting with migrated options, an unchanged second migration, and no opt-out setup artifacts.

## Existing inline fmt snapshot coverage

Added `migration_existing_oxc_configs_inline_fmt` in response to review. It checks formatting before and after migration with inline `singleQuote: false` overriding the standalone `singleQuote: true`, while `semi: false` verifies that formatting actually applies the inline options. It also verifies standalone config removal, unchanged inline configuration, and a no-op retry.

- The official PTY runner generated the snapshot; both `migration_existing_oxc_configs` cases then passed in comparison mode (2 passed).
- With the pre-fix migration entry, the new case failed specifically because `.oxfmtrc.json` still existed.
- `git diff --check` passed. This update changes only test metadata and its recorded snapshot.
- Validation rebuilt the PR's ESM JavaScript CLI and prompts and compiled the snapshot runner. It reused the existing source-built global binary, native binding, and core package; a complete native/toolchain rebuild and the full repository suite were not run.


fengmk2:
@codex review

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/migration_existing_oxc_configs/vite.config.ts
```diff
@@ -0,0 +1 @@
+export default {};
```
fengmk2, line 1: What will happen if there is already an `fmt` config field in vite.config.ts?

SaKaNa-Y: If `vite.config.ts` already has a `fmt` configuration, formatting uses it both before and after this fix. The difference is what happens when running `vp migrate`: before the fix, `.oxfmtrc.json` is left behind; after the fix, it is removed, while the existing `fmt` configuration remains unchanged.

fengmk2: Can you a one more snapshot test to cover this case?

fengmk2 (approved):

fengmk2:
@SaKaNa-Y I plan to revert this PR for now. Ecosystem tests found regressions in [freecut](https://github.com/vite-plus-ecosystem-ci/freecut/pull/9), [delta-comic](https://github.com/vite-plus-ecosystem-ci/delta-comic/pull/9), [pdfme](https://github.com/vite-plus-ecosystem-ci/pdfme/pull/7), and [sigle](https://github.com/vite-plus-ecosystem-ci/sigle/pull/8): `vp migrate` deletes Oxc configs that imports or scripts still use, or copies JSON `extends` paths into incompatible inline config.

`0.3.1` preserves these configs. We can reintroduce this migration step after #2695 addresses these cases and passes ecosystem validation.

PR #2659, opened 2026-09-10
feat(env): add per-manager shim versions and update env use

liangmiQwQ:
This PR adds `VP_NPM_VERSION`, `VP_PNPM_VERSION`, `VP_YARN_VERSION`, and `VP_BUN_VERSION` to override each package manager's direct shim version independently. Each variable covers the family's aliases and takes priority over its session file, project configuration, and default version in managed mode.

`vp env use pnpm@10.20.0` now sets only `VP_PNPM_VERSION`, following the existing `VP_NODE_VERSION` model. Switching Yarn preserves the pnpm override. Scoped `--unset pnpm` clears only pnpm, while `--unset pm` clears all four package-manager overrides. These operations neither set nor clear `VP_PACKAGE_MANAGER`, which continues to select the manager and version for commands such as `vp install`.

`vp env exec --npm` and `--package-manager` also pass explicit versions through the corresponding family variable, so child shims retain the requested version after a PATH reset.

Without a shell wrapper, versions are stored in separate `.session-<manager>-version` files. The old `.session-package-manager` file is no longer read or migrated; rerunning `vp env use` creates the new session state. Concrete-family environment queries use the shim resolution and report its source, while `vp env current pm` reports the manager selected for vp commands. Integrity suffixes are preserved.

The existing `vp-use.cmd` wrapper executes the new `set VP_*_VERSION=...` output without template changes.

🤖 Generated with Codex

### DIFF
--- a/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_env_install_no_arg/snapshots/command_env_install_standalone_npm_fallback.md
+++ b/crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_env_install_no_arg/snapshots/command_env_install_standalone_npm_fallback.md
@@ -7,7 +7,7 @@ Explicit npm family scopes use standalone registry npm; only the directly invoke
 an explicit npm scope exports the standalone npm fallback
 
 ```
-export VP_PACKAGE_MANAGER=npm@12.0.2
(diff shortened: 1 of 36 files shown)


liangmiQwQ:
@fengmk2 I'm ready for that

fengmk2 (approved):

PR #2667, opened 2026-09-11
fix(create): remove nested monorepo lint config

### LINKED ISSUE #2640: `vp create` monorepo vite.config.ts is invalid
### Description

at the moment, `vp create` monorepo config is not ideal. packages/utils `vite.config.ts` has `lint.options.typeCheck` and `lint.options.typeCheck`. but it works only on root config file. so we can remove it. `vp create` is entry of using vite+, so it should be ideal and valid.

### Suggested solution

remove that config.

### Alternative

support that option. but it's not include roadmap. so we should fix this problem with remove option IMO.

### Additional context

_No response_

### Validations

semimikoh:
- remove the generated library lint block when scaffolding a Vite+ monorepo
- preserve the lint configuration for standalone `vite:library` projects
- add an AST-based top-level config key removal helper and regression coverage

Fixes #2640


fengmk2:
Can you add a new snapshot test case to cover this bug fix?

fengmk2:
@codex review

semimikoh:
@fengmk2 I added snapshot test coverage for this bug fix. Thanks for the suggestion!

fengmk2:
@semimikoh According to the description at https://github.com/voidzero-dev/vite-plus/pull/2668, the lint and fmt configurations do not support nested configurations, so we need to further process the `fmt` field.

semimikoh:
@fengmk2 Addressed in 2722bae. Nested monorepo libraries now remove both `lint` and `fmt` configuration, with unit and snapshot coverage for both creation paths.

fengmk2 (approved):

PR #2677, opened 2026-09-12
fix(create): honor non-interactive mode in scaffolded generators

SaKaNa-Y:
Running a scaffolded local generator with `vp create <generator> --no-interactive` still entered Bingo's prompts when required arguments were missing. With closed stdin, this exited with an unsettled top-level await instead of an actionable validation error. Bingo's `--skip-requests` flag does not disable prompts.

Pass `VP_CREATE_INTERACTIVE` to local Bingo generators. The starter uses Bingo's programmatic API in non-interactive mode, validates CLI arguments and required template options before generation, and rejects existing output directories. Interactive runs, help, and version requests retain the Bingo CLI path.

This applies to newly scaffolded generators. Existing copied entrypoints require an update, and custom template options need corresponding CLI argument declarations; the starter README documents both requirements.

Validation:
- `pnpm exec vitest run packages/cli/src/create/__tests__`: 194 tests passed.
- `pnpm exec tsc --project packages/cli/templates/generator/tsconfig.json`: passed.
- `cargo test -p vp_cli_snapshots -- create_generator_`: both generator cases passed in comparison mode after recording and reviewing the updated snapshot.
- Focused formatting and lint checks passed.

The regression covers missing arguments without prompts, no output directory on validation failure, preservation of existing files, successful file generation, and the interactive execution path. Validation was scoped to generator behavior; the full repository test suite was not run.


fengmk2:
@codex review

SaKaNa-Y:
When `--directory` was provided without `--name`, Bingo created an empty directory before prompting for the missing name, leaving it behind when was closed.

fengmk2 (approved):

PR #2679, opened 2026-09-12
feat(config): add an entry without pack types

fouteox:
Importing `defineConfig` from `vite-plus` brings tsdown's declarations into the project, even when it never uses `vp pack`.

`vite-plus/config` exposes the existing configuration helpers without the pack augmentation. The root entry keeps its current behavior by re-exporting the types from `vite-plus/pack`. Static extraction and migrations recognize the new entry; a migration that adds pack switches the import back to the root.

The two native CLI snapshots check the installed packages, strict TypeScript 7, static execution and ESM/CommonJS identity. Pack integrations are checked with TypeScript 6. The 1,210 JavaScript tests, 429 affected Rust tests, repository type checks and formatting pass; the full cross-platform snapshot matrix has not been run locally.

Importing the root or pack entry elsewhere in the same TypeScript program still loads the augmentation. This leaves the optional integrations' declarations unchanged. [OXC #20416](https://github.com/oxc-project/oxc/issues/20416) previously suggested `/config` for another API, so the entry's name and scope need maintainer agreement.


fengmk2:
@fouteox Can I get more details about the reason for this change? Could you describe what problem you encountered?

fouteox:
Thanks, I should have included the original failure.

This came from a Laravel/Inertia/React application that uses Vite+ for its frontend, without `vp pack`. Removing `skipLibCheck` exposed errors from pack's optional integrations. A small example is:

```ts

export default defineConfig({
  build: { sourcemap: true },
  fmt: { semi: true },
});
```

With both `vite-plus` and `@voidzero-dev/vite-plus-core` at 0.3.0, TypeScript 7.0.2, `strict: true`, `skipLibCheck: false` and `moduleResolution: "bundler"`, this reports errors such as:

```text
TS2307: Cannot find module 'publint' or its corresponding type declarations.
TS2305: Module '"typescript"' has no exported member 'FileExtensionInfo'.
```

I rechecked with both packages at 0.3.1 from npm, and this example already passes. The bundled declarations now contain targeted `@ts-ignore` comments for these cases. I should have made clear that the original failure no longer requires this PR.

The remaining difference is which declarations TypeScript loads. With 0.3.1, `tsc --listFiles` still includes `dist/tsdown/index-types.d.ts` through the root import. Changing the import to `vite-plus/config` with this PR leaves that file out. The existing root and pack entries keep their behavior, including those suppressions; this PR does not remove them.

Given that 0.3.1 already resolves the failure, I don't have a strong enough use case to justify another public entry. I'm closing this PR. Thanks for asking for the concrete example.

PR #2683, opened 2026-09-13
fix(runtime): explain package.json parse errors

### LINKED ISSUE #2664: Improve error reporting for invalid `package.json` syntax
### Description

If the current directory's `package.json` contains a JSON syntax error, Vite+ outputs an unclear, unhelpful error message e.g. `error: Failed to download Node.js runtime: expected value at line 1 column 1`. It is not immediately clear that the issue is due to a syntax error in the `package.json` file.

### Suggested solution

Improve the error messaging to explain the issue is due to an invalid `package.json` file.

### Alternative

Perhaps more commands could be updated to allow execution regardless of `package.json` syntax issues, e.g. `vp lint`, `vp format`, etc, assuming the `node_modules` are at least already installed.

### Additional context

The following reproduces the issue:

```sh
mkdir vp-test
cd vp-test
echo '.' > package.json
vp install
```

This will output:

```

error: Failed to download Node.js runtime: expected value at line 1 column 1
```

_Everything performed using the latest Vite+ v0.3.1._

### Validations

adamaveray:
Closes #2664.

When a project’s `package.json` file contains invalid JSON, a generic invalid JSON error message is currently presented. For example, a `package.json` file containing `hello world` would result in this output when running e.g. `vp install`:

```

error: Failed to download Node.js runtime: expected value at line 1 column 1
```

This PR updates the error output to include the file path, clarifying which file caused the error, so the previous example will now display this output:

```

error: Failed to download Node.js runtime: Failed to parse <workspace>/package.json: expected value at line 1 column 1
```

I intentionally kept this PR minimal, just inserting the actual file path to help explain the issue, but future expansion could further improve handing this case and bypass the "Failed to download Node.js runtime" component but this addresses the main source of confusion.


fengmk2 (approved):
@adamaveray Thanks!

liangmiQwQ:
/cc @semimikoh

Just notice, if you haven't started your implementation.

PR #2699, opened 2026-09-14
fix(create): honor system package manager versions

SaKaNa-Y:
> [!IMPORTANT]
> Please take a moment to read this. Thank you!
>
> I should include a brief explanation of the problem in my own words in every PR. If that explanation is missing, please @mention me and do not merge this PR until I have added it. You may also leave this PR unaddressed (because this means I have not fulfilled my responsibilities as the author).
>
> If my explanation is unclear or difficult to follow, please ask me to clarify or provide reproduction steps or supporting evidence.
>
> I welcome suggestions and counterarguments, especially questions about anything I may have overlooked. (Your feedback helps me learn and improve. 🙏)
>
> I hold myself to this standard for every PR, regardless of its size.

`vp create` always downloaded a managed package-manager release before scaffolding, even when that family was configured to use the system installation. The generated `devEngines.packageManager.version` could therefore disagree with the executable used for installation. With npm, this can fail with `EBADDEVENGINES`; the same incorrect pinning also affects pnpm, Yarn, and Bun.

Resolve the selected family through `vp env current` before scaffolding. When it selects a system executable, use that version and binary directory. Keep managed downloads and the missing-system fallback, locate the global resolver through the shared directory API, and preserve local-only creation when it is absent. Narrow scaffold metadata to the fields consumers actually need.

System Yarn Classic also needs a compatible template runner: Yarn 1 has no `dlx` command. Select `npx --yes` from the resolved Yarn version, matching the existing `vp dlx` fallback; Yarn 2+ continues to use `yarn dlx`. This runner choice does not change the selected package-manager family or version.

The Unix external-template fixture sets npm's script shell to `/bin/sh`. The snapshot runner exposes `sh` through a symlink, while fspy identifies system shells by their canonical parent directory for macOS compatibility. Real npx execution and generated-file assertions remain covered.

## Verification

- 647 create, workspace, and migration unit tests pass, including Yarn 1/2/4 runner selection and differing ambient/system versions.
- All 15 system package-manager snapshots pass with `--test-threads=2` and cover all four families, managed selection, missing-system fallback, and a real npm dependency installation.
- The default-concurrency run hit seven 50-second timeouts; the lower-concurrency run passed without changing timeouts or expectations.
- New local/global snapshots verify Yarn Classic external-template execution and, separately, its version pin in a built-in template.
- Format, lint, type checks, and whitespace checks pass.
- npm version-pin mismatch reproduced before the fix and corrected afterward.

- After setting the canonical script shell, all 15 system package-manager snapshots pass again; JavaScript syntax and whitespace checks pass. All three macOS snapshot shards pass in CI, including both previously failing external-template cases. The final CI run passes across macOS, Linux, and Windows.

## Limitations

Executable fixtures are Unix-only. The external-template cases verify real execution and generated files; they do not assert every subsequent migration step.


SaKaNa-Y:
For example, even when a user has configured Vite+ to prefer the system-installed pnpm, `vp create` still obtains a Vite+-managed pnpm version and writes it to `devEngines.packageManager.version` in the new project's `package.json`. The same version-pinning issue also affects npm, Yarn, and Bun. With npm, this mismatch can trigger `EBADDEVENGINES` and cause dependency installation to fail.

liangmiQwQ:
> Executable fixtures are Unix-only.

Is that possible to run fixtures on Windows?

packages/cli/src/create/package-manager.ts
```diff
@@ -0,0 +1,61 @@
...
+    process.env.VP_CLI_BIN ??
+    path.join(getVpDirs().bin, process.platform === 'win32' ? 'vp.exe' : 'vp');
+  // The local package can be used without a global installation. In that case
+  // there is no global environment selection to honor.
+  if (!process.env.VP_CLI_BIN && !fs.existsSync(globalVp)) {
+    return downloadPackageManager(packageManager, version, interactive, silent);
+  }
+  // Use the same mode and system-tool lookup as the subsequent `vp install`.
+  // A system-first mode can still fall back to managed when no system tool exists.
+  const result = await runCommandSilently({
```
liangmiQwQ, line 34: I'm thinking about whether it will be better to run `<pm> --version`.

Vite+ allows users who only use `local-cli`. For these users, it's more likely to have the problems this PR want to fix, but they can't be affected by this PR.

liangmiQwQ (commented):
Not sure whether the direction is right. IMO, following `devEngines.packageManager.version`'s version spec should be expected even if users are using their own package manager. Also, pms that used by Vite+'s local cli are always its own, and I'm worried about maintaining an expection would be a bit off. I'm not sure whether this will be hold off for now.

BTW, I am incredibly grateful for your contributions; I was deeply moved by your responsible attitude toward the PR ❤️

SaKaNa-Y:
> > Executable fixtures are Unix-only.
> 
> Is that possible to run fixtures on Windows?

Sorry, this was an oversight on my part. These tests should cover Windows too. Once we’ve discussed and agreed on how to address this issue, I’ll add the necessary Windows support and test coverage.

packages/cli/src/create/package-manager.ts
SaKaNa-Y: You have a good point, I hadn’t considered users who only use the local CLI. I tested this workflow: local `create` successfully installs dependencies using the managed npm version, but subsequently running system npm in the generated project fails with the same `EBADDEVENGINES` version mismatch described in this PR. So local CLI users can encounter this conflict too, and this PR’s fallback logic doesn’t address that scenario.

SaKaNa-Y:
> Not sure whether the direction is right. IMO, following `devEngines.packageManager.version`'s version spec should be expected even if users are using their own package manager. Also, pms that used by Vite+'s local cli are always its own, and I'm worried about maintaining an expection would be a bit off. I'm not sure whether this will be hold off for now.
> 
> BTW, I am incredibly grateful for your contributions; I was deeply moved by your responsible attitude toward the PR ❤️

I agree that existing package-manager version requirements declared by a project should be respected. My question is mainly about new projects that don’t have such a declaration yet. In the scenario this PR aims to fix, the user has already selected system-first mode, but `create` generates a requirement for the managed package-manager version selected by Vite+, which conflicts with the system version used for the subsequent installation.

In this case, shouldn’t the initial requirement match the version of the system package manager the user has chosen to use?

I also think your concern about introducing an exception to the local CLI’s behavior is worth discussing.

Thank you for the thoughtful review ❤️

liangmiQwQ:
> My question is mainly about new projects that don’t have such a declaration yet.

Do you mean users whose installed package manager doesn't automatically use the correct version when they execute the corresponding binary?

> `create` generates a requirement for the managed package-manager version selected by Vite+

Actually, rather than `selected by Vite+`, I think `latest` is semantically better here. But if users set to system-first mode, I agree that Vite+ should honor users' installed package manager.

Anyway, I might try to find a way to solve the problem I mentioned (Vite+'s local cli pm ownership), but not very soon. After that, I think this can be continued. I'll tell you when I'm ready.

SaKaNa-Y:
> > My question is mainly about new projects that don’t have such a declaration yet.
> 
> Do you mean users whose installed package manager doesn't automatically use the correct version when they execute the corresponding binary?

My main point is how `create` should choose the initial package-manager version requirement when a new project doesn’t have one yet, rather than just whether the package manager can automatically switch to the declared version.

For example, a user is running npm 11 in system-first mode, and the template has no package-manager declaration, but `create` writes a requirement for the latest npm version, npm 12. Even if the tool can automatically switch to 12, my question is still: should the initial requirement match the system version the user is already using?

> > `create` generates a requirement for the managed package-manager version selected by Vite+
> 
> Actually, rather than `selected by Vite+`, I think `latest` is semantically better here. But if users set to system-first mode, I agree that Vite+ should honor users' installed package manager.
> 
> Anyway, I might try to find a way to solve the problem I mentioned (Vite+'s local cli pm ownership), but not very soon. After that, I think this can be continued. I'll tell you when I'm ready.

“Latest” is indeed more accurate here. From your reply, I think we agree on how the initial version should be selected in system-first mode.

Thank you for continuing to look into this ❤️

PR #2704, opened 2026-09-14
docs: clarify Nushell XDG startup requirements and recovery

### LINKED ISSUE #2491: Handle Nushell vendor autoload mismatch when XDG variables are set in config.nu
As discussed in the thread at https://github.com/voidzero-dev/vite-plus/pull/2346#discussion_r3805834411, the installer may resolve a different Nushell vendor autoload directory from the session that launched it. This is a follow-up and does not block the current directory layout work.

The installer discovers the user-local vendor autoload directory by starting a new Nushell process and reading `$nu.vendor-autoload-dirs`:

```sh
nu -c '$nu.vendor-autoload-dirs | reverse | each {|dir| $dir } | str join (char nl)'
```

Nushell resolves `XDG_CONFIG_HOME` and `XDG_DATA_HOME` at startup. Its documentation states that these variables must be set before Nushell is launched:

https://www.nushell.sh/book/configuration.html#changing-default-directories

If Nushell starts without `XDG_DATA_HOME` in its parent environment and `config.nu` later sets it to `/custom/data`, the current session continues using the autoload directories that were resolved before `config.nu` was loaded. However, the installer inherits the updated environment and starts a new `nu -c` process, which selects `/custom/data/nushell/vendor/autoload`.

A new top-level Nushell started from a parent where `XDG_DATA_HOME` is still unset will again resolve its original autoload directory before loading `config.nu`. As a result, the installer may report that Nushell was configured successfully while normal new sessions do not load the generated `vite-plus.nu`.

Although setting these variables in `config.nu` is outside Nushell’s documented startup model, leaving `XDG_DATA_HOME` unset is common, and managing environment variables in `config.nu` is a natural choice for users.

As a follow-up, we should confirm this behavior with an integration test and decide whether to handle the mismatch, warn about it, or document that XDG startup variables must be provided by Nushell’s parent process.

I can take responsibility for this task after the current PR is merged.

naokihaba:
resolves: #2491

## Context

Nushell sets up its autoload directories before `config.nu` runs. If `config.nu` then changes `XDG_DATA_HOME` to a different location, the Vite+ installer picks up the new path and writes its snippet to a directory that fresh sessions launched from the original parent environment will not load. If `XDG_DATA_HOME` is never set, this is not an issue.

This setup does not meet [Nushell's documented startup requirement](https://www.nushell.sh/book/configuration.html#changing-default-directories), which states that directory-related XDG variables must be set before Nushell launches. While assigning them in `config.nu` is valid, it only affects child processes and will not update the autoload paths already resolved for the current session.

As described in the [investigation results](https://github.com/voidzero-dev/vite-plus/issues/2491#issuecomment-5668609368), we reproduced the mismatch on macOS with Nushell 0.115.1. Passing the original autoload path explicitly would require changing how the installer is invoked, switching to user autoload only addresses part of the problem, and warning whenever `XDG_DATA_HOME` is set would be noisy for correctly configured users.

Based on those findings, this PR updates the Getting Started guide by documenting the startup requirement and outlining a workaround for affected users, who can manually source Vite+'s `env.nu` in their `config.nu`. We have left the installer itself untouched.

Integration tests reproduce the mismatch, verify that everything works when `XDG_DATA_HOME` is set before startup, and confirm that sourcing the file allows a fresh session to load Vite+ and run `vp help`.


fengmk2 (approved):

naokihaba:
I will fix the conflicts.

naokihaba:
> - rolldown-binaries (ID: 10377436348, Size: 7982906, Expected Digest: sha256:52124c7d1dd27f3f0d3d7a5c581fe0062aea227df5f6898b12841825a26c8bf0)
Error: Unable to download artifact(s): Failed to ListArtifacts: Received non-retryable error: Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403 "Forbidden"

🤔 retry

PR #2705, opened 2026-09-15
fix(pm): reject filtered add and remove on Yarn Classic

jong-kyung:
## Summary
Yarn Classic currently receives `yarn workspaces foreach` for filtered add/remove commands, but it does not support `foreach` and fails with `Invalid subcommand`.

Reject `--filter` with a clear error on Yarn Classic while preserving the existing Yarn Berry behavior. This also covers `vp install <packages> --filter` and Classic remove calls combining `--filter` with `--recursive`.

### Reference
- https://classic.yarnpkg.com/en/docs/cli/workspaces
- https://yarnpkg.com/cli/workspaces/foreach


fengmk2 (approved):

fengmk2:
@codex review

PR #2716, opened 2026-09-15
fix(cli): honor NO_COLOR in Rust output

liangmiQwQ:
Vite+ currently colors Rust diagnostics and status messages unconditionally. `vp check` also forces colors in captured lint output whenever stdout is a terminal, even when `NO_COLOR` is set.

This PR switches Rust CLI styling to `console`, reusing the color detection already used by the installer and Rust prompts. Diagnostic styling checks stderr. Captured lint output follows the parent's color support and preserves an explicit `FORCE_COLOR` value.

🤖 Generated with Codex


liangmiQwQ:
@codex review

fengmk2 (approved):

PR #2720, opened 2026-09-16
fix(create): inline generator dependencies without catalog support

SaKaNa-Y:
> [!IMPORTANT]
> Please take a moment to read this. Thank you!
>
> I should include a brief explanation of the problem in my own words in every PR. If that explanation is missing, please @mention me and do not merge this PR until I have added it. You may also leave this PR unaddressed (because this means I have not fulfilled my responsibilities as the author).
>
> If my explanation is unclear or difficult to follow, please ask me to clarify or provide reproduction steps or supporting evidence.
>
> I welcome suggestions and counterarguments, especially questions about anything I may have overlooked. (Your feedback helps me learn and improve. 🙏)
>
> I hold myself to this standard for every PR, regardless of its size.

## Problem

In an npm monorepo, running `vp create vite:generator --directory tools/gen --no-interactive` creates the generator but fails to install its dependencies:

```text
EUNSUPPORTEDPROTOCOL
Unsupported URL Type "catalog:": catalog:
```

Subsequent `npm install` commands also fail. The generator template declares `@types/node` and `typescript` as `catalog:`, and scaffolding copies those declarations without checking whether the selected package manager supports catalogs.

## Fix

Use the existing package-manager capability check when scaffolding a generator. For managers without catalog support, replace the template's catalog references with dependency ranges from the bundled monorepo catalog. Catalog-capable workspaces keep their existing references.

This restores dependency installation without duplicating version constants. The installation-failure reporting tracked in #2453 and #2543 remains separate.

## Verification

- Regression tests fail before the fix and pass afterward for npm, Yarn 4.9 and pnpm 9.4. Additional cases verify catalog preservation for supported pnpm, Yarn and Bun versions.
- Full unit suite: **1,214 passed, 1 skipped**. Formatting, lint and type checks passed.
- `cargo test -p vp_cli_snapshots --offline -- create_generator`: **3 cases passed**, covering npm installation, the existing pnpm generator workflow and rejection outside a monorepo.
- Manual npm workflow: after scaffolding, dependency installation and the generated executable's `--help` both succeed.

The full CLI snapshot suite was not run; CLI verification covered the three generator cases above.


SaKaNa-Y:
When running `vp create vite:generator` in an npm-managed monorepo, the files are generated successfully, but the subsequent dependency installation fails because the generated `package.json` contains `catalog:` dependency references, which npm does not support. Running `npm install` manually afterward fails for the same reason.

fengmk2 (approved):

fengmk2:
@SaKaNa-Y Thanks!

PR #2729, opened 2026-09-17
fix(cli): handle Homebrew installs during first-run setup

### LINKED ISSUE #2719: error: Command execution failed: Permission denied (os error 13) - 0.3.2
### Describe the bug

after installing vite-plus 0.3.2 via brew on macOS, we cannot use vp cli anymore. getting following error 

```shell
 vp env off   
✔ Would you like Vite+ to manage your Node.js and package-manager versions? · no
error: Command execution failed: Permission denied (os error 13)


VP_NODE_MANAGER=no VP_PM_MANAGER=no vp env off
error: Command execution failed: Permission denied (os error 13)
```

we have to use `export NPM_CONFIG_REGISTRY=company-private-registry`  in my org and this did not allow me to pass `npmToken` 

### Reproduction

https://github.com/voidzero-dev/vite-plus/issues

### Steps to reproduce

``` 
vp dev         
✔ Would you like Vite+ to manage your Node.js and package-manager versions? · no
error: Command execution failed: Permission denied (os error 13)
```

### System Info

```shell
vp --version
✔ Would you like Vite+ to manage your Node.js and package-manager versions? · no
error: Command execution failed: Permission denied (os error 13)
```

### Used Package Manager

pnpm

### Logs

```shell
vp --version
✔ Would you like Vite+ to manage your Node.js and package-manager versions? · no
error: Command execution failed: Permission denied (os error 13)
```

### Validations

fengmk2:
Fixes #2719.

First-run setup checks `.vp-setup-complete` beside `vp`, but writes it beside the managed copy. External installations therefore repeat setup and ignore their bundled JavaScript. A retry can also fail with `EACCES` when an interrupted setup leaves a read-only binary copy.

External binaries now record completed setup under `<STATE>/self-setup`. Complete Unix packages retain their binary and bundled JavaScript. Binaries without a bundled CLI reuse their completed managed installation. Setup replaces incomplete binary copies atomically.

Unix shims follow a public `vp` entrypoint that resolves to the running binary. This keeps them valid after package replacement. Setup preserves saved Node.js and package-manager preferences when the external binary changes.

Homebrew installations direct `vp upgrade` to `brew upgrade vite-plus` and disable npm update checks. `vp implode` removes user data and directs users to `brew uninstall vite-plus` for package removal. `vp env doctor` reports Homebrew ownership and checks the CLI path separately from the user shim directory.

Explicit installer requests still run setup. Vite+-managed installations keep their existing upgrade and rollback behavior.


docs/guide/implode.md
```diff
@@ -21,3 +21,14 @@ Skip the confirmation prompt with:
...
+## Homebrew
+
+Run `vp implode` first to remove Vite+-managed runtimes, global packages, configuration, shims, and shell entries. Then remove the Homebrew package:
+
+```bash
+vp implode
+brew uninstall vite-plus
+```
+
+Running `vp` again before removing the Homebrew package starts first-run setup again.
```
fengmk2: I reproduced this with `vite-plus 0.3.2_1` built by Homebrew. Bash kept `$VP_HOME/bin/vp` in its command hash after `vp implode` removed the shim. In the same shell, the next `vp` exited with status 127 and `No such file or directory`. `hash -r` or a new terminal let Bash find `/opt/homebrew/bin/vp` and start setup. Could we qualify this sentence?

```suggestion
After restarting your terminal, or running `hash -r` in Bash, running `vp` before you remove the Homebrew package starts first-run setup again.
```

liangmiQwQ (changes requested):
When a user run Vite+ for the first time, it will run self-setup logic first and then continue running the original commands.

This is an intentional design in my previous PR, and I think we shouldn't do that for HomeBrew installation. Commands like `vp env use` doesn't work, shims like `node` / `pnpm` wouldn't be loaded at that time, either.

We'd better prompt users to restart the terminal after self-setup, and stop running users' commands like normal installation for HomeBrew. Adding a document section for that might also be good.

That said, we still need to keep commands running behavior by default to ensure the compatibility.

liangmiQwQ (approved):
The new `self-setup` mark file location and other ideas LGTM :)

Brooooooklyn (approved):

PR #2737, opened 2026-09-18
fix(ci): exclude Void packages from release age checks

fengmk2:
New `void` releases can block docs deployment because their `@void/*` dependencies fail the `minimumReleaseAge` check. Add `@void/*` to `minimumReleaseAgeExclude` alongside the existing `void` entry so deployments can use new CLI releases.


cpojer (approved):
You need to add both "@void/*" and "void".

cpojer:
Ah I see that void itself was already in the list!

fengmk2:
> You need to add both "@void/*" and "void".

`void` already in minimumReleaseAgeExclude list 😢

PR #2742, opened 2026-09-18
fix(cli): default unpinned npm to Node's bundled version

### LINKED ISSUE #2726: Use the npm version bundled with node
### Description

Hi 👋🙂

At my place of work we always use the npm version that's bundled with Node. This keeps things simple and means we don't have to manage two different version numbers. It also always guarantees compatibility.

Currently VP, when only the Node version is set, will install the latest npm version. Could it instead default to the version bundled with Node?

### Suggested solution

Change the priority order for npm to:

1. Explicit command override
2. VP_PACKAGE_MANAGER
3. Top-level packageManager
4. devEngines.packageManager
5. ~Lockfile or manager-specific configuration~ <-- This is what is causing the override.
6. The named package manager's global default version
7. The named shim's latest release

### Alternative

_No response_

### Additional context

```
  "devEngines": {
    "runtime": {
      "name": "node",
      "version": "^24.14.0"
    }
  },
```

Would, at time of writing, resolve to Node v24.21.0 and npm v11.19.1. However VP installs npm v12.0.2 stating source as "lockfile or config".

### Validations

liangmiQwQ:
Close #2726

An npm lockfile now selects npm without requesting its latest release. When no npm version is pinned or configured as the default, vp commands and npm/npx shims use the npm bundled with the selected Node.js runtime. Explicit npm versions and configured defaults keep their precedence.

Direct shims and vp commands share default-version resolution after selecting the package-manager family, while keeping their independent environment overrides.

`vp env current` reports the bundled npm version and source. Explicit standalone installation with `vp env install npm` keeps its registry fallback.

🤖 Generated with Codex


crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_env_install_no_arg/snapshots.toml
```diff
@@ -13,14 +13,14 @@ steps = [
...
+comment = "Explicit npm installation uses standalone registry npm without installing Node.js."
 local-registry = true
 skip-platforms = ["windows"]
 seed-runtime = false
 env = { VP_ENV_USE_EVAL_ENABLE = "1", VP_SHELL = "bash" }
 steps = [
   { argv = ["vp", "env", "use", "npm", "--no-install"], comment = "an explicit npm scope exports the standalone npm fallback" },
   { argv = ["vp", "env", "install", "npm"], comment = "an explicit npm scope installs the standalone registry fallback" },
-  { argv = ["vp", "env", "current", "npm", "--json"], comment = "the standalone npm fallback is installed" },
+  { argv = ["vp", "env", "list", "npm", "--json"], comment = "the standalone npm fallback is installed" },
```
fengmk2: Given the addition of the `list` command, can we avoid deleting the tests for the `current` command?

fengmk2 (approved):

liangmiQwQ:
@codex review

liangmiQwQ:
@codex review

fengmk2:
@codex review

Brooooooklyn:
@codex review

liangmiQwQ:
Codex review seems broken.

fengmk2:
I try to use codex app review from now on.

crates/vp_global_cli/src/commands/env/current.rs
```diff
@@ -215,13 +215,59 @@ async fn resolve_package_manager_info(
...
+            && let Some(node_path) = crate::shim::dispatch::find_system_tool("node")
+        {
+            let node_path = crate::shim::dispatch::resolve_external_node_executable(node_path)
+                .map_err(|error| Error::Other(error.into()))?;
+            node_path
+                .parent()
+                .ok_or_else(|| Error::Other("Node has no bin directory".into()))?
+                .to_absolute_path_buf()
+        } else {
+            super::resolve_node_bin_dir(cwd, config).await?
```
fengmk2: **[P2] Keep `env current` from installing Node**

When the selected Node version is not installed, `resolve_node_bin_dir` calls `download_runtime`, so `vp env current --json` and `vp env current npm --json` now try to install Node merely to report the environment. With `VP_NODE_VERSION=22.18.0` and an unavailable Node mirror, the command exits 1 without JSON while fetching `SHASUMS256.txt`; `env current node --json` still reports `installed: false`. Resolve the installation path without downloading, read the npm manifest if present, and report `installed: false` with an unknown npm version when it is absent.

crates/vp_pm_cli/src/package_manager.rs
```diff
@@ -228,9 +228,30 @@ impl PackageManagerBuilder {
...
             get_package_manager_type_and_version(&workspace_root, self.client_override)?;
 
+        // A lockfile selects npm, but does not request a version separate from Node's npm.
+        if package_manager_type == PackageManagerType::Npm
+            && matches!(
+                source,
+                PackageManagerSource::LockfileOrConfig | PackageManagerSource::Default
+            )
+        {
+            let mut manager = crate::helpers::default_npm_package_manager(&self.cwd);
```
fengmk2: **[P2] Preserve the resolved npm executable directory**

`default_npm_package_manager` sets `bin_prefix` to `<cwd>/bin`, and this branch only updates its version. `resolve_for_manager` later prepends that directory to `PATH`, so an npm-lockfile project containing an executable `bin/npm` runs that project executable instead of bundled npm, even though the version probe inspected bundled npm. Adding a `bin/npm` that prints its arguments makes `vp pm version --json` print the project script's output instead of npm's JSON. Preserve the actual resolved npm directory, or explicitly retain inherited-PATH execution, instead of treating `<cwd>/bin` as the npm installation.

crates/vp_pm_cli/src/package_manager.rs
```diff
@@ -553,20 +572,23 @@ fn environment_package_manager_default(
...
-            .filter(|(package_manager, _, _)| *package_manager == expected)
-            .map(environment_package_manager_default);
-    }
+    let kind =
+        expected.or_else(|| resolution.as_ref().map(|resolution| resolution.package_manager_type));
+    // A matching project version wins; an npm lockfile only selects the family.
+    let resolution = resolution.filter(|resolution| {
+        Some(resolution.package_manager_type) == kind
+            && !(resolution.package_manager_type == PackageManagerType::Npm
+                && resolution.source == PackageManagerSource::LockfileOrConfig.description())
```
fengmk2, line 595: **[P2] Handle bundled npm in `env print pm`**

For an npm-lockfile project without an npm pin or default, filtering out this resolution returns `None`, which `print_env` also interprets as no package manager selected. Since the `pm` scope excludes Node, `vp env print pm` then exits 1 with `no selected environment component could be resolved`, even when Node and bundled npm are already installed and `vp pm version --json` succeeds. Update `print_env` to emit the bundled npm directory for this selection, or preserve an explicit bundled-npm result so callers can distinguish it from no selection.

crates/vp_pm_cli/src/package_manager.rs
```diff
@@ -228,9 +228,30 @@ impl PackageManagerBuilder {
...
+            )
+        {
+            let mut manager = crate::helpers::default_npm_package_manager(&self.cwd);
+            // Version gates must describe the npm on PATH, not the latest registry release.
+            let npm = vp_command::resolve_bin("npm", None, &self.cwd)?;
+            let output = tokio::process::Command::new(npm.as_path())
+                .arg("--version")
+                .current_dir(&self.cwd)
+                .output()
+                .await?;
```
fengmk2, line 249: **[P2] Exclude user preloads from the npm version probe**

This probe inherits `NODE_OPTIONS` and uses all of stdout as the npm version. A preload containing `console.log('preload start')` makes the version invalid, so `resolve_for_manager` falls back to `Npm::unknown_version()` and permits newer commands. With bundled npm 10.9.3 and `NODE_OPTIONS='--require ./preload.cjs'`, `vp pm patch example` and `vp pm approve-builds` consequently fail with `Unknown command: "patch"` and `Unknown command: "approve-scripts"` instead of reporting the expected unsupported-command warnings. Remove user preload options for this internal probe, preserve them for the actual command, and validate the version output.

crates/vp_global_cli/src/commands/env/mod.rs
```diff
@@ -192,6 +196,13 @@ async fn print_env(cwd: AbsolutePathBuf, scope: Option<String>) -> Result<ExitSt
                 )
                 .await?;
                 bin_dirs.insert(0, install_dir.join("bin").as_path().display().to_string());
+            } else if selected_type == Some(vp_pm_cli::PackageManagerType::Npm)
+                && !scope.includes_node()
+            {
+                // Unpinned npm shares Node's bin directory, including in the pm-only scope.
+                bin_dirs.push(
+                    resolve_node_bin_dir(&cwd, &modes).await?.as_path().display().to_string(),
```
fengmk2, line 204: **[P2] Resolve the actual Node directory before printing bundled npm PATH**

With Node in `system_first` mode and npm in `managed` mode, `resolve_node_bin_dir` returns the parent of the Node executable found on PATH. If that executable is an external version-manager shim, this is the shim directory, which can contain a separately selected npm. Applying `vp env print pm` or `vp env print npm` then puts that external npm ahead of the managed shim. I reproduced `npm --version` and `env current npm` selecting bundled npm `10.9.3`, but npm changing to `10.8.2` after applying either printed PATH statement. Resolve the external Node shim through `resolve_external_node_executable` before deriving this directory, as `env current npm` already does, and add a mixed-mode test that actually executes npm after applying the output.

PR #2749, opened 2026-09-19
fix(cli): import local versions module via file URL

YanChenBai:
## Summary

Fix loading tool versions from the locally installed `vite-plus` package by converting the generated `dist/versions.js` filesystem path to a `file://` URL before passing it to dynamic `import()`.

## Why

Dynamic `import()` expects a valid module specifier. Passing an absolute filesystem path directly is not portable, especially for Windows drive-letter paths such as `C:\...`, which may be interpreted incorrectly by Node's ESM loader.

Using `pathToFileURL()` produces a valid file URL and allows the generated versions module to be imported consistently across platforms.

## Changes

* Build the path to `dist/versions.js` with `path.join()`
* Convert the filesystem path with `pathToFileURL()`
* Import the generated versions module using the resulting `file://` URL

🤖 Generated with Codex


fengmk2:
@YanChenBai can you add a snapshot test to cover this bug fix?

YanChenBai:
> @YanChenBai能否添加一个快照测试来验证这个bug修复？

Yep, I'll add a snapshot test covering this regression. Thanks!

fengmk2 (approved):

PR #2750, opened 2026-09-19
fix(pm): default to pnpm when no package manager is detected

liangmiQwQ:
When no package manager is detected, package-manager commands fall back to npm even though the install guide documents pnpm as the default.

Default these commands, including `dlx` and `vpx`, to pnpm without prompting. Detected project settings still take precedence, and commands outside a project use the same fallback without creating `package.json`.

🤖 Generated with Codex


crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/command_vpx_no_package_json/snapshots/command_vpx_no_package_json.md
```diff
@@ -2,9 +2,37 @@
 
 ## `vpx -s cowsay hello`
 
-should work without package.json
+select a manager without package.json
```
fengmk2: In this case, is it not necessary to let users choose? Can we just default to pnpm?

liangmiQwQ: I intentionally unified the behavior for all package manager usecases here, but I agree `vpx` should be different. Do you think read-only commands like `vp info` should also default to pnpm? (or npm?)

fengmk2: If no package manager type is detected, I think using pnpm by default is the most appropriate.

liangmiQwQ:
https://github.com/voidzero-dev/vite-plus/pull/2750/changes#diff-8f574945e940cc82ca9c381d8dfa8bebb8741c4ecab3fe53e92daac66449ae22R199-R224

This shows what commands would require users' select.

crates/vp_cli_snapshots/tests/cli_snapshots/fixtures/install_package_manager_selection/snapshots/install_package_manager_selection.global.md
```diff
@@ -0,0 +1,70 @@
...
+**→ expect-milestone:** `select:package-manager:0`
+
+```
+VITE+ - The Unified Toolchain for the Web
+
+No package manager detected. Please select one:
+   Use ↑↓ arrows to navigate, Enter to select, 1-4 for quick selection
+   Press Esc, q, or Ctrl+C to cancel installation
+
+  ▶ [1] pnpm (recommended) ←
```
fengmk2: Will it be selected every time it is executed, or only the first time?

If there is only the first time, then you need to add a test command to execute `vp install` again to ensure that the next execution does not require the user to make a selection again.

liangmiQwQ: every time for now.

liangmiQwQ: if you think it is not right, default to `pnpm` for all commands might make sense.

fengmk2: let's default to pnpm 👍

fengmk2 (approved):

PR #2763, opened 2026-09-21
fix(env): initialize package-manager preferences during setup

fengmk2:
An upgrade can leave package-manager preferences unset. The first `pnpm` call can then remove the shim that Bash cached. The next call fails with exit code `127`.

`vp env setup` now saves missing preferences before it creates or removes shims. Older releases already invoke this command in the new binary during upgrade. The migration also covers fresh installations with incomplete preferences.

Setup preserves saved choices. For each missing family, it uses this order:

- Keep managed mode if the main bin directory contains a Vite+ shim or a recorded global package-manager installation.
- Use system-first mode if a fallback shim or a system package manager exists.
- Otherwise, use managed mode.

Setup checks aliases such as `pnpx` with their package-manager family. Package-manager calls no longer prompt, save preferences, or move shims.

```mermaid
flowchart LR
    subgraph before["Before"]
        direction TB
        B1["Upgrade creates main-bin shims<br/>Preferences remain unset"] --> B2["First pnpm call<br/>Bash caches the shim path"]
        B2 --> B3["User chooses system pnpm<br/>Call removes the cached shim"]
        B3 --> B4["First call succeeds"]
        B4 --> B5["Next call uses the missing path<br/>Exit code 127"]
    end
    subgraph after["After"]
        direction TB
        A1["Upgrade runs new binary setup"] --> A2["Read saved choices<br/>Infer missing choices from old layout"]
        A2 --> A3["Existing main-bin shim<br/>Select managed mode"]
        A2 --> A4["Existing system tool<br/>Select system-first mode"]
        A3 --> A5["Save preferences<br/>Keep main path or create fallback shim"]
        A4 --> A5
        A5 --> A6["Repeated calls succeed<br/>Bash cache remains valid"]
    end
    classDef failure fill:#ffebe9,stroke:#cf222e,color:#82071e
    classDef success fill:#dafbe1,stroke:#1a7f37,color:#116329
    class B5 failure
    class A6 success
```

Fixes the [reported Bash command-cache failure](https://gist.github.com/fengmk2/c8d91cd507eed72299524c678ea2ca9b).


docs/guide/env.md
```diff
@@ -122,7 +122,9 @@ vp env off
 
 This switches both components to system-first mode. Vite+ prefers system tools and falls back to managed installations. Mixed configurations compose: package-manager launchers that look up Node.js through PATH receive the runtime selected by the Node mode.
 
-Using `pm` records the selected mode for all currently supported package managers and replaces their individual choices. An unscoped `on` or `off` does the same while also changing Node.js. A family without a recorded mode defaults to managed mode until an `on` / `off` command configures it.
+Using `pm` records the selected mode for all currently supported package managers and replaces their individual choices. An unscoped `on` or `off` does the same while also changing Node.js.
+
+Setup fills missing family preferences from the existing tool layout. An existing managed package manager or Vite+ shim in the main bin directory keeps its family in managed mode. Otherwise, a fallback shim or an existing system package manager selects system-first mode; a family with neither uses managed mode. Upgrades run this setup automatically. Package-manager calls do not prompt or change these preferences.
```
liangmiQwQ: As this is a very cold branch (Only for users who update Vite+), I prefer keep the document unchanged or shorter. Saying that `Package-manager calls do not prompt or change these preferences.` is unnecessary.

liangmiQwQ (approved):
The direction LGTM if we decide to let Vite+ automatically select management preferences. 💯

Brooooooklyn (approved):

PR #2770, opened 2026-09-21
fix: support pn and pnx package-manager shims

iruoy:
Vite+ creates `pnpm` and `pnpx` shims but omits the official `pn` and `pnx` aliases introduced in pnpm 11. Add both aliases to shim setup, family detection, and diagnostics, routing `pn` to the managed `pnpm` binary and `pnx` to `pnpx` (`pnpm dlx`).

Reusing the existing binaries supports cached installations and pnpm versions before 11. Nested alias calls preserve an explicitly selected package-manager version. Documentation and setup snapshots include the new shims.

Validation:
- `just check`, `just test`, and `just lint` passed (2,143 Rust tests; 8 ignored).
- 43 focused CLI snapshot cases passed in comparison mode (3 ignored), including pnpm 10 and native pnpm 12, version overrides, nested invocations, setup, and mode switching.
- `cargo fmt --all -- --check` and `git diff --check` passed.
- Windows execution was not tested locally.


liangmiQwQ:
To fix failing lint CI, add this line to codebase's root `.typos.toml`

https://github.com/voidzero-dev/vite-plus/blob/bcc11eecb30f00623071c2333cc2a3dfd14dfdd3/.typos.toml

```toml
pn = "pn" # pnpm short alias
```

liangmiQwQ (approved):
Nailed it 💯

The reason why Vite+ previously skipped these binaries is that `pn` and `pnx` are not included under `pnpm@10`, and this PR solves the problem perfectly!

fengmk2 (approved):

